PR #1, opened 2024-10-03
feat: added episodic stats with eval_analyzer.py and stats_library.py with Elo test, and Mann-Whitney U test

Al-does:
### Description
This PR introduces episodic stats reporting, Mann-Whitney U tests, and Elo ratings for evaluating RL agent performance.

- Updated evaluator.py to output episodic stats for every agent.
- Updated tools/eval.py to use the print_policy_stats() function from eval_analyzer.py and to use the arguments for eval_method and stat_category. 
- Added util/eval_analyzer.py which replaces stats.py to handle statistical analysis logic. It includes lookups for routing game stats to different stats functions and for different groups of agent performance attributes.
- Added util/stats_library.py: Added function stubs for future statistical methods. Implemented a Mann-Whitney U test function with output to the console, showing significance and effect size with color coding (green for positive significant, red for negative). The U test currently supports 1v1 agent matchups, assuming a 50% agent control per policy. 
Also integrated an Elo rating system with a dynamic K value that decreases across episodes for comparison. This implementation only compares agents against Policy 1 and does not track baseline-versus-baseline matchups.

### Future Work
The Elo system is currently limited to comparing agents with Policy 1 which limits the use of Policy 1's Elo score as an absolute value for comparison. Expanding to baseline-vs-baseline comparisons is a possible next step, as is implementing a Glicko-2 system.


daveey:
Merged into main manually

PR #3, opened 2024-11-09
Implemented Glicko2 scoring

Al-does:
- Updated DevOps/setup_build.sh to pull where needed.
- Implemented Glicko2 scoring in stats_library.py and added verbose results for inspection.
- Implemented the ability to read and write historical scores for Elo and Glicko2 to and from JSON files in the evals folder. The read and write blocks in eval.py can readily be swapped for another method of reading the historicals. Updating what needs to be written to disk takes place within the stats class and involves taking in the entire dictionary of historical data. Each item is one policy, its score, number of games played, and (for Glicko2 only) rating deviation (RD) and sigma.


evals/elo_scores.json
```diff
@@ -0,0 +1,26 @@
+{
```
daveey, line 1: don't check this in

.vscode/launch.json
```diff
@@ -109,6 +109,16 @@
             "type": "debugpy",
             "request": "launch",
             "module": "tools.eval",
+            // "args": [
```
daveey, line 112: don't check this in

evals/glicko_scores.json
```diff
@@ -0,0 +1,38 @@
+{
```
daveey, line 1: dont include

rl/pufferlib/trainer.py
```diff
@@ -99,6 +102,40 @@ def train(self):
         self._save_policy_to_wandb()
         logger.info(f"Training complete. Total time: {self.train_time:.2f} seconds")
 
+        # ---update Glicko2 scores and log them---
+        try:
+            evaluator = hydra.utils.instantiate(self.cfg.evaluator, self.cfg, self.policy_store)
```
daveey, line 107: This doesn't look right. I think it would evaluate a policy in the config, not the newly-trained policy.

tools/eval.py
```diff
@@ -33,21 +33,59 @@ def main(cfg):
...
-        #delete below after testing. Some dummy historical elo scores :)
-        initial_elo_scores = {
-            'b.gpop.simple.2:v0': {'score': 500, 'episodes': 37},
-            'Policy 2': {'score': 1500, 'episodes': 12}
-        }
-
-        elo_analysis = Analysis(stats, eval_method='elo_1v1', stat_category='altar', initial_elo_scores=initial_elo_scores)
+        #--------Elo Analysis--------#
+        # Load the historical elo scores from disk
+        with open("evals/elo_scores.json", "r") as file:
```
daveey, line 38: the path should be an optional config argument

daveey (commented):
The stats logic looks good, but the saving / loading / train flow can be better. Let's merge this in together and I can show you some ways to re-structure

daveey (changes requested):
i think this is no-longer needed?

Al-does:
The latest version of main has everything necessary to make Glicko work (although I still have to run some tests to ensure it's still working correctly). It also seems to have all of the necessary `git pull` lines in setup_build.sh, at least from eyeballing it. That is to say, correct, not needed anymore.

PR #10, opened 2024-12-10
Fixed the divisibility bug.

Al-does:
Ensured that the number of environments is an integer multiple of the number of baseline policies. It will print to console telling the user that it adjusted that configuration variable.

### DIFF
--- a/rl/pufferlib/evaluator.py
+++ b/rl/pufferlib/evaluator.py
@@ -20,7 +20,6 @@ class PufferEvaluator():
         self._cfg = cfg
         self._device = cfg.device
 
-        self._vecenv = make_vecenv(self._cfg, num_envs=cfg.evaluator.num_envs)
         self._num_envs = cfg.evaluator.num_envs
         self._min_episodes = cfg.evaluator.num_episodes
         self._max_time_s = cfg.evaluator.max_time_s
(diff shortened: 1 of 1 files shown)


rl/pufferlib/evaluator.py
```diff
@@ -34,12 +33,18 @@ def __init__(
             self._policy_agent_pct = 0.9
 
         self._agents_per_env = cfg.env.game.num_agents
-        self._total_agents = self._num_envs * self._agents_per_env
         self._policy_agents_per_env = max(1, int(self._agents_per_env * self._policy_agent_pct))
         self._baseline_agents_per_env = self._agents_per_env - self._policy_agents_per_env
+        modulo = self._num_envs % len(self._baseline_prs)
```
daveey, line 38: self._num_envs = self._num_envs // self._baseline_prs * self._baseline_prs will do the same thing

rl/pufferlib/evaluator.py
```diff
@@ -48,7 +53,7 @@ def __init__(
             .reshape(self._policy_agents_per_env * self._num_envs)
 
         self._baseline_idxs = []
-        if len(self._baseline_prs) > 0:
+        if len(self._baseline_prs) > 0: #shouldn't this always be true given line 33 above?
```
daveey, line 56: i think so

daveey (approved):

daveey:
looks good, feel free to merge after updating

PR #32, opened 2025-01-08
Comment out the huggingface token section

daphnedemekas:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1209230969804461/1209239114831103) by [Unito](https://www.unito.io)


daveey (changes requested):
Did you run this? Seems like the variable is being used lower in the file and should also be removed.

sasmith:
Sorry for using this as a playground, but here's another comment.

PR #34, opened 2025-01-09
Alex bigger models

Al-does:
Update actor and/or critic hidden sizes via params in config/agent/simple.yaml. This change is likely not backwards compatible with older models.


rl/pufferlib/puffer_agent_wrapper.py
```diff
@@ -22,15 +24,26 @@ def __init__(self, env, policy, input_size=512, hidden_size=512, num_layers=1):
         super().__init__(env, policy, input_size, hidden_size, num_layers)
 
 class PufferAgentWrapper(nn.Module):
-    def __init__(self, agent: MettaAgent, env: PettingZooPufferEnv):
+    def __init__(self, agent: MettaAgent, actor_hidden_sizes: List[int], env: PettingZooPufferEnv):
```
daveey, line 27: can we move this code into MettaAgent, make the actor in there, and then just use it here?

daveey: then we have access to the cfg in there

daveey (approved):

PR #35, opened 2025-01-10
Updated README

Al-does:
Added additional game-mechanic descriptions.


daveey (changes requested):
Can you put this in the mettagrid repo instead? That's where the environment details should live.

PR #38, opened 2025-01-13
Auto Glicko2 logging

Al-does:
- Since evaluation is now the default for trainer, I added a way to turn it off via `evaluate: true` in `puffer.yaml`
- The evaluation defaults to using wandb-checkpointed baselines instead of local checkpoints as I figured that the former would be more stable. 
- Historical scores will be saved in the default location $run/eval as it's not critical to the scoring that we use historical data. If we want to make this more central then please let me know.

### DIFF
--- a/rl/pufferlib/trainer.py
+++ b/rl/pufferlib/trainer.py
@@ -115,11 +115,10 @@ class PufferTrainer:
         logger.info(f"Training complete. Total time: {self.train_time:.2f} seconds")
 
     def _evaluate_policy(self):      
-        if self.cfg.evaluator.baselines.uri:
-            baseline_records = self.policy_store.policies(self.cfg.evaluator.baselines)
-        else:
-            self.cfg.evaluator.baselines.uri = f"wandb://run/{self.cfg.run}"
-            baseline_records = self.policy_store.policies(self.cfg.evaluator.baselines)
+        if not self.cfg.evaluator.baselines.uri:
+            self.cfg.evaluator.baselines.uri = f"file://{self.cfg.trainer.checkpoint_dir}"
+        
+        baseline_records = self.policy_store.policies(self.cfg.evaluator.baselines)
         
         evaluator = hydra.utils.instantiate(self.cfg.evaluator, self.cfg, self.last_pr, baseline_records)
         stats = evaluator.evaluate()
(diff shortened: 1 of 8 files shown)


rl/pufferlib/trainer.py
```diff
@@ -114,11 +114,13 @@ def train(self):
...
-    def _evaluate_policy(self):
-        if self.cfg.evaluator.baselines.uri is None:
-            return
-
-        baseline_records = self.policy_store.policies(self.cfg.evaluator.baselines)
+    def _evaluate_policy(self):      
+        if self.cfg.evaluator.baselines.uri:
+            baseline_records = self.policy_store.policies(self.cfg.evaluator.baselines)
+        else:
+            self.cfg.evaluator.baselines.uri = f"wandb://run/{self.cfg.run}"
```
daveey, line 121: We don't need to set this, it's not used other than to load the baselines

daveey, line 121: We only upload checkpoints to Wandb rarely, so I think we should use the local ones. I think just pointing to the run_dir/checkpoints should automatically use the latest ones

daveey (changes requested):

rl/pufferlib/trainer.py
Al-does: For sure. Made changes and they worked locally. Now testing on aws.

Al-does: Makes sense. Thanks for the pointer.

rl/pufferlib/trainer.py
```diff
@@ -102,10 +102,10 @@ def train(self):
             self._process_stats()
             if self.epoch % self.trainer_cfg.checkpoint_interval == 0:
                 self._checkpoint_trainer()
+            if self.epoch % self.trainer_cfg.evaluate_interval == 0 and self.trainer_cfg.evaluate:
```
daveey, line 105: maybe we should nest this in the previous if, so that this only happens if the checkpoint is saved?

rl/pufferlib/trainer.py
```diff
@@ -114,11 +114,12 @@ def train(self):
         self._save_policy_to_wandb()
         logger.info(f"Training complete. Total time: {self.train_time:.2f} seconds")
 
-    def _evaluate_policy(self):
-        if self.cfg.evaluator.baselines.uri is None:
-            return
-
+    def _evaluate_policy(self):      
+        if not self.cfg.evaluator.baselines.uri:
+            self.cfg.evaluator.baselines.uri = f"file://{self.cfg.trainer.checkpoint_dir}"
```
daveey, line 119: we don't need file://, but it's fine to have it

daveey (changes requested):
sorry, just one more change :)

PR #42, opened 2025-01-15
Clone the pufferlib branch `metta` that we want at depth=1 for speed.

treeform:


daveey (approved):
Looks fine, I had to do stash because the build left some modified files. But maybe that's all fixed now

PR #48, opened 2025-01-20
Map generation config

daphnedemekas:
Currently have a tool that generates ascii maze maps, given a width parameter, a height parameter, start and end positions (which dictate the agent's initial position and the position of the heart altar), and a branching parameter. 
The agent starts with initial energy 255 and the heart altar has 255 cooldown. 

The branching parameter ranges between 0 and 1. 
In large mazes, more branching is likely to lead to higher complexity, because there are more dead ends, however in smaller mazes, more branches can make it easier, since it is more likely to open up multiple paths towards the end state. 

Evidently, the difficulty of the maze is also governed by how far away the agent begins from it, i.e. the distance between the start and end position relative to the size of the map. 
Currently this is specifiable by command line arguments, otherwise it's opposite corners. 

Here are some examples of mazes with different parameters

11x11, branching 0.1: 
Screenshot 2025-01-20 at 11 38 12 AM

11x11, branching 0.9:
Screenshot 2025-01-20 at 11 43 43 AM

19x19, branching 0.1:
Screenshot 2025-01-20 at 11 44 45 AM

 19x19, branching 0.9
Screenshot 2025-01-20 at 11 45 39 AM

39x39, branching 0.9
Screenshot 2025-01-20 at 11 46 45 AM


daphnedemekas:
Screenshot 2025-01-20 at 1 01 40 PM
Multi-room maze generation

daphnedemekas:
Multi room cylinder also . Going to want to make a better structure for tooling and stacking as we generate more maps 
Screenshot 2025-01-20 at 1 31 08 PM

configs/env/scripts/cylinder_task.py
```diff
@@ -0,0 +1,306 @@
+import random
```
daveey, line 1: we shouldn't keep code in the configs directory. also, this code is useful in mettagrid repo

let's put it into mettagrid:mettagrid/level_gen/cylinder.py

configs/env/scripts/maze_generation.py
```diff
@@ -0,0 +1,276 @@
...
+
+def save_grid(grid, width, height, filename):
+    """Save grid to file"""
+    with open(filename, 'w') as f:
+        for y in range(height):
+            for x in range(width):
+                f.write(grid[(x, y)])
+            f.write('\n')
+
+if __name__ == "__main__":
```
daveey, line 227: move the main() for both of these into mettagrid:level_gen/maze.py

daveey (changes requested):
Fast turnaround!

These should go into the mettagrid repo, and we should separate the logic from the script that runs it. we want to call that from game_builder.py

daphnedemekas:
Moved all the code into mettagrid, all that remains here is an example config of how to use the maps

daveey (approved):
Thanks for moving the code out. can we have a better name than "custom_map"?

maybe examples/maze_and_cylinder.yaml

feel free to merge after renaming

PR #50, opened 2025-01-24
Add user launch configs

sasmith:
This adds config yamls that take over a lot of what launch.json used to do. This is missing the corresponding launch.json changes to use the configs, since we want to test them with less disruption to other folks.

The layout still has some smell. There's certainly room for common setups for each run type, to further minimize what needs to be in each user config. Dave's also excited about getting all of a user's configs into a single file, and being able to pull out a specific section, but I haven't gotten that to work.

---
  - https://app.asana.com/0/0/1209230815163490

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1209230969804461/1209239112253087) by [Unito](https://www.unito.io)

### DIFF
--- a/configs/user/daveey/eval.yaml
+++ b/configs/user/daveey/eval.yaml
@@ -1,0 +1,16 @@
+# @package __global__
+
+run: ${user.name}.local.eval.1
+
+defaults:
+  - /user/sandbox
+
+evaluator:
+  policy:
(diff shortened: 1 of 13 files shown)


sasmith:
test comment for integration

daveey (approved):
looks great, consider adding launch-template.json as well

PR #54, opened 2025-01-24
Evaluation suite

daphnedemekas:


configs/env/mettagrid/maps/triangle.map
```diff
@@ -0,0 +1,11 @@
+WWWWWWWWWWW
```
daveey, line 1: i think these already got merged in?

rl/pufferlib/eval_suite.py
```diff
@@ -0,0 +1,82 @@
...
+        npcs = eval_item.get("npcs", 0)
+
+        # 1) Create a working copy of the base config
+        #    (which currently includes the base evaluator config).
+        eval_cfg = copy.deepcopy(cfg)
+
+        # 2) Merge environment config into our copy
+        #    We load the environment override from /envs/<env_name>.yaml
+        env_override = hydra.compose(config_name=f"{env_name}.yaml", 
+                                     overrides=[f"env.game.num_agents={4 + npcs}"])
```
daveey, line 45: why 4+ ?

rl/pufferlib/eval_suite.py
```diff
@@ -0,0 +1,82 @@
...
+        eval_cfg.baselines.uri = cfg.baseline_uri
+
+        logger.info(f"--- Running evaluation for environment: {env_name} ---")
+        logger.info(f"Metrics to log: {metrics_to_log}")
+        logger.info(f"Merged config:\n{OmegaConf.to_yaml(eval_cfg, resolve=True)}")
+
+        # 4) Instantiate the evaluator
+        #    (If PufferEvaluator only needs (eval_cfg, policy_record, baseline_records),
+        #     you'll need to create the policy/baseline records. Here's a simplified placeholder.)
+        dummy_policy_record = ...  # e.g. your PolicyRecord from policy_store
```
daveey, line 64: i'm confused, is this example code?

tools/eval_suite.py
```diff
@@ -0,0 +1,47 @@
...
+        self.cfg["evaluator"]["policy"]["uri"] = cfg.eval_suite.policy_uri
+        #this should be a list of uris
+        self.cfg["evaluator"]["baselines"]["uri"] = cfg.eval_suite.baseline_uri 
+        self.evals = cfg.eval_suite.evals
+    
+    def run_all(self):
+        for eval in self.evals:
+            current_cfg = self.cfg.copy()
+
+            env_name = eval["env"]  # e.g. 'env/mettagrid/cylinder'
```
daveey, line 24: .env

util/stats_library.py
```diff
@@ -6,13 +6,17 @@
 from scipy.stats import mannwhitneyu
 from tabulate import tabulate
 from termcolor import colored
+from collections import defaultdict
+
 
 
 class StatisticalTest:
-    def __init__(self, data, categories: List[str]):
+    def __init__(self, data, categories: List[str], mode = 'sum', env_name: str = None):
```
daveey, line 14: env_name doesn't belong in a stats package. you can call it "label" if you want

daveey (changes requested):

daphnedemekas:
Note:

Upon trying to add in NPCs, I realized that the way I was doing it was plugging in baselines into the Evaluator class, which treats those baselines as NPCs.

I changed it now, so that we actually first iterate over the baselines, run the evaluator on the baselines (as candidates) against the NPCs, and then we run the candidate against the NPCs. 

This means we get a list of stats per baseline + stats per policy. 

I'm currently just storing those but haven't written any code to aggregate them together. 

I am wondering:
1) whether the way it's written makes sense intuitively - I feel like the evaluator may be doing unecessary work with the NPCs so want to double check and 
2) whether we should change some of the variable names (like maybe in evaluator, what is called baselines should be called NPCs) 
3) whether we should change some of the logging code or whether I should take it as it is and write some aggregation to be able to have all the baselines and the candidate in a single logged table

@daveey

agent/policy_store.py
```diff
@@ -48,12 +48,15 @@ def local_path(self):
...
 
 class PolicyStore:
-    def __init__(self, cfg: OmegaConf, wandb_run: wandb_run.Run):
+    def __init__(self, cfg: OmegaConf, wandb_run: wandb_run.Run, log: bool = True):
         self._cfg = cfg
         self._device = cfg.device
         self._wandb_run = wandb_run
         self._cached_prs = {}
 
+        if not log:
```
daveey, line 57: i don't think we should do this. if we want to turn off loging for the policy store, we can disable it via setting the loglevel for the logger in main.

configs/config.yaml
```diff
@@ -16,6 +16,7 @@ defaults:
   - evaluator: puffer
   - sweep: sweep
   - hardware: macbook
+  - eval_suite: null
```
daveey, line 19: don't need a default suite

configs/eval_suite/eval_suite_1.yaml
```diff
@@ -0,0 +1,24 @@
+defaults:
+  - ../evaluator/evaluator.yaml
+
+policy_uri: "wandb://run/v.daveey.species.sm.0"
+baseline_uri: ["wandb://run/v.daveey.species.sm.0", "wandb://run/v.daveey.species.sm.0"]
```
daveey, line 5: seems like we don't actually need baseline_uris here?

configs/eval_suite/eval_suite_1.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - env: env/mettagrid/maze
+    metrics_to_log:
+      - time_to_targets
+      - elo
+    npc_policy_uri: null
+
+  - env: env/mettagrid/simple
+    metrics_to_log:
+      - elo
+      - gleicko
```
daveey, line 23: typo?

configs/eval_suite/eval_suite_1.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+policy_uri: "wandb://run/v.daveey.species.sm.0"
+baseline_uri: ["wandb://run/v.daveey.species.sm.0", "wandb://run/v.daveey.species.sm.0"]
+
+evals:
+  - env: env/mettagrid/cylinder
+    metrics_to_log:
+      - time_to_targets
+      - elo
+    npc_policy_uri: null
```
daveey, line 12: best to avoid specifying and defaulting to null

rl/pufferlib/evaluator.py
```diff
@@ -14,19 +14,23 @@ def __init__(
         cfg: OmegaConf,
         policy_record: PolicyRecord,
         baseline_records: List[PolicyRecord],
+        log: bool = True,
         **kwargs
     ) -> None:
 
+        if not log:
```
daveey, line 21: nope

util/stats_library.py
```diff
@@ -53,6 +82,9 @@ def withHistoricalData(self, prior_scores: Dict[str, Any]):
         return self
 
 class MannWhitneyUTest(StatisticalTest):
+    def __init__(self, data, categories: List[str], mode = 'sum', label: str = None):
```
daveey, line 85: why?

daveey (changes requested):
Let's pair on the Eval / EvalSuite logic

daphnedemekas:
This new version has the analysis code using a custom wanduckdb class


Still need to figure out what to do in train and sweep.. did we decide that we were going to log stuff? Or just save stuff and do it after the fact. @daveey

configs/analyzer/analyzer.yaml
```diff
@@ -0,0 +1,50 @@
+wandb:
```
daveey, line 1: can we do db_uri: wandb://artifact-name

and get everything else from the wandb context?

configs/analyzer/analyzer.yaml
```diff
@@ -0,0 +1,50 @@
...
+  entity: metta-research
+  project: metta
+  artifact_name: daphne_eval
+  table_name: eval_stats
+
+output_dir: ${hydra:runtime.output_dir}
```
daveey, line 7: i think we want to use ${run_dir}

configs/analyzer/example_complex.yaml
```diff
@@ -0,0 +1,81 @@
+wandb:
```
daveey, line 1: should this get defaults from analyzer?

configs/env/mettagrid/explore_exploit.yaml
```diff
@@ -0,0 +1,29 @@
+defaults:
```
daveey, line 1: i think you want to merge main into this branch? i think this was already merged?

rl/carbs/rollout.py
```diff
@@ -109,69 +109,71 @@ def _run(self):
         logger.info(f"Evaluating policy {final_pr.name} against {initial_pr.name}")
         self._log_file("eval_config.yaml", eval_cfg)
 
-        evaluator = hydra.utils.instantiate(eval_cfg.evaluator, eval_cfg, final_pr, [initial_pr])
-        stats = evaluator.evaluate()
-        evaluator.close()
+        eval = hydra.utils.instantiate(eval_cfg.eval, policy_store, eval_cfg.env, _recursive_ = False)
+        stats = eval.evaluate()
+        eval.close()
```
daveey, line 114: we should just call close inside evaluate()

rl/carbs/rollout.py
```diff
@@ -109,69 +109,71 @@ def _run(self):
...
+        #     EloTest(stats, eval_cfg.evaluator.stat_categories[eval_cfg.sweep.metric]),
+        #     eval_cfg.evaluator.baselines.elo_scores_path)
+        # logger.info("\n" + fr)
+        # self._log_file("sweep_elo.yaml", r)
+        # test_results["elo"]["initial"] = r[initial_pr.name]["score"]
+        # test_results["elo"]["final"] = r[final_pr.name]["score"]
+
+        # eval_metric = test_results[self.cfg.sweep.test]["final"]
+
+        # stats_update = {
```
daveey, line 146: this should stay

daveey: we no-longer have elo/glicko scores, but we should still save the rest of the metrics.

rl/carbs/rollout.py
```diff
@@ -180,8 +182,8 @@ def _run(self):
         # )
 
         total_time = time.time() - start_time
-        logger.info(f"Carbs Observation: {eval_metric}, {total_time}")
-        self.carbs.record_observation(eval_metric, total_time)
+        # logger.info(f"Carbs Observation: {eval_metric}, {total_time}")
```
daveey, line 185: just set eval_metric to 0 up top, and leave this in

rl/wandb/wanduckdb.py
```diff
@@ -0,0 +1,236 @@
+import os
+import json
+import duckdb
+import wandb
+import pandas as pd
+from typing import Optional, Dict, Any, List
+from omegaconf import OmegaConf
+
+class WandbDuckDB:
+    def __init__(self, entity: str, project: str, artifact_name: str, alias: Optional[str] = None, table_name: str = "artifact_table"):
```
daveey, line 10: this can take a wandb.Run and use that instead of creating it's own connection. look at policy_store

tools/eval.py
```diff
@@ -1,34 +1,151 @@
...
+            cfg.eval,
+            policy_store,
+            cfg.env,
+            cfg.run_dir,
+            _recursive_=False
+        )
+        stats = eval.evaluate()
+
+        # Build additional fields that you want to inject into each record.
+        additional_fields = {}
```
daveey, line 110: this needs to happen any time we log stats, not just when we run eval.py. You should make an EvalStatsLogger(eval) that has a .log(stats) method. And then subclass it to make WandbEvalStatsLogger(eval, artifact) and FileEvalStatsLogger(eval, path)

tools/test.py
```diff
@@ -0,0 +1,99 @@
+#%%
```
daveey, line 1: what's this?

daveey: do you want this in tools/notebooks/daphne.py?

rl/wandb/wanduckdb.py
```diff
@@ -0,0 +1,236 @@
+import os
+import json
+import duckdb
+import wandb
+import pandas as pd
+from typing import Optional, Dict, Any, List
+from omegaconf import OmegaConf
+
+class WandbDuckDB:
```
daveey, line 9: let's go over this one together

daveey (changes requested):

configs/analyzer/analyzer.yaml
daphnedemekas:
```
artifact_name: daphne_eval
table_name: eval_stats
```


```
        self.wandb_db = WandbDuckDB(
            self.cfg.wandb.entity,
            self.cfg.wandb.project,
            self.analyzer_cfg.artifact_name,
            table_name=self.analyzer_cfg.table_name
        )
```

How is that?

daphnedemekas: nevermind, i see what you mean , using the WandbContext class .not seeing a db_uri class variable though, let's talk about how to do this properly

tools/eval.py
daphnedemekas: Or maybe just one class that has log_to_wandb and log_to_file functions?

daphnedemekas:
We now have 

`EvalStatsDB`

with subclasses 

`EvalStatsDbFile` and `EvalStatsDbWandb`

then we have an `example file_analyzer.yaml` and artifact_analyzer.yaml which analyze based on a file path and an artifact name respectively. 

Analysis can:

- pattern match metrics (for example agent.altar.use) to log the average usage of each of these patterned matched metrics 
- take in metrics to log average per episode (averaged over policy)
- take in metrics to log  average overall (averaged over policy and episode)
- take in global filters to filter the entire database 
- each section can take in local filters to additionally filter specific metric logging
- run statistical tests (MannWhitney, Elo, Glicko2) also given filters 

We also have `example_complex.yaml` to showcase a more complicated analysis config

Currently, we aren't doing any console logging of stats in trainer.py or rollout.py, we are simply logging to file and/or wandb artifact. 
Could add some code here to match analyze.py if we want to log stuff to console during training and sweeping

@daveey

daphnedemekas:
TODO

- change db classes to take in uris
- change the analyzer to only take in metrics and group by and glob
- - automatically add significance tests in the analysis

daphnedemekas:
Screenshot 2025-02-11 at 11 06 20 PM
Screenshot 2025-02-11 at 11 06 33 PM

daphnedemekas:
@daveey 
Things should look good now except for rollout.py, still need to figure out how to integrate the new analysis there.

daphnedemekas:
Now added analysis in both `trainer.py` and `rollout.py `
Default analysis will just be the heart altar use metric as well as the significance results (which will use all saved policy checkpoints so far in training) 

For example in training we log this to console: 

Screenshot 2025-02-12 at 12 37 48 PM

And in sweep:
Screenshot 2025-02-12 at 3 07 19 PM

Screenshot 2025-02-12 at 3 07 33 PM
Screenshot 2025-02-12 at 3 07 42 PM

rl/eval/eval_stats_analyzer.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+    def log_result(self, result, metric, filters, significance):
+        result_table = tabulate(result, headers=["policy_name"] + list(result.keys()), tablefmt="grid", maxcolwidths=25)
+        logger.info(f"Results for {metric} with filters {filters}:\n{result_table}")
+        if len(significance) > 0:
+            self.log_significance(significance, metric, filters)
+
+    @staticmethod
+    def _significance_test(metrics_df: pd.DataFrame, metric_name: str) -> List[List[str]]:
```
daveey, line 42: can we put this in stats.py

rl/eval/eval_stats_analyzer.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def _analyze_metrics(self, metric_fields: List[str], filters: Optional[Dict[str, Any]] = None, group_by_episode: bool = False) -> pd.DataFrame:
+        result_dfs = []
+        significance_results = []
+        for metric in metric_fields:
+            if not metric in self.stats_db.available_metrics:
+                logger.info(f"Metric {metric} not found in stats_db")
+                continue
+            df_per_episode = self.stats_db._metric(metric, filters)
+            if not group_by_episode:
```
daveey, line 100: can that go into stats_db._metric(metric, filters, group_by_episode)

rl/eval/eval_stats_analyzer.py
```diff
@@ -0,0 +1,146 @@
...
+
+        metrics_df = pd.concat(result_dfs, axis=1) if len(result_dfs) > 0 else None
+
+        return metrics_df, significance_results
+
+    def analyze(self):
+        metric_configs = {}
+        for cfg in self.analysis.metrics:
+            metric = cfg.metric
+            if '*' in metric or '?' in metric:
```
daveey, line 123: do you need the if? or will fnmatch work if there are no */?

daveey (changes requested):
getting close!

daveey:
Also, do we need 2 separate reports? would be nice if we included mean, stdev, p-val, effect size, etc.. all in one table

configs/eval/eval.yaml
```diff
@@ -0,0 +1,14 @@
...
+policy_agents_pct: 1.0
+
+device: ${device}
+vectorization: ${vectorization}
+
+num_envs: 128
+num_episodes: 128
+max_time_s: 60
+
+eval_artifact_name: null  #if not null, will log to wandb
```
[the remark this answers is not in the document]
daphnedemekas: But do I still always want to be saving locally? Or only if a file path is specified? And what if I want both to save locally and to wandb?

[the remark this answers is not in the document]
daphnedemekas: And what if I am doing eval_suite, in which case I will want to log to multiple files, see here:

```
    def log(self, eval_stats, file_name: str, artifact_name: str = None):

        if isinstance(eval_stats, dict):
            logger.info(f"Evaluation suite: logging {len(eval_stats)} evaluations")
            for eval_name, stats in eval_stats.items():
                self.log(stats, file_name=f"{file_name}_{eval_name}", artifact_name=artifact_name)
            return

        eval_stats = self._add_additional_fields(eval_stats)

        json_path = self._log_to_file(eval_stats, file_name)

        if artifact_name is not None:
            self._log_to_wandb(artifact_name, json_path)

        return eval_stats
```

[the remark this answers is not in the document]
daphnedemekas: I could do like this in eval.py 

```
        artifact_name = None
        if cfg.eval.eval_db_uri.startswith("wandb://"):
            artifact_name = cfg.eval.eval_db_uri.split("wandb://")[1]
            file_path = Path(cfg.eval.policy_uri).name
        elif cfg.eval.eval_db_uri.startswith("file://"):
            file_path = cfg.eval.eval_db_uri.split("file://")[1]
        else:
            file_path = cfg.eval.eval_db_uri

        stats_logger.log(stats, file_name=file_path, artifact_name=artifact_name)
```

daphnedemekas:
@daveey I have two reports because the statistical tests are pairwise (one p-value for two policies) whereas the other stats are per-policy

rl/eval/eval_stats_db.py
```diff
@@ -0,0 +1,122 @@
...
+            data = self._prepare_data(data)
+            all_records.extend(data)
+
+        super().__init__(pd.DataFrame(all_records))
+
+class EvalStatsDbWandb(EvalStatsDB):
+    """
+    Database for loading eval stats from wandb.
+    """
+    def __init__(self, artifact_name: str, wandb_run):
```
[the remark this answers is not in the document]
daphnedemekas: Yes cacheing is now owrking

rl/eval/eval_stats_logger.py
```diff
@@ -0,0 +1,65 @@
...
+        artifact = wandb.Artifact(name=artifact_name, type=artifact_name)
+        artifact.add_file(json_path)
+        artifact.save()
+        artifact.wait()
+        self._wandb_run.log_artifact(artifact)
+        logger.info(f"Logged artifact {artifact_name} to wandb")
+
+    def log(self, eval_stats, file_name: str, artifact_name: str = None):
+
+        if isinstance(eval_stats, dict):
```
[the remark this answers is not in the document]
daphnedemekas: it's only a dict for EvalSuite, otherwise its a list

rl/pufferlib/eval.py
```diff
@@ -0,0 +1,203 @@
...
+        self._num_envs = num_envs
+        self._min_episodes = num_episodes
+        self._max_time_s = max_time_s
+
+
+        policy_selector_cfg = OmegaConf.create({
+            "uri": self._policy_uri,
+            "type": "top",
+            "range": 1,
+            "metric": "elo",
```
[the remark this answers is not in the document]
daphnedemekas: I can't see what you're referring to here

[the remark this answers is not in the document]
daphnedemekas: num_epochs instead of num_episodes?

rl/eval/eval_stats_db.py
```diff
@@ -0,0 +1,125 @@
...
+        self.api = wandb.Api()
+        self.artifact_identifier = os.path.join(wandb_run.entity, wandb_run.project, artifact_name)
+
+        cache_dir = os.path.join(wandb.env.get_cache_dir(), "artifacts", self.artifact_identifier)
+        if os.path.exists(cache_dir):
```
daveey, line 106: if cache_dir exists, this will not download any artifact versions not in the cache, right?

rl/eval/eval_stats_logger.py
```diff
@@ -0,0 +1,87 @@
...
+            json_path = os.path.join(cfg.run_dir, "eval_stats")
+        elif cfg.eval.eval_db_uri.startswith("file://"):
+            json_path = cfg.eval.eval_db_uri.split("file://")[1]
+        else:
+            if "://" in cfg.eval.eval_db_uri:
+                raise ValueError(f"Invalid eval_db_uri: {cfg.eval.eval_db_uri}")
+            json_path = cfg.eval.eval_db_uri
+
+        self._json_path = json_path if json_path.endswith('.json') else json_path + '.json'
+        os.makedirs(os.path.dirname(self._json_path), exist_ok=True)
```
daveey, line 28: are you making a dir named xxx.json? that seems wrong?

rl/eval/eval_stats_logger.py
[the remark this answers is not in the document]
daveey: you can just do:

if isintance(eval_stats, dict):
  eval_stats = [stat for stats in eval_stats.values() for stat in stats]

then you don't need a recursive call

rl/eval/eval_stats_logger.py
```diff
@@ -0,0 +1,87 @@
...
+            all_stats = []
+
+            for eval_name, stats in eval_stats.items():
+                all_stats.extend(stats)
+
+            self.log(all_stats)
+
+            return all_stats
+
+        eval_stats = self._add_additional_fields(eval_stats)
```
daveey, line 80: does it need to return eval_stats, it modifies them anyway

daveey (changes requested):

rl/eval/eval_stats_logger.py
daphnedemekas: No, it takes the dirname
os.path.dirname(self._json_path)

rl/eval/eval_stats_db.py
daphnedemekas: Yes, this is probably what we want:

```
        for v,a in enumerate(artifact_versions):
            if not os.path.exists(os.path.join(cache_dir, f"v{v}")):
                a.download(root=os.path.join(cache_dir, f"v{v}"))
            artifact_dirs.append(os.path.join(cache_dir, f"v{v}"))
```

daphnedemekas: (and i changed it to that)

daveey (approved):
fix tests

PR #58, opened 2025-01-26
Update configs for cylinder and ascii

daphnedemekas:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1209230969804461/1209243457565709) by [Unito](https://www.unito.io)


configs/env/mettagrid/ascii.yaml
```diff
@@ -0,0 +1,35 @@
+defaults:
```
daveey, line 1: Is this a specific eval? In that case, let’s give it a real name, not “ascii.yaml”. If it’s a test or example, let’s put it into examples

configs/env/mettagrid/cylinder.yaml
```diff
@@ -0,0 +1,40 @@
...
+      border_width: 1
+      width: ${make_odd:${uniform:10,40,21}}
+      height: ${make_odd:${uniform:10,40,21}}
+      cylinder_params:
+        length: ${uniform:10,40,21}
+        orientation: 'horizontal'
+        both_ends: true
+      
+
+  objects:
```
daveey, line 23: Are these the settings we want for the cylinder test? If so, great, just wanted to make sure

daveey (changes requested):

configs/env/mettagrid/cylinder.yaml
daphnedemekas: ideally, the orientation would be like orientation: ${choice:['horizontal','vertical']} in the same way that you have the uniform sampling. and for cylinder both_ends we could also have those be randomly sampled
But that doesn't seem to be compatible with omegaconf or something, i can't seem to figure out how to do that, do you know how?

configs/env/mettagrid/ascii.yaml
daphnedemekas: The triangle ascii map is a specific eval which this file references as uri, but I named it ascii because we can specify any ascii map here and keep it general if we wanted to, but I can see how that would make more sense in an examples dir. Do we have one? I can't find one but happy to make. I changed this to triangle.yaml for now, it is a super basic one, maybe unnecessary to have explicitly

daphnedemekas:
This PR should only be merged once we have merged in https://github.com/Metta-AI/mettagrid/pull/34 for mettagrid due to some filename changes

configs/env/mettagrid/ascii.yaml
daveey: feel free to make an examples/ dir

configs/env/mettagrid/cylinder.yaml
daveey: this comment was refering to the "objects" clause, since that's where game rules are set. 

for orientation, it seems like you can just use "horizontal: true | false", which would let us sweep over int=[0|1]

daveey (changes requested):

configs/env/mettagrid/cylinder.yaml
daphnedemekas: Oh right yeah I see, I had just copied those over from the mazes. I removed it so we can have them be default for now, since this test is just about locating the generator. can think more about if we want to specialize the objects later on. And OK i changed to bools

daveey (approved):

PR #59, opened 2025-01-28
Alex regularize 2

Al-does:
- Effective rank tracking
- L2 norm loss
- L2 norm to initial weight params loss
- Weight clipping
- Moved actor network creation to metta_agent.py
- Moved some network initializations to WeightTransformer to coordinate defining the gain and weight clipping limit as both of those are a function of the network's nonlinear layer and its size, respectively.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1209230969804461/1209258472504105) by [Unito](https://www.unito.io)


daveey (changes requested):
I assume this is old and now part of your refactor

PR #65, opened 2025-02-12
New configs for cogeval training

georgejwdeane:
This depends on the mettagrid branch:
https://github.com/Metta-AI/mettagrid/pull/42

and it just contains yaml files to be able to run new cognitive eval maps


daphnedemekas:
@georgejwdeane can you please move all of the `cogeval` configs inside of a `cognitive_eval/` directory in `env/mettagrid `
and then make sure you change the corresponding paths in your user config to have that run correctly.

daphnedemekas (changes requested):
Can you please move all of the cogeval configs inside of a` cognitive_eval/ `directory in `env/mettagrid ` and then make sure you change the corresponding paths in your user config to have that run correctly.

Also, once you make the changes on your other PR in mettagrid, to move those python fiels into a cognitive_evals directory, the paths within each of your yaml files here will need to change in order to accomodate that new directory. 

So first make that change, and then move these yaml files into a new directory and make sure they point to the right python files in __target__
Once that's ready we can merge and that will be nice because we can then train on AWS

PR #67, opened 2025-02-18
Eval suite fixes

daphnedemekas:
These fixes include:

- Making sure the name of each eval in the `EvalSuite` is included in the database so that we can analyze per-eval data rather than averaging across evals (this means we have to run `add_additional_fields` separately, so we have to go back to the for-loop implementation)
- Adding a default env: ??? parameter to eval.yaml so we don't get an error if that's not specified. However (see comments below) we will still get an error if we aren't specifying it somewhere, i.e. in the command line, if we are using eval.yaml as our eval for training, for eg. Also, this change means we need env: null in example_suite since it technically inherits from eval.yaml - but the eval suite envs will actually be specified in evals: . A potential thing to do here is to just migrate over to using eval suite for everything (training, sweeping) so we get around this subtlety. 
- Loads the candidate policy externally for eval, so that in the EvalSuite case, we don't have to reload the policy record for every eval
- Using cache in policy store when loading policies from wandb
- Fixes the `_load_from_uri()` issue in PolicyStore


daveey (commented):
i don't think we should have a default, let's set it to ??? and make evals specify it

daveey (changes requested):
i don't think we should have a default, let's set it to ??? and make evals specify it

daphnedemekas:
Ok I changed it to env: ??? but I'm not entirely sure what you mean by "make evals specify it"

daphnedemekas:
Screenshot 2025-02-19 at 10 57 06 AM

Analysis currently outputs this. Next step (on a separate branch) will be to implement PFS

daveey (approved):

PR #68, opened 2025-02-19
Support new card-based mettagrid

daveey:


daveey (commented):
Updating to match mettagrid repo

PR #70, opened 2025-02-20
Zip files and dont extend data for artifacts

daphnedemekas:


rl/eval/eval_stats_db.py
```diff
@@ -105,15 +106,20 @@ def __init__(self, artifact_name: str, wandb_run):
...
         logger.info(f"Loaded {len(artifact_dirs)} artifacts")
         all_records = []
         for artifact_dir in artifact_dirs:
-            json_files = [f for f in os.listdir(artifact_dir) if f.endswith('.json')]
-            if len(json_files) != 1:
-                raise FileNotFoundError(f"Expected exactly one JSON file in {artifact_dir}, found {len(json_files)}")
-            json_path = os.path.join(artifact_dir, json_files[0])
-            with open(json_path, "r") as f:
-                data = json.load(f)
+            files = [f for f in os.listdir(artifact_dir) if f.endswith('.json') or f.endswith('.json.gz')]
```
daveey, line 113: let's just do this once, when you download it? that way you can always just open .json, and the only time you have to think about gzip is after download and before upload

daveey (changes requested):

daveey (approved):

PR #71, opened 2025-02-21
inherit policy uri

daphnedemekas:
Plus some small print statement changes

### DIFF
--- a/configs/analyzer/analyzer.yaml
+++ b/configs/analyzer/analyzer.yaml
@@ -1,6 +1,6 @@
 _target_: rl.eval.eval_stats_analyzer.EvalStatsAnalyzer
 
-policy_uri: null
+policy_uri: ${policy_uri}
 
 analysis:
 
--- a/configs/analyzer/analyzer.yaml
+++ b/configs/analyzer/analyzer.yaml
@@ -1,6 +1,6 @@
 _target_: rl.eval.eval_stats_analyzer.EvalStatsAnalyzer
 
-policy_uri: ${policy_uri}
+policy_uri: ???
 
 analysis:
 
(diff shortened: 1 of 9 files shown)


configs/analyzer/analyzer.yaml
```diff
@@ -1,6 +1,6 @@
 _target_: rl.eval.eval_stats_analyzer.EvalStatsAnalyzer
 
-policy_uri: null
+policy_uri: ${policy_uri}
```
daveey, line 3: This gets it from global scope, right? Do we set it there?

daphnedemekas: My understanding is that doing this makes it equal to what is defined in config.yaml, which is what we want I believe so that we don't have to specify it again

daphnedemekas: (which was causing the error you sent me, which I didn't get because in daphne.yaml I was specifying it as analysis:
      {...policy_uri} 

I also just changed it in eval.yaml as well since it was set to ??? there but I don't think that would work if we didn't pre-specify it as well with eval.policy_uri (which we probably don't want to have to do) 

Does that make sense?

daphnedemekas: And also, I have learned from this to always run on a clean user file which maps to what we would have when training via command line before merging in!

daphnedemekas:
Okay @daveey it should actaully be good now, for training we don't have to specify anything and we set the policy uri for both eval and analysis in the trainer code:
        self.cfg.eval.policy_uri = self.last_pr.uri
        self.cfg.analyzer.policy_uri = self.last_pr.uri

When running eval, we need to specify a policy_uri in the eval config, but it gets set for analysis in the eval code
cfg.analyzer.policy_uri = policy_pr.uri

And when running analysis, we need to specify a policy_uri in the analysis config. 

Feels like that's what we want

PR #87, opened 2025-03-11
Kickstarting

Al-does:


agent/metta_agent.py
```diff
@@ -210,3 +207,42 @@ def compute_effective_rank(self, delta: float = 0.01) -> List[dict]:
             if rank is not None:
                 effective_ranks.append(rank)
         return effective_ranks
+
+# ------------------From cleanrl.py--------------------------------
+
+    def log_prob(self, logits, value):
```
daveey, line 213: let's put these in agent/util.py, they are just functions and don't need "self"

configs/trainer/puffer.yaml
```diff
@@ -45,3 +44,13 @@ stats:
   overview:
     episode/reward.mean: episode_reward
   step: train/agent_step
+
+kickstart: false
```
daveey, line 48: make kickstart a section

kickstart:
  teachers: []
  kickstart_step_pct: 0.005

if teachers is not empty, assume kickstarting is enabled

rl/pufferlib/kickstarter.py
```diff
@@ -0,0 +1,63 @@
...
+        if not self.enabled:
+            return
+
+        self.teacher_cfgs = cfg.trainer.teachers
+        self.compile = cfg.trainer.compile
+        self.compile_mode = cfg.trainer.compile_mode
+        self.policy_store = policy_store
+        self.kickstart_steps = cfg.trainer.kickstart_steps_pct * cfg.trainer.total_timesteps
+
+        if isinstance(single_action_space, pufferlib.spaces.Discrete):
```
daveey, line 18: only need to deal with multi_discrete

rl/pufferlib/kickstarter.py
```diff
@@ -0,0 +1,63 @@
...
+            policy_record = self.policy_store.policy(teacher_cfg['policy_uri'])
+            policy = policy_record.policy()
+            policy.action_coef = teacher_cfg['action_coef']
+            policy.value_coef = teacher_cfg['value_coef']
+            if self.compile:
+                policy = torch.compile(policy, mode=self.compile_mode)
+            self.teachers.append(policy)
+
+    def loss(self, agent_step, student_normalized_logits, student_value, o, teacher_lstm_state=None):
+        ks_value_loss = torch.tensor(0.0, device=self.device)
```
daveey, line 37: can we create these in the constructor, rather than every time loss() is called?

rl/pufferlib/kickstarter.py
```diff
@@ -0,0 +1,63 @@
...
+        if not self.enabled or agent_step > self.kickstart_steps:
+            return ks_action_loss, ks_value_loss, teacher_lstm_state
+        
+        if teacher_lstm_state is None:
+            teacher_lstm_state = [None for _ in range(len(self.teachers))]
+
+        for i, teacher in enumerate(self.teachers):
+            teacher_value, teacher_normalized_logits, teacher_lstm_state[i] = self._forward(teacher, o, teacher_lstm_state[i])
+
+            if self.multi_discrete:
```
daveey, line 49: only handle multi-discrete. also, write as a loop rather than hardcoding [0] and [1]

for t_logits, s_logits in zip(teacher_normalized_logits, student_normalized_logits):
   ...
   
 also, multiply by action_coef after the loop, to save on a mult

configs/trainer/puffer.yaml
```diff
@@ -45,3 +44,13 @@ stats:
...
+kickstart_steps_pct: 0.005
+teachers:
+  - policy_uri: wandb://run/mettabox_cogeval_defaults_lowent_initialized0005:v100
+    action_coef: 1
```
daveey, line 52: let's call these action_loss_coef, value_loss_coef

daveey (changes requested):
This is really well done! Clean and clear. I left a few small suggestions, but overall, looks great.

rl/pufferlib/kickstarter.py
Al-does: Doing so threw an error on the optimization step in trainer. GPT gave a better answer than I can write:

> The error you're encountering is related to PyTorch's autograd system and how computational graphs are handled.
> When you initialize self.ks_value_loss and self.ks_action_loss as class variables and modify them in-place during the loss() method, PyTorch is adding these operations to its computational graph. Since these tensors persist across multiple calls to loss(), PyTorch tries to backward through the same tensors multiple times, which isn't allowed by default.

Leaving it initialized at each call of loss() for now but lmk.

rl/pufferlib/kickstarter.py
Al-does: (teacher_normalized_logits[i].exp() * student_normalized_logits[i]).sum(dim=-1) is calculating the cross entropy between the entire teacher log prob distribution (converting them to normalized probs via exponentiation) and multiplying by the student's log prob distribution. If i is 0 then it runs this calc on the action param and if i is 1 then it runs on the action type. 

In case you wanted me to loop this over the number of spaces (action type, action param, a third action space in the future, etc....) I made that change.

daveey (approved):

PR #90, opened 2025-03-15
Made CNN/flatten robust to changes in obs window

Al-does:
Previously, the flatten layer assumed the output from the last CNN would be 1x1. This now passes the obs window sizes through all of the CNN layers until we get a final output window. It also checks if your kernel and stride sizes don't map the input width and height by striding an integer number of times.


agent/lib/nn_layer_library.py
```diff
@@ -13,27 +13,42 @@ def _make_net(self):
             **self._nn_params
         )
 
-class Conv1d(ParamLayer):
```
daveey, line 16: Did you mean to remove 1d conv?

daveey (approved):

agent/lib/nn_layer_library.py
Al-does: Yeah... it also needs a method of calculating what its flattened output dimension would be but, not knowing exactly what the specific use cases will look like, figured I'd just leave it to whoever is trying to use Conv1D to implement by copying and modifying the pattern I used for Conv2D. Just thought it might be safer that way, to not assume I know how it will be used. 

The specific issue is that I don't know where to get Conv1D's input dim from. Here, I assume that Conv2D is getting its input dim from obs width and height which is brittle too, unfortch. Figured it's enough for now. I just added a warning comment to Conv2D saying as much.

Al-does: I thought a bit more about this whole update and I think there's a much more general way to improve how shapes are passed between all layers so I'm going to implement that in this branch. Will follow up when that's ready!

agent/lib/merge_layer.py
```diff
@@ -24,13 +23,19 @@ def setup(self, input_source_components=None):
 
         self.input_source_components = input_source_components
 
-        self.sizes = []
+        # self.sizes = []
```
daveey, line 26: clean these up?

daveey (approved):

Al-does:
Fixed this in `alex-flatten-pr`

PR #96, opened 2025-03-20
Daphne colored ore and fixes

daphnedemekas:
Add support for colored ore observations, and make some small changes including zipping all data in evaluation

Tests are currently failing because mettagrid needs to updated, see mettagrid PR 
https://github.com/Metta-AI/mettagrid/pull/59


daveey (approved):
looks like gzip of eval stats got into this as well.

PR #98, opened 2025-03-27
Eval fixes

daphnedemekas:
Fix the evaluation pipeline so that we average overall of the run results for a particular eval and policy_uri when running analysis. 

Still have work to do in terms of improving tooling but I think it would be good to merge this in first


agent/policy_store.py
daphnedemekas, line 1: This is functional but still needs to be improved which I will do in a separate PR
- I think the selector_type "top" is deprecated. I want a nice way of finding the best checkpoint for a particular eval

configs/eval/simple.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - eval
+
+env: null #the env is set in evals:
+
+evals:
+  simple:
+    env: env/mettagrid/simple
+    policy_agents_pct: 1.0
+
+selector_type: latest
```
daveey, line 13: this is confusing, why does an evalsuite need a selector?

rl/eval/eval_stats_analyzer.py
```diff
@@ -105,8 +96,11 @@ def get_latest_policy(all_policies, uri):
         policy_versions = [i for i in all_policies if uri in i]
         if len(policy_versions) == 0:
             raise ValueError(f"No policy found in DB for candidate policy: {uri}, options are {all_policies}")
-        if len(policy_versions) > 1 and 'wandb' in uri:
-            policy_versions.sort(key=lambda x: int(x.split(':v')[-1]))
+        if len(policy_versions) > 1:
+            try:
+                policy_versions.sort(key=lambda x: int(x.split(':v')[-1]))
+            except:
```
daveey, line 102: why would this fail? seems like a weird error to just ignore

rl/eval/eval_stats_analyzer.py
```diff
@@ -136,7 +130,7 @@ def policy_fitness(self, metric_data, metric_name):
             candidate_mean = metric_data.loc[candidate_uri][metric_mean]
             baseline_mean = metric_data.loc[baseline_policies][metric_mean].mean()
             fitness = candidate_mean - baseline_mean # / np.std(baseline_data.loc[eval][metric_std])
-            policy_fitness.append({"eval": eval, "metric": metric_name, "candidate_mean": candidate_mean, "baseline_mean": baseline_mean, "fitness": fitness})
+            policy_fitness.append({"eval": evals[0], "metric": metric_name, "candidate_mean": candidate_mean, "baseline_mean": baseline_mean, "fitness": fitness})
```
daveey, line 133: why is evals[0] special here? shouldn't we log this per-eval?

rl/pufferlib/eval.py
```diff
@@ -136,21 +140,30 @@ def evaluate(self):
...
+            game_cfg = OmegaConf.to_container(self._env_cfg.game, resolve=False)
+            flattened_env = flatten_config(game_cfg, parent_key = "game")
+            flattened_env['run_id'] = self._run_id
+            flattened_env['eval_name'] = self._env_name
+            flattened_env['timestamp'] = datetime.now().isoformat()
+            flattened_env['npc'] = self._npc_policy_uri
+
              # infos is a list of dictionaries
-            if len(infos) > 0:
+            if len(infos) > 0: #fix this
```
daveey, line 152: fixthis

tools/eval.py
```diff
@@ -16,22 +16,28 @@ def main(cfg: DictConfig):
...
-            policy_store,
-            policy_pr,
-            cfg_recursive_=False
-        )
-        stats = eval.evaluate()
-
-        stats_logger = EvalStatsLogger(cfg, eval._env_cfg, wandb_run)
-
-        stats_logger.log(stats)
+        if cfg.eval.selector_type == "all":
```
daveey, line 19: why do we want an "if" here instead of handling it in policy_store.policy()

daveey (changes requested):

configs/eval/simple.yaml
daphnedemekas: Sometimes I want to evaluate all versions of the policy, sometimes just hte latest

rl/eval/eval_stats_analyzer.py
daphnedemekas: If the policies don't have version numbers in them with the format :v then this will fail, and also we probably don't want to sort them. But I can make this clearer

rl/eval/eval_stats_analyzer.py
daphnedemekas: The evals[0] is only under "if len(evals) == 1"

if we have multiple evals then we are doing it below this, because in that case we need to reset the index of the dataframe. just annoying pandas stuff. I can make a note to make this more elegant

tools/eval.py
daphnedemekas: sure, can handle it there. still need to check if it returns a list or not so still need the "if" but can instead do 

```
        policy_prs = policy_store.policy(cfg.eval.policy_uri, cfg.eval.selector_type)
        if not isinstance(policy_prs, list):
            policy_prs = [policy_prs]
```

configs/eval/simple.yaml
daphnedemekas: or are you saying that's part of the policy config instead? Where would that be defined? right now i'm just passing in the uri on its own

daphnedemekas: but i guess i could remove this from here and make people do it via command line

daveey: can that go into configs/eval.yaml

daveey (changes requested):

tools/eval.py
daveey: policy_store.policy() should always return a single PR. use .policies() if you want a list.

configs/analyzer.yaml
```diff
@@ -0,0 +1,17 @@
+defaults:
+  - common
+  - agent: simple
+  - env/mettagrid@env: simple
```
daveey, line 4: i dont think we need this

configs/config.yaml
```diff
@@ -2,9 +2,9 @@ defaults:
   - agent: simple
```
daveey, line 2: i dont think this file is used anymore, let's just delete it instead?

rl/eval/eval_stats_analyzer.py
```diff
@@ -102,12 +93,13 @@ def get_latest_policy(all_policies, uri):
...
             uri = uri.replace("wandb://run/", "")
-        policy_versions = [i for i in all_policies if uri in i]
-        if len(policy_versions) == 0:
+        matching_policies = [i for i in all_policies if uri in i]
+        if len(matching_policies) == 0:
             raise ValueError(f"No policy found in DB for candidate policy: {uri}, options are {all_policies}")
-        if len(policy_versions) > 1 and 'wandb' in uri:
-            policy_versions.sort(key=lambda x: int(x.split(':v')[-1]))
-        candidate_uri = policy_versions[-1]
+        if len(matching_policies) > 1:
```
daveey, line 99: i dont think we need the "if" here, right?

rl/eval/eval_stats_analyzer.py
```diff
@@ -132,25 +124,17 @@ def policy_fitness(self, metric_data, metric_name):
...
-            fitness = candidate_mean - baseline_mean # / np.std(baseline_data.loc[eval][metric_std])
-            policy_fitness.append({"eval": eval, "metric": metric_name, "candidate_mean": candidate_mean, "baseline_mean": baseline_mean, "fitness": fitness})
-            return policy_fitness
-
-
-        candidate_data = pd.DataFrame(metric_data.loc[candidate_uri]).set_index(eval)
-        baseline_data = metric_data.loc[baseline_policies].set_index(eval)
-
-        evals = metric_data[eval].unique()
+        if len(evals) > 1:
```
daveey, line 127: can we just always do this? or does this break if len(evals) == 1?

rl/pufferlib/trainer.py
```diff
@@ -21,7 +21,7 @@
 from rl.pufferlib.experience import Experience
 from rl.pufferlib.kickstarter import Kickstarter
 from rl.pufferlib.profile import Profile
-from rl.pufferlib.trace import save_trace_image
+# from rl.pufferlib.trace import save_trace_image
```
daveey, line 24: what's this? if we don't use it, just remove the line?

rl/pufferlib/trainer.py
```diff
@@ -223,7 +223,9 @@ def _evaluate(self):
             lstm_h, lstm_c = experience.lstm_h, experience.lstm_c
             e3b_inv = experience.e3b_inv
 
-        while not experience.full:
+        start = time.time()
+
+        while not experience.full and time.time() - start < 100:
```
daveey, line 228: what's this? won't this break training?

daveey (changes requested):

configs/analyzer.yaml
daphnedemekas: without it, if i try to run analyze on vscode at least, I get 

`Cannot find primary config 'analyzer'. Check that it's in your config search path.
`

rl/eval/eval_stats_analyzer.py
daphnedemekas: It breaks if len(evals) == 1

rl/pufferlib/trainer.py
daphnedemekas: Sorry I missed this. Was a temporary hack

daveey (approved):

PR #99, opened 2025-03-27
Add trace test to github actions.

treeform:


configs/trace.yaml
```diff
@@ -0,0 +1,20 @@
+defaults:
+  - common
+  - agent: simple
+  - trainer: puffer
+  - env/mettagrid@env: simple
```
daveey, line 5: i dont think we need this one

tools/trace.py
```diff
@@ -2,29 +2,30 @@
...
-from util.runtime_configuration import setup_metta_environment
 from agent.policy_store import PolicyStore
+from rl.wandb.wandb_context import WandbContext
 from rl.pufferlib.trace import save_trace_image
+from util.runtime_configuration import setup_metta_environment
+from util.config import config_from_path
 
-
-@hydra.main(version_base=None, config_path="../configs", config_name="config")
+@hydra.main(version_base=None, config_path="../configs", config_name="trace")
```
daveey, line 14: does trace need it's own config, or can we just use "eval" the way we do in tools/play.py?

daveey (approved):

configs/trace.yaml
treeform: eval.yaml has some stuff that appear to be problematic with trace.
```
cmd: eval
eval_db_uri: null
run: ???
data_dir: ./train_dir
run_dir: ${data_dir}/${run}
```

Trace should not be confined to `cmd: eval` or `data_dir: ./train_dir`.

I am not sure how to make it work?

tools/trace.py
treeform: I could not get the eval config to work. Maybe you can show me how?

treeform: I made it work with some changes.

configs/eval.yaml
```diff
@@ -15,3 +15,13 @@ run_dir: ${data_dir}/${run}
 
 torch_deterministic: true
 vectorization: multiprocessing
+
+policy_uri: ./train_dir/${.run}/checkpoints/
```
daveey, line 19: why are you assuming the policy is in the current run? 

i think it's better to set policy_uri: ??? and then pass it on the command line

configs/eval.yaml
```diff
@@ -15,3 +15,13 @@ run_dir: ${data_dir}/${run}
...
+
+eval:
+  env: env/mettagrid/simple
```
daveey, line 22: do you need to override the defaults with this? we already pull eval_suite as the default

daveey (changes requested):

configs/eval.yaml
treeform: I am just following the pattern train.yaml has? 

If `python -m tools.train +hardware=github` works `python -m tools.trace +hardware=github` should also work?

configs/eval.yaml
treeform: It appears so. Does not work without it: 

```
 /Users/me/p/metta/util/config.py:6 in config_from_path                                    │
│                                                                                           │
│    3                                                                                      │
│    4 def config_from_path(config_path: str, overrides: DictConfig = None) -> DictConfig:  │
│    5 │   env_cfg = hydra.compose(config_name=config_path)                                 │
│ ❱  6 │   if config_path.startswith("/"):                                                  │
│    7 │   │   config_path = config_path[1:]                                                │
│    8 │   path = config_path.split("/")                                                    │
│    9 │   for p in path[:-1]:                                                              │
╰───────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: 'NoneType' object has no attribute 'startswith'```

File: configs/eval.yaml (hunk 3 above)
Comment (treeform) [reply configs/eval.yaml:19]: I am fine with doing:
`python -m tools.train +hardware=github --ouput:some_path` 
Then 
`python -m tools.trace +hardware=github --input:some_path`
But it should be in both places?

File: configs/eval.yaml (hunk 4 above)
Comment (daveey) [reply configs/eval.yaml:22]: can you figure out the issue?

Comment (treeform) [reply configs/eval.yaml:22]: Yes it fails because `eval_suite` does not actually set the env
```
env: null #the env is set in evals:
```
I don't understand why its configured this way.

So I have to bring my own env.

Comment (treeform) [reply configs/eval.yaml]: Talked via chat.

File: configs/eval.yaml (hunk 3 above)
Comment (treeform) [reply configs/eval.yaml]: Talked on discord.

File: configs/simulator.yaml [hunk 5]
```diff
@@ -0,0 +1,7 @@
+defaults:
+  - common
+  - wandb
+
+policy_uri: ./train_dir/${.run}/checkpoints/
```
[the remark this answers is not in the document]
treeform: Done

PR #113, opened 2025-04-03
Training suite

daphnedemekas:


rl/pufferlib/vecenv.py
```diff
@@ -30,8 +32,13 @@ def make_vecenv(
     else:
         raise ValueError('Invalid --vector (serial/multiprocessing/ray).')
 
+    if hasattr(env_cfg, 'envs'):
```
daveey, line 35: we shouldn't need to touch this

rl/pufferlib/vecenv.py
```diff
@@ -5,9 +5,11 @@
 import pufferlib.utils
 import pufferlib.vector
 import hydra
-from mettagrid.mettagrid_env import MettaGridEnv
+from mettagrid.mettagrid_env import MettaGridEnv, MettaGridEnvSet
 
 def make_env_func(cfg: DictConfig, buf=None, render_mode='rgb_array'):
+    if hasattr(cfg, 'envs'):
```
daveey, line 11: can we just use hydra.instanciate here, and set the correct _target_ in the yaml?

configs/env/mettagrid/multienv_mettagrid.yaml
```diff
@@ -0,0 +1,15 @@
+_target_: mettagrid.mettagrid_env.MettaGridEnvSet
+
+defaults:
```
daveey, line 3: do we need to inherit mettagrid here?

daveey (changes requested):

configs/env/mettagrid/multienv_mettagrid.yaml
daphnedemekas: I removed that and added an error handling when parsing env overrides because if we specify env overrides and the environment is multienv that will crash

daphnedemekas:
Tests will fail until https://github.com/Metta-AI/mettagrid/pull/66 is merged

util/config.py
```diff
@@ -9,5 +12,8 @@ def config_from_path(config_path: str, overrides: DictConfig = None) -> DictConf
     for p in path[:-1]:
         env_cfg = env_cfg[p]
     if overrides is not None:
-        env_cfg = OmegaConf.merge(env_cfg, overrides)
+        try:
+            env_cfg = OmegaConf.merge(env_cfg, overrides)
+        except ConfigKeyError:
+            logger.warning(f"Cannot parse overrides when using multienv_mettagrid")
```
sasmith, line 18: Similar to my other review, I expect we'd want to throw here, since `env_cfg` will be in some unknown state afterward.

rl/pufferlib/vecenv.py
daphnedemekas, line 1: This code change is for some reason breaking evaluation at the moment. 

I am getting 

hydra.errors.InstantiationException: Error in call to target 'rl.pufferlib.eval.EvalSuite':
UnsupportedInterpolationType('Unsupported interpolation type uniform')
full_key: eval

I am not sure if this is due to Jack's new changes or something else, it's another weird hydra error. I've tried to debug it, but I can't seem to understand it, the config does have $uniform in it, but the exact same code gets called in play with a similar structure and parses fine

rl/pufferlib/vecenv.py
```diff
@@ -40,4 +40,4 @@ def make_vecenv(
     )
 
     vecenv = pufferlib.vector.make(make_env_func, **vecenv_args)
-    return vecenv
+    return vecenv
```
sasmith, line 43: If you turn on `files.insertFinalNewline`, cursor will ensure there's a newline when you save. Generally this is what we want, because tradition.

rl/pufferlib/vecenv.py
```diff
@@ -31,7 +31,7 @@ def make_vecenv(
         raise ValueError('Invalid --vector (serial/multiprocessing/ray).')
 
     vecenv_args = dict(
-        env_kwargs=dict(cfg = dict(**env_cfg), render_mode=render_mode),
+        env_kwargs= dict(cfg = env_cfg, render_mode=render_mode),
```
sasmith, line 34: Petty, but prefer no space after the `=` when setting parameter values like this. It's a small style thing, but consistency makes it easier to read.

https://peps.python.org/pep-0008/#other-recommendations

configs/env/mettagrid/multienv_mettagrid.yaml
```diff
@@ -0,0 +1,11 @@
+_target_: mettagrid.mettagrid_env.MettaGridEnvSet
```
sasmith, line 1: Oh this is neat, and is up the alley of something @jacklionheart was looking for. It also helps me understand why `MettaGridEnvSet` is structured like it is.

sasmith (changes requested):
Requesting changes based on the `maybe throw an exception` piece.

But I want to chat about the right way to build configs. I'm not sure it makes sense to make config objects that have _implicit_ randomness.

util/config.py
```diff
@@ -1,5 +1,7 @@
 import hydra
 from omegaconf import DictConfig, OmegaConf
+import logging
+logger = logging.getLogger(__name__)
```
sasmith, line 4: This should be removable.

sasmith (approved):
Please address the (minor) changes before merging.

PR #114, opened 2025-04-03
Fix setup_build.sh to checkout mettagrid with git pull.

treeform:
Setup_build was changed to only update mettagrid when `METTAGRID_REF` was set. Changing it always update, even when METTAGRID_REF is not set. Just use main branch then.

### DIFF
--- a/devops/setup_build.sh
+++ b/devops/setup_build.sh
@@ -37,6 +37,11 @@ git fetch
 if [ -n "$METTAGRID_REF" ]; then
   echo "Checking out mettagrid reference: $METTAGRID_REF"
   git checkout "$METTAGRID_REF"
+  git pull origin "$METTAGRID_REF"
+else
+  echo "No mettagrid reference specified, checking out main"
+  git checkout main
+  git pull
 fi
 
 echo "Installing mettagrid in to $(pwd)"


daveey (changes requested):
I don't think we want this. people run setup_build while having changes in their mettagrid branch

treeform:
I made this script (thats easy to copy paste), that nukes everything in deps and rebuilds for any one that finds this PR:

```
(
set -e
pip install -r requirements.txt

rm -rf deps
mkdir -p deps
cd deps

# fast_gae
git clone --depth 1 https://github.com/Metta-AI/fast_gae.git
cd fast_gae
python setup.py build_ext --inplace
pip install -e .
cd ..

# pufferlib (explicitly clone 'metta' branch)
git clone --depth 1 --branch metta https://github.com/Metta-AI/pufferlib.git
cd pufferlib
pip install -e .
cd ..

# mettagrid (clone main branch)
git clone --depth 1 --branch main https://github.com/Metta-AI/mettagrid.git
cd mettagrid
python setup.py build_ext --inplace
pip install -e .
cd ..

# carbs
git clone --depth 1 https://github.com/kywch/carbs.git
cd carbs
pip install -e .
cd ..

# wandb_carbs
git clone --depth 1 https://github.com/Metta-AI/wandb_carbs.git
cd wandb_carbs
pip install -e .
cd ..
)
```

PR #116, opened 2025-04-05
Training instructions fixes

berekuk:
- fixed `+hardware`
- refactored wandb configs to provide a `wandb=off` option


configs/wandb/metta_research.yaml
berekuk, line 1: Unsure if `metta_research` or `metta-research` is best, decided to go with the convention of `_` used in other config names.

configs/wandb/off.yaml
berekuk, line 1: Some configs in `configs/user` could use this through `override: /wandb: off`.
I opted to keep them as is, authors can update themselves and this refactor is not a breaking change.

configs/analyzer.yaml
```diff
@@ -4,7 +4,7 @@ defaults:
   - trainer: puffer
   - eval: eval_suite
   - analyzer: eval_analyzer
-  - wandb
+  - wandb: metta_research
```
berekuk, line 7: Maybe these should be moved to `common.yaml`? AFAICT all root configs include wandb.

configs/wandb/off.yaml
```diff
@@ -0,0 +1,4 @@
+name: ${run}
+enabled: false
+track: false
+run_id: null
```
berekuk, line 4: Note: `WandbContext` fails if `name` and `run_id` are not provided, even though they shouldn't be needed.

configs/analyzer.yaml
daveey: Yup

daveey (approved):

daveey:
Assigning to @sasmith to take a look.

configs/wandb/off.yaml
sasmith: Feel free to fix this too!

PR #120, opened 2025-04-08
Update analyze to push heatmap report to s3

jacklionheart:
Also includes some bug fixes that resulted in missing data and re-adding the short name that I accidentally removed when refactoring.


daphnedemekas (approved):
This is great!

PR #123, opened 2025-04-10
(draft) feat: add ruff and pre-commit hook

rjwalters:
# Code Quality and Developer Experience Improvements

This PR introduces some improvements to our development workflow and code quality tools:

## Key Changes
- **Added Ruff for code formatting and linting**: Replaces pylint with the faster and more modern Ruff tool
- **Added Git hooks**: Pre-commit hook to automatically run Ruff checks before commits
- **Improved README installation instructions**: added Conda and UV setup options
- **Fixed omegaconf warnings**: Added `_self_` as default
- **Enhanced type safety**: 
  - Added mypy for better type checking
  - Added type annotations in two utilities (util/config.py and util/datastruct.py)
- **VSCode setup improvements**: 
  - Added recommended extensions
  - Updated settings for better editor integration

## Developer Setup
After pulling this PR, run the following to set up git hooks:
```bash
chmod +x devops/setup_git_hooks.sh
./devops/setup_git_hooks.sh
```

These changes should make the codebase more maintainable and improve the developer experience for both new and existing contributors.


.vscode/settings.json
```diff
@@ -1,37 +1,42 @@
...
+  "python.testing.unittestArgs": [
+    "-v",
+    "-s",
+    "./tests",
+    "-p",
+    "test_*.py"
+  ],
+  "python.testing.pytestEnabled": false,
+  "python.testing.unittestEnabled": true,
+  // "python.analysis.extraPaths": ["../sample-factory", "../Griddly/python"],
```
daveey, line 32: this can go

README.md
```diff
@@ -100,24 +110,55 @@ conda activate metta
...
+
+This project uses Git hooks to enforce code quality standards.
+
+### Setup
+
+To set up the Git hooks, run:
+
+```bash
+# Make the setup script executable
+chmod +x devops/setup-git-hooks.sh
```
daveey, line 132: lets put this in ./devops/sandbox

configs/sweep.yaml
```diff
@@ -1,4 +1,5 @@
 defaults:
+  - _self_
```
daveey, line 2: this is so dumb, is there a way to just tell hydra to not require it?

tools/eval.py
```diff
@@ -2,15 +2,32 @@
...
-from util.runtime_configuration import setup_metta_environment
-from rl.wandb.wandb_context import WandbContext
+from rich.logging import RichHandler
+
 from eval import simulate_policy
+from rl.wandb.wandb_context import WandbContext
+from util.runtime_configuration import setup_metta_environment
+
+# Configure rich colored logging
+logging.basicConfig(
```
daveey, line 12: we do this a lot. can we replace with logger = get_metta_logger("name"). also look at train.py / trainer.py where the logger includes a local-rank. this is helpful for when we train on multiple gpus and each process has its own rank. would be good to just auto-handle all that

configs/sweep.yaml
sasmith: Also, I think this is the opposite of where we want it. I think we want to apply all of the other defaults, and then _self_ at the end. I.e., so the values defined here can override the defaults from common, and not the other way around.

devops/git-hooks/pre-commit
```diff
@@ -0,0 +1,25 @@
...
+
+# Check if ruff is installed
+if ! command -v ruff &> /dev/null; then
+    echo "Error: ruff is not installed. Please install it with 'pip install ruff'."
+    exit 1
+fi
+
+# Run ruff check on staged files - output will only show if there are errors
+if ! echo "$STAGED_FILES" | xargs ruff check; then
+    echo "❌ Ruff checks failed. Please fix the issues before committing."
```
sasmith, line 20: I'm happy to experiment with this, but I think I want people (me) to be able to commit to staging branches without running lint checks like this. E.g., if I'm switching branches and want to commit a "work in progress".

Is there a reasonable way to distinguish between commits like this, and commits-for-outside-consumption?

sasmith (commented):
Would you mind pulling this into a couple-or-so separate PRs? E.g., I'd like to be able to review the ruff changes independently from other parts.

devops/git-hooks/pre-commit
rjwalters: > When _self_ appears early in the defaults list (like in the PR):
>  - The config's own values are loaded first
>  - Then other defaults are merged on top, potentially overriding the config's values
> 
> 
> When _self_ appears at the end of the defaults list:
>  - Other defaults are loaded first
>  - Then the config's own values are merged on top, overriding any defaults

I think you are correct that we want it at the end

rjwalters: on the precommit hook... you always have the option of bypassing tests:
```
git commit --no-verify -m "Your commit message"
```

our PRs will be smaller if we avoid whitespace changes. Most other projects I have worked on require regular linting -- and you can set up vscode to automatically fix problems when you save a file using the ruff extension

tools/eval.py
rjwalters: see #125

devops/git-hooks/pre-commit
berekuk: Maybe pre-push hook instead of pre-commit?

sasmith:
I think the contents of this moved elsewhere. Want to close this out?

rjwalters:
seems like we don't need the precommit hook and the rest of the changes here have found their way into other PRs

PR #124, opened 2025-04-10
Mettagrid symlinks refactor

berekuk:
My [mapgen PR in mettagrid repo](https://github.com/Metta-AI/mettagrid/pull/73) is currently blocked because I split the configuration into multiple files, but then realized that metta repo symlinks only some specific files.

This PR uses the new more robust approach:
- `./configs/env/mettagrid-base` is a symlink
- `./configs/env/mettagrid` is for this repo exensions
- mettagrid defaults are loaded with `  - /env/mettagrid-base/mettagrid@`, this seems to work

I've updated all configs that used the old paths, hope I didn't miss anything.

There are a couple of user configs that use `  - override /env/mettagrid@env: simple`, my belief is that this line was broken before this PR, so I didn't touch it.


sasmith (approved):
I'm not sure how much we'll want to keep the symlink as it's set up here, but the single symlink seems much cleaner and more explicit about what's happening that the four. Thanks for cleaning this up.

> override /env/mettagrid@env: simple

Yeah, this was broken by a change some ~2 weeks ago.

PR #127, opened 2025-04-11
Check and help people log into AWS or W&B.

treeform:


daveey (approved):

util/config.py
```diff
@@ -11,3 +13,36 @@ def config_from_path(config_path: str, overrides: DictConfig = None) -> DictConf
...
+def read_file(path: str) -> str:
+    with open(path, "r") as f:
+        return f.read()
+
+def setup_metta_environment(
+    cfg: DictConfig,
+    require_aws: bool = True,
+    require_wandb: bool = True
+):
+    if require_aws:
```
sasmith, line 26: Can we be more precise about when this is actually required, to make it match when we're going to use AWS? Right now I think this happens all the time.

PR #129, opened 2025-04-11
add job types to wandb runs so we can filter to job_type=train

jacklionheart:
I would like to eliminate most/all runs that aren't training, but this seems like a useful tag regardless. I tested it with https://wandb.ai/metta-research/metta/runs/jack.wandb-runtype

### DIFF
--- a/rl/wandb/wandb_context.py
+++ b/rl/wandb/wandb_context.py
@@ -5,14 +5,14 @@ import sys
 from omegaconf import OmegaConf
 
 class WandbContext:
-    def __init__(self, cfg, resume=True, name=None, run_id=None, data_dir=None):
+    def __init__(self, cfg, job_type=None, resume=True, name=None, run_id=None, data_dir=None):
         self.cfg = cfg
         self.resume = resume
         self.name = name or cfg.wandb.name
         self.run_id = cfg.wandb.run_id or self.cfg.run or wandb.util.generate_id()
         self.run = None
         self.data_dir = data_dir or self.cfg.run_dir
-        self.job_type = self._job_type()
+        self.job_type = job_type
     def __enter__(self):
         if not self.cfg.wandb.enabled:
             assert not self.cfg.wandb.track, "wandb.track won't work if wandb.enabled is False"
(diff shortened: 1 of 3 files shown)


rl/wandb/wandb_context.py
```diff
@@ -41,6 +43,27 @@ def __enter__(self):
...
         return self.run
 
+
+    def _job_type(self) -> str:
+        """Detect job type based on which script is being run."""
+        # Get the entry point script
+        entry_point = sys.argv[0]
+        
+        # Check for module execution style (python -m tools.X)
+        if entry_point.endswith('__main__.py') and len(sys.argv) > 1:
```
sasmith, line 53: I'm impressed by the magic, but I think I'd rather us just put the job type in the config, like `wandb.run_type`. This feels like it's setting up dependencies and surprises that we don't need to have.

sasmith (changes requested):

rl/wandb/wandb_context.py
```diff
@@ -11,7 +12,7 @@ def __init__(self, cfg, resume=True, name=None, run_id=None, data_dir=None):
         self.run_id = cfg.wandb.run_id or self.cfg.run or wandb.util.generate_id()
         self.run = None
         self.data_dir = data_dir or self.cfg.run_dir
-
+        self.job_type = self._job_type()
```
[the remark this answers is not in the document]
jacklionheart: Just seems nice to be able to inspect if you care so recomputing

rl/wandb/wandb_context.py
jacklionheart: I don't think setting it in config maps well to the world where we currently have one master config that we reuse across all job types -- having to set your cmd in addition to invoking your cmd just seems error prone to me. i want us to require less careful management of configs, and make it less possible to accidentally register a non training run as training.

i'd be more open to just setting it directly in your tool/<cmd>.py file, but i'm not sure of a good place to put that other than in the config without a lot of additional plumbing, which, again, feels like opening the door for mistakes or confusion.

sasmith: Maybe we just pass the job_type to `WandbContext` when we instantiate it? That seems pretty clean in `train.py`.

sasmith (approved):
Happy to discuss more on how the job_type is determined if you want, but I don't want to bike shed on it. You've heard my concerns, and if you think your current method is the best option, ship it. In either case, I agree that having the job_type information is something we want.

jacklionheart:
i agree that does work well for train. it doesnt work as cleanly for all other contexts but since the goal is to mostly eliminate them, its probably fine to leave them as null for now.

rl/wandb/wandb_context.py
daveey: fine to leave as is, but i would:

add job_type to the wandb config. set it to "unset" by default. then override it in the train.yaml/eval.yaml/.... or in the code as needed

PR #131, opened 2025-04-11
Basic dev machine setup script

sasmith:
This add a basic machine setup script that installs homebrew, and then installs a number of things from homebrew that we expect developers will want.

Testing: ran it locally and confirmed it was a noop. Ran the brew command earlier and confirmed it installed things as needed. Hard to test on a tabula rasa machine.


daveey (approved):
lets put in devops/sandbox/

PR #134, opened 2025-04-12
Add muon optimizer

daveey:
1. Add support for the Muon optimizer
2. Create a trainer.optimizer config, and set to Adam by default

Tested on mettabox

### DIFF
--- a/rl/pufferlib/trainer.py
+++ b/rl/pufferlib/trainer.py
@@ -11,7 +11,7 @@ import torch
 import torch.distributed as dist
 from fast_gae import fast_gae
 from omegaconf import OmegaConf
-import heavyball
+from heavyball import ForeachMuon
 
 from util.config import config_from_path
 import wandb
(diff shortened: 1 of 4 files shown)


rl/pufferlib/trainer.py
Al-does, line 1: I think Muon only works for 2D parameters and other layers have to be registered with AdamW. As it's written, the optimizer would leave out things like the last layer of the critic MLP, bias vectors, and, soon, the one of the layers of the actor net since it'll be 1D too. Depending on how it tests, the embedding layer would also be not be optimized. Not sure if the best approach is to have MettaAgent register its parameters into two groups or do a something like for param in self.policy.parameters and send the 2d ones to Muon and the rest to AdamW.

daveey: Great point. I didn't know that.

It seems to learn faster initially, but doesn't do as well as Atom. Maybe using 2 optimizers would make it even better.

Al-does: I'm definitely excited about it.

PR #135, opened 2025-04-12
Deal with missing keys better in policy_fitness

sasmith:
We were previously branching based on whether we had one or multiple evals. We presumably did this because `dataframe.loc[x]` returns a different type depending on whether `x` is unique in the dataframe or not. This protection wasn't quite perfect, since the size of `evals` is based on the candidate policy and the baseline policies, so it's possible for `evals` to have size greater than 1, but for `candidate_uri` to still only have one eval, and thus have `metric_data.loc[candidate_uri]` be "the wrong type".

I haven't fully figured out if we should be trying to process evals that we didn't even perform for our candidate. It seems a little silly. So it's possible that part of "the right solution" is to filter down `evals` more. However, in this diff I'm side stepping that question. Instead, I'm changing `dataframe.loc[x]` to `dataframe.loc[[x]]`. The latter has a consistent return type / dimensions regardless of whether `x` is a unique key in `dataframe`. This avoids us needing to differentiate between situations where we have unique keys and not, and fixes a crash I was hitting.

This also updates how we were providing some default values. Previously we appeared to crash when keys were missing, and now we should return 0 when keys are missing. It's possible that the default values were covering a different case than I was hitting here, so there could be a regression if our dataframe is sometimes missing `metric_mean`, but this would be surprising to me.

I've added https://app.asana.com/1/1209016784099267/project/1209185339447618/task/1209967199782635 to track adding unit tests here.

### DIFF
--- a/rl/eval/eval_stats_analyzer.py
+++ b/rl/eval/eval_stats_analyzer.py
@@ -124,17 +124,14 @@ class EvalStatsAnalyzer:
 
         evals = metric_data[eval].unique()
 
-        if len(evals) > 1:
-            candidate_data = pd.DataFrame(metric_data.loc[candidate_uri]).set_index(eval)
-            baseline_data = metric_data.loc[baseline_policies].set_index(eval)
+        candidate_data = pd.DataFrame(metric_data.loc[[candidate_uri]]).set_index(eval)
+        baseline_data = metric_data.loc[baseline_policies].set_index(eval)
 
         for eval in evals:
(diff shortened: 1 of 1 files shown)


daveey (approved):

sasmith:
@daphnedemekas you merged a revert (d802760660b575cd11c3c977b2265e2ba2c0d677), which conflicts with this to some degree. Did you determine the revert was needed on top of this PR, or was the revert based on testing before this PR?

PR #140, opened 2025-04-15
Add a metta package

jacklionheart:
This puts the following directories inside "metta": agent, eval, rl, util
tests/ is intentionally left separate. It's possible tests/eval should become tests/metta/eval.

Similarly, tools is left outside. I could see putting this inside but i mostly think of it as scripts *on top* of meta.
By the way, have I mentioned that _target_ is not robust to refactoring?

See: https://app.asana.com/1/1209016784099267/project/1209041170403492/task/1209981926746401?focus=true for motivation.

Testing:
* Trained (briefly) locally
* Trained on a pufferbox: https://wandb.ai/metta-research/metta/runs/jack.test.imports
* Ran `pytest tests/eval`


devops/aws/batch/job_queue.py
```diff
@@ -13,54 +13,57 @@
 
 from .job import format_time_difference
 
-def get_boto3_client(service_name='batch'):
+
+def get_boto3_client(service_name="batch"):
     """Get a boto3 client with standard configuration."""
-    config = Config(retries={'max_attempts': 10, 'mode': 'standard'}, max_pool_connections=50)
+    config = Config(retries={"max_attempts": 10, "mode": "standard"}, max_pool_connections=50)
```
sasmith, line 19: Sure -- seems like a simple enough diff to pull in whitespace changes as well.

env/wrapper/feature_masker.py
sasmith, line 1: A number of files (all of `env/wrapper` and `wrapper`?) were deleted, versus appearing to be moved. Was this intentional?

sasmith (changes requested):
1) Approve in general -- just sending it back because of the deleted files. These are cases where it seems like we have two copies of the files and you're deleting both.
2) In general this would have been cleaner if the lint changes had been done first in a separate diff. That would also have allowed git to more accurately understand that these were all moves (it seems to have missed in observation_normalizer?), which would make git-related tools slightly easier to use across this boundary in the future.

sasmith:
Also, tests should mirror our directory structure. So either we should put them alongside the original files, or the tests directory should be changed like this is (but in a different PR!).

env/wrapper/feature_masker.py
jacklionheart: yes, they appeared to be unused

sasmith:
> > Also, tests should mirror our directory structure. So either we should put them alongside the original files, or the tests directory should be changed like this is (but in a different PR!).
> 
> Yes. The one existing test file besides tests/eval should be organized
> 
> It is I believe standard procedure to have pkg/pkg/foo and then pkg/tests/foo. That is what I have here, but i have also seen pkg/tests/pkg/foo

@daphnedemekas is adding a second test file!

sasmith (approved):

PR #143, opened 2025-04-15
Pull out simulation into its own module

jacklionheart:
This pulls out all of the code that is used to do inference (i.e. simulation) into its own library.  This makes it easier to avoid cyclic imports, because rl should depend on simulation, but simulation doesnt need RL. We also pulled out wandb into util because while mostly we should only use wandb during RL, that's not the case right now, so this was necessary to preserve the fact that sim doesnt depend on RL.

This is all a prelude to introducing a typed SimConfig.

Also, add some backwards-compatability to allow us to load old policies with the new refactor into a metta-scoped agent module.


daphnedemekas (approved):
Hey, this looks good. The only thing that is potentially confusing is that now we have a sim/simulator.py and a sim/simulation.py and they are quite different. Its OK with me, but maybe want to check with Andre / David to see if its worth renaming the other one or adding little docstrings to the top of the files to make the difference clear.

jacklionheart:
They aren't really all that different, its just that they both only do the work they personally care about, but it should be one class with the combined competency. I did add a comment in simulator.py.

PR #144, opened 2025-04-15
fix: remove custom ruff setting for "lines-after-import"

rjwalters:
There seems to be some problem with the vscode Ruff extension and this setting:

https://github.com/astral-sh/ruff-vscode/issues/738


jacklionheart (approved):
lg2m

PR #146, opened 2025-04-16
feat: improve github smoke tests

sasmith:
This updates our github smoke test to run evaluation and replay code as well.

Because we're now smoke testing eval code, this hit a couple of issues that I addressed:
* (pretty trivial) we were silently overwriting `self.trainer_cfg.evaluate_interval` when is was smaller than `self.trainer_cfg.checkpoint_interval`
* (is the cause for me doing this, since I was hitting this error) our evaluation was always failing if wandb wasn't used. I'm not confident that it fully works without wandb now, but I've address the crash I was hitting

Testing:
Ran the github training locally, and saw that the smoke test ran training / evaluation, and passed after the above fixes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,15 +16,15 @@ jobs:
     env:
       CHECKPOINT_PATH: ./train_dir/my_experiment/checkpoints/
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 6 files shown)


sasmith:
Well, I see I'm failing a smoke test. Not clear why this is working for me locally but failing on github, but I guess it might be doing it's job?

sasmith:
Thanks for the fixes, @rjwalters -- we're passing now. I still think there's something funny about the environment between laptops and github being different in some way I don't understand, but I'm okay addressing this when it pops up as a practical issue.

rjwalters (approved):

PR #147, opened 2025-04-16
Av-ls-ag-eigen

lars-sandved:
This branch contains the code for plotting various metrics during training that can be used to infer the degree of criticality for each layer. 

The metrics are calculated using the analyze_sv method in dynamics.py and include:

- mean singular value  -- This is related to the branching coefficient and is near 1 at criticality
- largest singular value
- ratio of max_sv / min_sv ("the condition number") 
- the R^2 of a linear fit in loglog space of the singular value distribution -- good fit is >0.95
- the slope of that fit -- in range [-1.5,-.5] at criticality
- the % of singular values that have collapsed to near zero -- gives information about the appropriateness of the layer size

These are calculated for each layer individually, currently plotting for actor_1, critic_1, fc1, and encoded_obs.

Agent configs should have the argument "effective_rank: true" for every layer of interest. Metrics are caluclated every effective_rank_interval timesteps.


lars-sandved (commented):
I think this is ready. Successfully merged main into this branch, and then ran a training run -- all worked and metrics showing on wandb.

metta/agent/lib/metta_layer.py
Al-does, line 1: Let's change anything "eigen" to "SV" or "singular value"

configs/hardware/macbook.yaml
Al-does, line 1: Leave this as cpu for now, when you check it in.

configs/env/mettagrid/prog_1b.yaml
Al-does, line 1: Delete the prog maps.

configs/agent/simple.yaml
Al-does, line 1: set all effective rank intervals to 300

metta/agent/lib/metta_layer.py
lars-sandved: Thanks good idea.

metta/agent/lib/metta_layer.py
```diff
@@ -220,7 +222,19 @@ def compute_effective_rank(self, delta: float = 0.01) -> dict:
...
+        except Exception as e:
+            avg_abs_weight = torch.abs(weights).mean()
+            cond_num = torch.linalg.cond(weights)
+
+            print(f"Error computing SVD or weight norm for layer {self._name}: {e}")
+            print(f"Average absolute weight: {avg_abs_weight}")
+            print(f"Condition number: {cond_num}")
+            return None
 
         # Calculate the cumulative sum of singular values
```
Al-does, line 239: @lars-sandved I'm so sorry but I should have mentioned a bigger change that I really think we should make: realizing that this project is getting more serious, we need to move away from the "insert hooks to play around with metrics" approach and more into "organizing things into where they should be." This will pay off in the future as we'll have greater flexibility to interact with weights, not just singular values. 

To me, I think that means moving all calculations _on_ weights into dynamics.py. So that's to say that weights should get passed to analyze_sv() and all calculations on weights including the SVD, calculating effective rank, and weight_norm should be moved there too. This also has the small benefit of returning a single metrics dictionary.

Again, sorry I didn't flag this earlier today.

lars-sandved: Hey @Al-does, that's a great idea, and I agree that will make things more scalable for the future. No problem at all, I'm happy you pointed this out. I'll get to updating that :)

lars-sandved: Ok done! Please have a look

metta/agent/util/weights_analysis.py
```diff
@@ -0,0 +1,74 @@
...
+        - Effective rank
+        - Basic statistics if SVD fails
+    """
+    metrics = {}
+
+    # Basic weight statistics
+    weight_norm = torch.linalg.norm(weights).item()
+    metrics["weight_norm"] = weight_norm
+
+    try:
```
daveey, line 27: i don't love the try-catch here. when do we expect this to fail? can we just check for that before trying this code path?

metta/rl/pufferlib/trainer.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, cfg: OmegaConf, wandb_run, policy_store: PolicyStore, **kwarg
         self.average_reward = 0.0  # Initialize average reward estimate
         self._current_eval_score = None
         self._eval_results = []
-        self._effective_rank = []
+        self._weight_metrics = []
```
daveey, line 65: how about we do:

self._weights_helper = WeightsMetricHelper(cfg)

metta/rl/pufferlib/trainer.py
```diff
@@ -184,8 +184,11 @@ def train(self):
                 self._checkpoint_trainer()
             if self.trainer_cfg.evaluate_interval != 0 and self.epoch % self.trainer_cfg.evaluate_interval == 0:
                 self._evaluate_policy()
-            if self.cfg.agent.effective_rank_interval != 0 and self.epoch % self.cfg.agent.effective_rank_interval == 0:
-                self._effective_rank = self.policy.compute_effective_rank()
+            if (
```
daveey, line 187: self._weights_helper.on_epoch_end(self.epoch)

metta/rl/pufferlib/trainer.py
```diff
@@ -603,7 +610,7 @@ def _process_stats(self):
                     **{f"performance/{k}": v for k, v in performance.items()},
                     **environment,
                     **policy_fitness_metrics,
-                    **effective_rank_metrics,
+                    **weight_metrics,
```
daveey, line 613: self._weights_helper.stats()

metta/rl/pufferlib/trainer.py
```diff
@@ -614,7 +621,7 @@ def _process_stats(self):
             )
 
         self._eval_results = []
-        self._effective_rank = []
+        self._weight_metrics = []
```
daveey, line 624: self._weights_helper.reset()

daveey (changes requested):

metta/agent/util/weights_analysis.py
lars-sandved: I made the possible svd errors explicity and removed the try.

lars-sandved:
Thanks for the suggestions David. I've updated the error handling and added a helper class -- see what you think :)

configs/agent/reference_design.yaml
```diff
@@ -159,7 +159,7 @@ components:
     _target_: metta.agent.lib.nn_layer_library.Linear
     input_source: critic_2
     output_size: 128
-    effective_rank: true # need to turn effective rank on if you want it
+    analyze_weights: true # need to turn analyze weights on if you want it
```
daveey, line 162: comment not helpful

lars-sandved: moved the comment up and edited

PR #148, opened 2025-04-16
fix: add explicit `_self_` to OmegConf defaults lists

rjwalters:
This PR implements two changes to improve our configuration management:

1. **Adding `_self_` to defaults lists**: Many of our YAML config files were missing the `_self_` directive at the end of their defaults lists, which is considered best practice in OmegaConf/Hydra. This ensures that local configuration values properly override inherited configs, maintaining the expected precedence order.

2. **Applying Prettier formatting**: Applied consistent formatting to all YAML files using Prettier, which improves readability and maintainability. This includes standardizing indentation, list formatting, and comment alignment.

## Changes
- Added `_self_` to defaults lists in configs that were missing it
- Applied Prettier formatting to all YAML configuration files
- Fixed inconsistent spacing and indentation
- Standardized list formatting for better readability

These changes are purely formatting-related and have no functional impact on the codebase.


configs/env/mettagrid/navigation/training/varied_terrain.yaml
```diff
@@ -1,7 +1,9 @@
 # Config: Dense Obstacles & Sparse Altars
 defaults:
   - /env/mettagrid-base/mettagrid@
-sampling: 20
+  - _self_
+
+sampling: 1
```
sasmith, line 6: This seems like it may be accidental? If not, can you say more about the change?

rjwalters: as per discussion on the intent behind our "sampling" distribution (@daveey) the sampling parameter must be between zero and one.

See https://github.com/Metta-AI/mettagrid/blob/13e0085b0b0964946f1a84f815e0db4104497b77/mettagrid/resolvers.py#L66 for details

sasmith: Got it. Then *approved* including this change.

But my usual request for splitting things like this into separate PRs, since it makes reviewing easier. I was going to say "they don't charge per PR", but we're using Unito to sync github and Asana, so that's not actually true. But it remains that PRs are cheap.

PR #152, opened 2025-04-16
improve (reduce) wandb dependence

sasmith:
This updates our readme commands to not depend on wandb in an explicit way, so the ideal of "commands in the readme should always work" is more attainable. This updates out github workflow tests to match the explicit turning off of wandb.

This also addresses two issues in code related to wandb usage. One where failure is reasonable, and the error is now a smidge better, and one where we were failing when wandb was none, but don't need to.

Testing:
ran
`AWS_ACCESS_KEY_ID=none AWS_SECRET_ACCESS_KEY=none python -m tools.train run=my_experiment +hardware=macbook wandb=off` locally. Saw no crashing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,7 +93,7 @@ jobs:
           AWS_SECRET_ACCESS_KEY: set_but_not_used
         run: |
           source venv/bin/activate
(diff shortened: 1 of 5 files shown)


rjwalters (approved):
lgtm

PR #162, opened 2025-04-17
Npc evals

daphnedemekas:
Add a new section on wandb that includes evals that were played against npcs within the environment

See what the wandb run looks like here: https://wandb.ai/metta-research/metta/runs/evals_with_npc_test?nw=nwuserdaphned

Note that the eval with the npc is playing on simple, the episode_reward is not between 0 and 1.


sasmith:
Similar comment as in https://github.com/Metta-AI/metta/pull/164

The main reason I'm showing up is I see that people are starting to prefix their PR titles with their name, and I wasn't sure what the history there was.

configs/eval/all.yaml
```diff
@@ -91,5 +91,9 @@ evals:
   temple_use_free:
     env: env/mettagrid/object_use/evals/temple_use_free
     policy_agents_pct: 1.0
+  simple_npc:
+    env: env/mettagrid-base/simple
+    policy_agents_pct: 0.5
+    npc_policy_uri: wandb://run/b.daveey.t.8.rdr9.3
```
daveey, line 97: can we make the eval name as the graph label? so it would be npc_evals/simple_npc:episode_reward

daveey: and/or a "stats_label" field?

metta/sim/eval_stats_analyzer.py
```diff
@@ -84,7 +83,7 @@ def analyze(self, include_policy_fitness=True):
                     policy_fitness_df,
                     headers=[self.candidate_policy_uri] + list(policy_fitness_df.keys()),
                     tablefmt="grid",
-                    maxcolwidths=25,
+                    # maxcolwidths=25,
```
daveey, line 86: ?

daveey (approved):
looks great! feel free to merge after addressing comments

metta/sim/eval_stats_analyzer.py
daphnedemekas: This raised an exception now, after adding npc fields to the policy_fitness_df (because a lot of the values for this fields i None).  I just removed it because i don't think it matters that much

PR #164, opened 2025-04-17
More navigation maps and configs 

georgejwdeane:
This adds a number of new maps and configurations being used to try and improve agent navigation. The main thrust of this PR is to add harder evals and make sure all of the eval scores are summing to 1.

Testing:
Ran training with these maps on a Mettabox and AWS


sasmith:
Hello from the peanut gallery.

My preference would be for PRs to have more descriptive titles and descriptions. Like, I'd love to know more about _why_ various decisions were made here. So something like

Title: `More navigation maps and configs [don't need to prefix with "George" -- git history will get that]`

Body:
```
This adds a number of new maps and configurations being used to try and improve agent navigation. The main thrust is to add maps to help with X, Y, and Z, to try and address problem ABC. See [some asana link] for details.

Testing:
Ran training with these maps on a Mettabox, primarily as a smoke test to see that the configurations were valid
```


(I don't want to be overly prescriptive here -- the main point is that I'd love a few sentences or a small paragraph around why you wrote this. Hopefully something we could look at in a year and understand the history.)

daphnedemekas (approved):

PR #165, opened 2025-04-17
Torch Profiler

Al-does:
Added a torch profile class that can be used as context in trainer. Set profiler_interval_epochs to a nonzero integer to arm profiling.

- Profiles are saved as json.gz files locally in train_dir/<your_run>/torch_traces/ and in S3 at torch_traces/<your_run>/
- A link to download the S3 file is dropped into wandb under "torch traces"

To view traces, go to chrome://tracing (or arc://tracing if you happen to use that browser which is fine) and select load. Navigate traces using WASD on your keyboard.

Future work could include support for TensorBoard.


daveey (changes requested):

daveey:
@Al-does I think this is not blocked on me. Let's get this merged?

Al-does:
Merging these features via another branch, `alex-trace-3`

PR #176, opened 2025-04-22
Add a lattice world

jacklionheart:


daphnedemekas (commented):
This diff has a lot of changes on it, I think you may need to request to merge it into george-navigation-training? 
Or we can get those changes merged into main first

PR #181, opened 2025-04-23
fix: handle single agent env in mettagrid/renderer/raylib

rjwalters:
This PR repairs `raylib_renderer.py` in `mettagrid` to correctly handle environments with a single agent.

Previously, the renderer assumed all action and observation tensors had shapes consistent with multi-agent environments. These assumptions caused runtime errors when only one agent was present. This update introduces safeguards and conversion logic to ensure compatibility across both single-agent and multi-agent cases.

### Why It Matters

- Prevents shape mismatches from crashing the renderer in single-agent mode  
- Improves clarity and type safety by making tensor shape expectations explicit  
- Enhances debuggability by raising informative errors for malformed input

---

### Summary of Changes

**Action tensor handling**  
- Added support for 1D `actions` tensors or arrays in single-agent cases  
- Ensured `actions` is always shaped `(max(1, num_agents), 2)`  
- Added fallback conversion from NumPy to Torch tensors where needed  

**Observation validation**  
- Converted NumPy arrays to Torch before applying `permute`  
- Added validation of permuted observation shapes  

**Reward shape checks**  
- Verified the shapes of `rewards` and `total_rewards`  

**Rendering safety checks**  
- Validated `actions` shape before processing in `draw_attacks`  
- Ensured action indices and values are within expected bounds


rjwalters:
I can't reproduce the issue this code was supposed to address so we'll close this for now

PR #182, opened 2025-04-23
dump stats when analyzing

daveey:
Adds a simple_solo eval, and dumps stats during analysis.

This should be handled better as described in: 

https://app.asana.com/1/1209016784099267/project/1209041170403478/task/1210061644002694?focus=true

Also adds a missing requirement and fixes some lint.


configs/eval/simple.yaml
```diff
@@ -7,6 +7,9 @@ defaults:
 env: null #the env is set in evals:
 
 evals:
+  simple_solo: #this will be in the title of the wandb eval plot
+    env: env/mettagrid-base/simple
```
jacklionheart, line 11: This is going to need some merge changes.

The way things are set up, we assume configs/sim/simple.yaml is going a single simulation. We can add configs/sim/simple_npc or configs/sim/simple_solo but i wouldnt add the one that has both for now unless you actively want it.

jacklionheart (approved):

PR #187, opened 2025-04-23
Add style-based training plots in wandb

daphnedemekas:
Adds plots to wandb for any rooms that have a .label environment variable

Labels can be either map labels (such as small medium big, or anything else we define in a Room class), or env variables, which can be defined in a cfg, or overwritten in MettagridEnv.

Map labels "small" "medium" and "big" are always generated based on area

See example here: 
https://wandb.ai/metta-research/metta/runs/daphne.local.3?nw=nwuserdaphned

under env_rewards/

This PR includes some changes like safe loading multienv and updated configs for single agent maps.


daveey (approved):
Looks good. Looks like you included the FileLock change in here, in the future, could you have this be 2 PRs?

PR #192, opened 2025-04-24
Alex Robust Action Agents

Al-does:
Usage: 
After loading a policy but before using it (i.e., running a forward pass or compiling), the policy should be given the action names and max parameters for the environment it is operating in. Call .activate_actions(actions_names, actions_max_params, self.device).

This code is not backwards compatible with non-robust agents.

---
Background:
Agents are now able to run on multi-discrete action spaces ranging from 1 to 100 actions. Increasing beyond 100 actions is as simple as provisioning a bigger embedding layer or even adding a second one. The agent's actor architecture has moved from a two-layer MLP to a variant of an attention mechanism. The agent generates an embedding, e, (a vector representation of some dimensionality) for each action. It then combines these representations with the hidden, h, representation in a bilinear interaction hWe. All three terms are learnable.

The actor components have been stuffed into a custom class, breaking away from the decomposed and reconfigurable setup of the rest of the agent. Future work can break this back out to make NAS easier.

nn.Bilinear() was found to run 10x slower than a hand-written bilinear function.

Testing:
- Agents have successfully incorporated new actions well into training without instability or degraded learning.
- Agent config simple learns at least as well as the previous simple agent. simple_multi_attn learns much better than the old simple_big. Learnable embeddings performed better than fixed hashes in representing actions. A bilinear interaction learned better than concatenating hashes with the hidden vector. Performance has not been optimized yet (a sweep over hyperparameters is needed).
- Wall clock time for the new simple versus the old is almost equivalent. simple_multi_attn is roughly 40% slower than simple.

Remaining work:
- The new agent architecture needs a hyperparameter sweep and to be trained with Muon.
- Kickstarter code has been updated but not tested.
- I believe 3D tensors (related to the bilinear layer) are compatible with Muon (their GitHub says $\ge 2$D) but have not confirmed this in our code.
- NAS experiments with simple_multi_attn.
- Break apart the actor classes into more configurable parts.
- General research using the affordances of embeddings (to be discussed).


metta/agent/lib/nn_layer_library.py
```diff
@@ -1,144 +1,229 @@
...
+        net = nn.Embedding(
+            **self._nn_params
+        )
+        if self.initialization.lower() == 'orthogonal':
+            weight_limit = 0.1
+            nn.init.orthogonal_(net.weight)
+            with torch.no_grad():
+                max_abs_value = torch.max(torch.abs(net.weight))
+                net.weight.mul_(weight_limit / max_abs_value)
+        elif self.initialization.lower() == 'max_1':
```
daveey, line 54: max_1 and max_0_1 are the same, is that intentional?

daveey: would be better if these were enums not inlined strings. als max_0_1 is confusing since a=-1, not sure if there's a better name

metta/agent/lib/action.py
```diff
@@ -1,13 +1,202 @@
...
+    def __init__(self, initialization='max_0_01', **cfg):
+        super().__init__(**cfg)
+        self._reserved_action_embeds = {} 
+        self.num_actions = 0 
+        # num_actions to be updated at runtime by the size of the active indices
+        self._out_tensor_shape = [self.num_actions, self._nn_params['embedding_dim']]
+        self.initialization = initialization
+        self.register_buffer('active_indices', torch.tensor([], dtype=torch.long))
+        
+    def activate_actions(self, strings, device):
```
daveey, line 24: self, action_names, device

metta/agent/lib/action.py
```diff
@@ -1,13 +1,202 @@
...
+        ''' each time we run this, we update the metta_agent object's (the policy's) known action strings and associated indices'''
+        for string in strings:
```
daveey, line 26: for action_name in action_names

metta/agent/lib/action.py
```diff
@@ -1,13 +1,202 @@
...
+    def _forward(self, td: TensorDict):
+        B_TT = td['_BxTT_']
+        td['_num_actions_'] = self.num_actions
+
+        # get embeddings, unsqueeze the 0'th dimension, then expand to match the batch size
+        td[self._name] = self._net(self.active_indices).unsqueeze(0).expand(B_TT, -1, -1).contiguous()
+        
+        return td
+    
+class ActionHash(metta_layer.LayerBase):
```
daveey, line 46: why do we need action_hash? I thought we were now just using embeddings.

daveey: > looks like it's not used. if you want to save this code, put it in lib/example_code.py or something

metta/agent/lib/action.py
```diff
@@ -1,13 +1,202 @@
...
+
+        return embedding
+
+    def _forward(self, td: TensorDict):
+        B_TT = td['_BxTT_']
+        td['_num_actions_'] = self.num_actions
+        td[self._name] = self.action_embeddings.unsqueeze(0).expand(B_TT, -1, -1)
+        return td
+    
+class MettaActorBig(LayerBase):
```
daveey, line 88: can we move this to actor.py

metta/agent/lib/action.py
```diff
@@ -1,13 +1,202 @@
...
+
+        # Reshape MLP output back to sequence and action dimensions
+        action_logits = mlp_output.view(B_TT, num_actions) # Shape: [B*TT, num_actions]
+
+        td[self._name] = action_logits
+        return td
+    
+class MettaActorSingleHead(LayerBase):
+    """
+    Implements a bilinear interaction layer followed by an MLP with a lot of reshaping.
```
daveey, line 156: same comment as above. seems wrong?

daveey (changes requested):
For future PRs can you put the notes in the description rather than discord message. I updated this one.

Overall looking great, left a bunch of small suggestions.

configs/agent/simple.yaml
```diff
@@ -50,46 +53,52 @@ components:
...
-  _action_param_:
-    _target_: metta.agent.lib.action.ActionParam
-    input_source: actor_1
-    nonlinearity: null
+  _action_embeds_:
+    _target_: metta.agent.lib.action.ActionEmbedding
+    input_source: null
+    nn_params:
+      num_embeddings: 100
+      embedding_dim: 16
```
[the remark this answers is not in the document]
Al-does: I can't give a definitive answer but I tried 8, 16, and 32 and, squinting at test results, it seems like 16 was best but this needs to be something we sweep over. FWIW, you can think of the embedding dim as a hidden size of a third mlp layer which is to say that my hunch is that you wouldn't want to make it too small. Also, this difference hardly costs us any perf (the difference of 8-24 params per action so, at worst, 600 params on a 500k param model). Of course, too big a model can struggle to get off the ground so, yes, 16 might be more likely to have that issue than 8. Let's make sure to sweep over this!

metta/agent/lib/lstm.py
```diff
@@ -1,34 +1,34 @@
...
-        self.obs_shape = obs_shape
+        self._obs_shape = list(obs_shape) # make sure no Omegaconf types are used in forward passes
         self.hidden_size = hidden_size
-        self.num_layers = self._nn_params["num_layers"]
+        self._out_tensor_shape = [hidden_size]
+        self.num_layers = self._nn_params['num_layers']
 
     def _make_net(self):
-        net = nn.LSTM(self._input_size, self.hidden_size, **self._nn_params)
+        net = nn.LSTM(self._in_tensor_shape[0], self.hidden_size, **self._nn_params)
```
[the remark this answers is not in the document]
Al-does: _in_tensor_shape has replaced all instances of _input_size as it's more robust of a concept for other layers. This is the big change I've made to the flexible agent architecture. It helps with CNN flattening, merge layers, reshapes, etc.

Meanwhile, nets that take a single input feature dim should expect that the layer above them feed only a 1 dim tensor (ignoring batch dim) and hence take _in_tensor_shape[0].

metta/agent/lib/nn_layer_library.py
```diff
@@ -1,144 +1,229 @@
...
+        )
+    
+class Bilinear(LayerBase):
+    def __init__(self, **cfg):
+        super().__init__(**cfg)
+
+    def _make_net(self):
+        self._out_tensor_shape = [self._nn_params.out_features]
 
+        self._nn_params['in1_features'] = self._in_tensor_shape[0][0]
```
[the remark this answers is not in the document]
Al-does: I've chosen bad variable names (mostly because I'm expanding on concepts sometime after naming them). 

Usually, _in_tensor_shape is a list that is the shape. However, some layers have two input sources, in which case, _in_tensor_shape becomes a list of its input sources tensor shapes. I assume that classes have the logic to unpack input tensors as they need to. 

_input_source has this naming problem too: often a layer has just one input source but sometimes there are two.

Perhaps the solution is to call these 1) _in_tensor_shapes and it's always a list of list and 2) _input_sources and it's always a list of strings. I'll go forward w that unless you suggest otherwise.

metta/agent/metta_agent.py
```diff
@@ -64,69 +63,89 @@ def __init__(
...
-                    input_source_components[input_source] = self.components[input_source]
+                for name in component._input_source:
+                    input_source_components[name] = self.components[name]
                 component.setup(input_source_components)
+            else:
+                component.setup(self.components[component._input_source])
         else:
             component.setup()
 
+    def activate_actions(self, action_names, action_max_params, device):
```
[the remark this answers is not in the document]
Al-does: FWIW, this inits actions if the agent is brand new or has not seen a given action. Otherwise, it will simply tell the agent which of the actions it already knows to use for this particular training run. For clarity, this is to be called at the beginning of every training run, whether the agent is brand new or old and loaded into trainer.py. It should be called in trainer.py.

With that in mind, I can't figure out how to run this flag (was the function called?) from within the agent unless there is a metta_agent.end_training() that's called on AbortingTrainer() but even that might not get called on certain situations. Otherwise, the flag would be set to true at the beginning of a training run but we can't guarantee that we set the flag to false at the end of a training run. By "run" I mean someone calling train.py and we assume that the policy will experience more "runs" in the future. As such, I've left the requirement to call this function in trainer.py (and will do so in play.py and eval.py once I get there).

configs/agent/simple_multi_attn.yaml
```diff
@@ -1,23 +1,27 @@
 # see reference_design.yaml for explanation of components
```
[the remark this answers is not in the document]
Al-does: simple_multi_attn.yaml overrides a component in simple.yaml but it also completely deletes another component. How do you say that we use simple.yaml defaults but not only override one component but also delete another?

metta/agent/lib/metta_layer.py
```diff
@@ -54,14 +58,18 @@ def setup(self, input_source_component=None):
...
-
-        if self._output_size is None:
-            self._output_size = self._input_size
+            if isinstance(self._input_source_component, dict):
+                self._in_tensor_shape = []
+                for _, source in self._input_source_component.items():
+                    self._in_tensor_shape.append(source._out_tensor_shape.copy())
+            else:
+                self._in_tensor_shape = self._input_source_component._out_tensor_shape.copy()
+            if not hasattr(self, "_out_tensor_shape"):
```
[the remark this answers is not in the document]
Al-does: I'm changing this significantly such that out_tensor_shape is set in the _initialize() or _make_net function. I'm leaving it to be explicitly done in the super class rather than automatically performed in the base class. Because it changes from class to class, I figured being explicit about it's construction, not assuming that we can get by with None, would be safer.

metta/agent/lib/metta_layer.py
```diff
@@ -155,8 +162,8 @@ def _initialize(self):
                 raise ValueError(f"Unsupported nonlinearity: {self.nonlinearity}") from e
 
     def _initialize_weights(self):
-        fan_in = self._input_size
-        fan_out = self._output_size
+        fan_in = self._in_tensor_shape[0]
```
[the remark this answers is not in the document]
Al-does: abandoning self._input_size. Will add asserts to a bunch of the layers

metta/agent/metta_agent.py
```diff
@@ -64,69 +63,89 @@ def __init__(
...
+        '''Run this at the beginning of training.'''
+        self.device = device
+        self.actions_max_params = action_max_params
+        self.active_actions = list(zip(action_names, action_max_params))
+        
+        # Precompute cumulative sums for faster conversion
+        self.cum_action_max_params = torch.tensor([0] + [sum(action_max_params[:i+1]) for i in range(len(action_max_params))],
```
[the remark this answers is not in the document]
Al-does: Awesome. Thanks.

metta/agent/lib/nn_layer_library.py
Al-does: I ended up eliminating the option here. Orthogonalization with these magnitudes has been working well and changes outside of it haven't affected things too much anyway.

metta/agent/metta_agent.py
```diff
@@ -64,69 +63,89 @@ def __init__(
...
+        self.active_actions = list(zip(action_names, action_max_params))
+        
+        # Precompute cumulative sums for faster conversion
+        self.cum_action_max_params = torch.tensor([0] + [sum(action_max_params[:i+1]) for i in range(len(action_max_params))], 
+                                                 device=self.device)
+        string_list = []
+        for action_name, max_arg_count in self.active_actions:
+            for i in range(max_arg_count + 1):
+                string_list.append(f"{action_name}_{i}")
+        self.components['_action_embeds_'].activate_actions(string_list, self.device)
```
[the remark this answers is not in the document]
Al-does: This changes each time we run train on the agent which means the MettaAgent constructor will have already ran. 

In trainer.py, I could move the activate_actions method call to above the line where we send self.policy to device but that won't move cum_action_max_params and action_index_tensor to device (at least according to Gemini). It doesn't seem like it would make sense to set those as registered buffers either since we update them when we run training or use the model as a kickstarter.

metta/agent/metta_agent.py
```diff
@@ -149,23 +168,46 @@ def get_action_and_value(self, x, state=None, action=None, e3b=None):
...
+        action_type_numbers = action[:, 0].long()
+        action_params = action[:, 1].long()
+        
+        # Use precomputed cumulative sum with vectorized indexing
+        cumulative_sum = self.cum_action_max_params[action_type_numbers]
+        
+        # Vectorized addition
+        action_logit_index = action_type_numbers + cumulative_sum + action_params
+        
+        return action_logit_index.reshape(*orig_shape[:2], 1)
```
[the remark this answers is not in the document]
Al-does: Thanks!

metta/agent/lib/nn_layer_library.py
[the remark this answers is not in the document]
Al-does: I wend forward and made a bunch of changes to metta_agent, metta_layer, and some to each of the other layer classes that now reflect the concepts above. The code is less ambiguous now and I added some docstrings describing what each of these variables is supposed to be.

configs/agent/simple_multi_attn.yaml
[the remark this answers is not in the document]
daveey: claude suggests: 

parent_key: ~  # This will delete 'parent_key' from the inherited config

configs/sim/all.yaml
```diff
@@ -88,7 +88,8 @@ simulations:
   object_use/temple_use_free:
     env: env/mettagrid/object_use/evals/temple_use_free
     policy_agents_pct: 1.0
-  simple_npc:
-    env: env/mettagrid/simple
-    policy_agents_pct: 0.5
-    npc_policy_uri: wandb://run/b.daveey.t.8.rdr9.3
+  # b.daveey.t.8.rdr9.3 needs to be replaced with a policy robust to varied action spaces
```
daveey, line 91: can you put this back in and use one of your policies? we just want some baseline

metta/agent/lib/action.py
```diff
@@ -1,13 +1,40 @@
...
+            for action_name in action_names
+        ], device=device)
+        self.num_actions = len(self.active_indices)
+
+    def _forward(self, td: TensorDict):
+        B_TT = td['_BxTT_']
+        td['_num_actions_'] = self.num_actions
+
+        # get embeddings, unsqueeze the 0'th dimension, then expand to match the batch size
+        td[self._name] = self._net(self.active_indices).unsqueeze(0).expand(B_TT, -1, -1)
```
daveey, line 38: td[self._name] = rearrange(
    self._net(self.active_indices), 
    'n d -> b n d', 
    b=B_TT
)

metta/agent/lib/actor.py
```diff
@@ -0,0 +1,123 @@
...
+    def _forward(self, td: TensorDict):
+        input_1 = td[self._sources[0]["name"]] # Shape: [B*TT, hidden]
+        input_2 = td[self._sources[1]["name"]] # Shape: [B*TT, num_actions, embed_dim]
+
+        B_TT = input_1.shape[0]
+        num_actions = input_2.shape[1]
+
+        # input_1: [B*TT, hidden] -> [B*TT * num_actions, hidden]
+        # input_2: [B*TT, num_actions, embed_dim] -> [B*TT * num_actions, embed_dim]
+        input_1_reshaped = input_1.unsqueeze(1).expand(-1, num_actions, -1).reshape(-1, self.hidden)
```
daveey, line 52: use rearrange

metta/agent/util/distribution_utils.py
```diff
@@ -1,64 +1,83 @@
...
+        
     Returns:
-        Log probability of the value
+        Log probability of the specific action selection
     """
-    value = value.long().unsqueeze(-1)
-    value, log_pmf = torch.broadcast_tensors(value, logits)
-    value = value[..., :1]
-    return log_pmf.gather(-1, value).squeeze(-1)
+    sampled_action = sampled_action.long().unsqueeze(-1)
```
daveey, line 16: rearrange() if you can

daveey (approved):
i didnt look closely at the torch stuff, assuming it works. a few suggestions, but feel free to merge fixing.

metta/agent/lib/actor.py
Al-does: Thank you

configs/sim/all.yaml
Al-does: This is commented out on main for some reason. Any reason why?

metta/agent/util/distribution_utils.py
Al-does: Will do. Rob refactored this whole file and seems to have some material speed ups so will work with him on that then get back to applying better tensor reshaping.

PR #194, opened 2025-04-24
feat: new omegaconf resolvers

rjwalters:
# New Resolvers for OmegaConf

This PR adds several new resolvers to our OmegaConf configuration system:

## New Comparison Operators
- `gt` (greater than)
- `lt` (less than)
- `gte` (greater than or equal)
- `lte` (less than or equal)

## New Value Processing Functions
- `scale` - Scales values between ranges with different scaling methods:
  - Linear (default)
  - Logarithmic
  - Exponential
  - Sigmoid

- `iir` - Implements a first-order infinite impulse response filter for smoothing values over time

## Implementation Details
- All resolvers include comprehensive docstrings
- Added unit tests for all new functionality
- Maintained type checking for numeric operations
- Preserved integer handling where appropriate

These additions enable more sophisticated configuration options while maintaining the declarative style of our configuration system. The new resolvers are useful for smoothing operations, and conditional logic in configuration files.


daphnedemekas (commented):
There seem to be a lot of changes to mettagrid action handler files in this PR, is it possible that you need to merge in main to get the right diff?

PR #196, opened 2025-04-24
feat: add clang formatting for C++ files

rjwalters:
# Add clang formatting for C++ files

## Description
This PR introduces C++ code formatting standards using clang-format. While we already use ruff for Python formatting, this extends our code style enforcement to C++ files in the mettagrid directory.

## Changes
- Added a `.clang-format` file with Google-based style configurations and customizations
- Created a Makefile in /deps/mettagrid to:
  - Install clang-format tools (with platform detection)
  - Format C++ files while preserving Cython files
- Formatted all existing C++ files according to the new style guide
- Added relevant VSCode extensions to improve C++ and Cython development experience

## Notes
- The formatter is currently only applied manually via `make format` command
- Not enforcing C++ formatting in CI yet, but can be added in the future
- Settings prioritize readability with 4-space indentation and 120 character line limit


sasmith (approved):
Great!

My nexted, ordered preference is:
* Have an AoR holder for cpp formatting, so we have a single person deciding, versus committee. That could be you! Calling this out because having someone in charge outweighs my personal preference on the formatting.
* Pick an existing bag of formatting rules and be hesitant to deviate from them. Google's seems very justifiable, and I can understand why we would decide to deviate on line-length, but I'm less sure on other settings.
* Specific preferences
  * BreakBrforeBraces: Attach
  * IndentWidth: 2
  * If we have to wrap function parameters, I think I want them to all be on separate lines.
  * For enums, prefer to have be able to have one value per line.
  * For vector constants (like ObjectTypeNames), also one per line. I _think_ this is a "if it's too long to fit on one line, then I want one per line".
    * For these two, it's about scanability.

PR #197, opened 2025-04-24
feat: C++ testing and benchmarking

rjwalters:
## Overview
This PR adds a testing and benchmarking infrastructure for our C++ code in the mettagrid component using Google Test and Google Benchmark for analysis.

## Features
- **Unit Testing**: Implemented Google Test framework with example tests for GridObject
- **Benchmarking**: Added Google Benchmark framework with performance tests for core operations
- **Makefile Integration**: Created a unified Makefile with targets for:
  - Code formatting and linting
  - Running tests with automatic discovery
  - Running benchmarks with granular timing metrics
  - Platform-aware tool detection and installation

## Implementation Details
- Used C++14 standard to ensure compatibility with testing libraries
- Created modular directory structure for tests and benchmarks
- Implemented platform detection for macOS and Linux
- Added thorough benchmark metrics for core operations (creation, lookups, etc.)

## How to Test
- Run `make test` to execute unit tests
- Run `make benchmark` to run performance benchmarks
- Run `make format` to format C++ code

## Next Steps
**Before we create more tests, I'd like feedback on the approach and structure. This foundation can be extended to cover all components as we migrate from Cython to pybind11.**

We will also eventually connect tests and benchmarks to CI via github actions and bencher


deps/mettagrid/benchmarks/grid_object_benchmark.cpp
```diff
@@ -0,0 +1,147 @@
...
+    int idx = dist(gen);
+    int value = grid[idx];
+    benchmark::DoNotOptimize(value);
+  }
+
+  state.SetItemsProcessed(state.iterations());
+}
+BENCHMARK(BM_GridLookup)->Range(8, 1 << 10);
+
+BENCHMARK_MAIN();
```
sasmith, line 147: I think this just prints to stdout, and isn't connected to CI in any way yet, is that right?

sasmith (approved):
To repeat my comment, I think both the perf and unit tests aren't hooked up to CI -- is that right? Or are the unit tests being run automatically in a way I don't understand?

sasmith:
Regarding makefiles, my expectation is that we'll want to move to bazel, and makefile-like things will need to be rewritten when this happens. I don't have a timeline for that, and am happy with this makefile as we build up an understanding of what we need testing-wise.

PR #202, opened 2025-04-24
fix: wandb should check minor version

rjwalters:
Our convenience check of the wandb version was mistakenly checking the major version rather than the minor version.


rjwalters:
alternatively we could drop this check entirely -- but the code in main is incorrect as written

sasmith (approved):
Happy with either fix.

PR #205, opened 2025-04-24
fix: update monorepo package installation and tests

rjwalters:
## Overview
This PR fixes dependency handling in our new monorepo structure, simplifying installation and test workflows.

## Changes

- **Build System Improvements**:
  - Replaced multiple dependency installation steps with a streamlined build process
  - Added proper caching for compiled extensions to speed up CI
  - Updated the GitHub Actions workflow for better efficiency
  
- **Package Management**:
  - Created a new `Makefile` with standard targets (install, build, test, clean)
  - Fixed dependency resolution by adding `__init__.py` files in deps directory
  - Updated `pyproject.toml` to correctly include all subpackages
  - Created a new `Makefile` with standard targets (install, build, test, clean) (use is 100% optional)
     
- **Testing Enhancements**:
  - Added `conftest.py` to ensure proper path resolution during tests
  - Renamed problematic test files to prevent CI failures

- **Developer Experience**:
  - added dependency symlinks for local development
  - Added sanity checks on devops/setup_build script for import verification

replaces #190 and #204


devops/setup_build.sh
```diff
@@ -5,88 +5,153 @@
...
+    echo "Completed installation of $repo_name"
+}
+
+
+# Check if we're in the correct conda environment
+if [ "$CONDA_DEFAULT_ENV" != "metta" ]; then
+    echo "Error: You must be in the 'metta' conda environment to run this script."
+    echo "Please activate the correct environment with: \"conda activate metta\""
+    exit 1
+fi
```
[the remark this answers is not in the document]
rjwalters: My solution was to fake the ENV in CI but this does seem better. I need to confirm that github actions set CI and that this is not hallucination

.github/workflows/build.yml
sasmith, line 1: We've had a couple of times in the past week or so that builds have started failing, and clearing github cache has fixed them. I hope this fixes that! If not, we'll need to debug when it happens again.

deps/mettagrid/mettagrid/config/room/room.py
sasmith, line 1: Seems pretty unrelated.

deps/mettagrid/pyproject.toml
sasmith, line 1: We want to be able to publish mettagrid as a pypi package, separately from metta, so I think we want to keep this. I also instinctively wanted to "clean it up" due to switch to a monorepro, but I think the instinct is wrong.

deps/mettagrid/requirements.txt
sasmith, line 1: Same spirit as above -- that we want to be able to publish mettagrid separately. If you know that requirements.txt isn't needed for this, great. But I assume this is being deleted because you're collapsing mettagrid and metta, and we don't want to do that, quite.

deps/mettagrid/setup.py
```diff
@@ -48,30 +49,21 @@ def build_ext(srcs, module_name=None):
 
 debug = os.getenv("DEBUG", "0") == "1"
 annotate = os.getenv("ANNOTATE", "0") == "1"
-
-build_dir = "build"
-if debug:
-    build_dir = "build_debug"
-
+build_dir = "build_debug" if debug else "build"
```
sasmith, line 52: I wish Python had gone a different way with their ternary operator, since I find them less readable than c-style ternary operators.

That said, I think if we're using Python ternaries, we should semi-strongly prefer for the first option to be the default case. So if this were
`"build" if build_type == "normal" else "build_debug"`
I'd be more on board. As it stands, I think this is less readable than the original.

(This is all petty, and you can ignore.)

deps/mettagrid/setup.py
```diff
@@ -48,30 +49,21 @@ def build_ext(srcs, module_name=None):
...
-        "console_scripts": [
-            # If you want to create any executable scripts in your package
-            # For example: 'script_name = module:function'
-        ]
-    },
     include_dirs=[numpy.get_include()],
     ext_modules=cythonize(
         ext_modules,
         build_dir=build_dir,
+        nthreads=num_threads,
```
sasmith, line 65: Oh, were we putting this in the wrong place?

sasmith (changes requested):
Requesting changes regarding wanting to keep mettagrid as separately publishable.

deps/mettagrid/setup.py
rjwalters: debug is default false unless you set it in the env... I don't think we want to set a build type

we could do 
```
build_dir = "build" if not debug else "build_debug"
``

?

that doesn't feel better to me

deps/mettagrid/setup.py
rjwalters: > Here's the short answer:
> 
>     No, nthreads should be inside the cythonize() call, not passed to setup().
> 
> In your code:
> You are passing nthreads=num_threads inside the call to cythonize(), which is correct.
> You are not passing nthreads directly to setup(), which would have been wrong.

entry_points was empty so I removed it

.github/workflows/build.yml
rjwalters: I think this action has improved how we're using the cache. We can revisit it if we have issues

deps/mettagrid/mettagrid/config/room/room.py
rjwalters: this follows a convention we've used in other places when satisfying the ruff rule to not have {} or [] as a default value

deps/mettagrid/pyproject.toml
rjwalters: I will restore

PR #207, opened 2025-04-24
Several MettaScope improvements.

treeform:
* Show building inventory like we show agent inventory.
* Add converting and frozen state.
* Add special state for alters when they have no hearts.
* Better padding and atlas tech.
* When walls move their paths are showing using arrows instead of footprints.
* Show agent visibility as a group, but when one agent is selected just show that agent's visual field.
* Make the "Generate Replay" launch.json action also upload a replay to S3 and on Mac open a browser window to it. So that it acts like "Play" command.

image

image

image

# Graceful resource degradation: 

if "resouce.color.png" exists use it:

image

if not use "resouce.png" but apply the specific color tint: 

image

if not use "unkown.png" and apply the tint:

image


deps/mettagrid/player/src/main.ts
```diff
@@ -121,14 +124,24 @@ const AGENT_STYLES = {
...
 const INVENTORY = [
-  "armor",
-  "battery",
-  "blueprint",
-  "heart",
-  "laser",
-  "ore.blue",
-  "ore.green",
-  "ore.red",
+  // Agent inventory.
```
daveey, line 127: can we make add to mettagrid.yaml

game:
  items = ["armor", "battery", ....]
  colors = [red, green, blue]

then generate this from the config?

deps/mettagrid/player/src/main.ts
```diff
@@ -620,6 +678,21 @@ function drawObjects(replay: any) {
         y * TILE_SIZE,
         colorFromId(agent_id)
       );
+    } else if (typeName == "altar") {
```
daveey, line 681: all objects should show their inventory, we shouldn't special-case altars. all building should have:

empty, converting, has-output. they can have output and be converting. maybe a spinning gear, or smoke coming out of it, or something?

for now, you can re-use the icons for ones that you haven't designed yet.

tools/replay.py
```diff
@@ -28,9 +30,18 @@ def main(cfg):
         replay_job = dictconfig_to_dataclass(ReplayJob, cfg.replay_job)
         policy_record = policy_store.policy(replay_job.policy_uri)
         replay_helper = ReplayHelper(replay_job.sim, policy_record, wandb_run)
-        replay_path = f"{cfg.run_dir}/replays/replay.json.z"
-        replay_helper.generate_replay(replay_path)
-        print(f"Replay saved to {replay_path}")
+        epoch = policy_record.metadata["epoch"]
```
daveey, line 33: metadata.get("epoch", 0)

tools/replay.py
```diff
@@ -28,9 +30,18 @@ def main(cfg):
...
+        replay_helper.generate_and_upload_replay(
+            epoch,
+            cfg.run_dir,
+            cfg.run,
+            dry_run=cfg.trainer.get("replay_dry_run", False),
```
daveey, line 38: do we want dry_run here? i don't think we do

daveey (changes requested):
This looks great! A few changes, mostly wanting to get inventory out of code and into something that we can reference from player/c++/python 

excited to get these features!

deps/mettagrid/player/src/main.ts
treeform: All objects now show a “converting” icon:

image

Only the altar shows its inventory because Jack specifically requested it. The inventory of other objects matters much less, and I already show all inventories as bars. 

For now, I’m displaying the altar’s inventory because of the prominent heart visual. It makes sense to show both of its states:

image  
and  
image

Does this work for you?

tools/replay.py
treeform: Yes we do.

The replay generator is used in github actions Ci smoke testing. We don't want github actions CI to upload to S3 so we pass a dry run flag. If I don't pass the flag the CI breaks. 

Does this make sense?

deps/mettagrid/player/src/main.ts
treeform: We talked about this the resource system will do a scan of keys present and images present and try to pick the best image out of available images:

```
  // Example: "inv:heart" -> ["resources/heart.png", [1, 1, 1, 1]]
  // Example: "inv:ore.red" -> ["resources/ore.red.png", [1, 1, 1, 1]]
  // Example: "agent:inv:heart.blue" -> ["resources/heart.png", [0, 0, 1, 1]]
  // Example: "inv:cat_food.red" -> ["resources/unknown.png", [1, 0, 0, 1]]
```

deps/mettagrid/player/src/main.ts
treeform: // Create object image mapping for faster access.
  // Example: 3 -> ["objects/altar.png", "objects/altar.empty.png"]
  // Example: 1 -> ["objects/wall.png", "objects/wall.png"]

deps/mettagrid/player/src/main.ts
```diff
@@ -398,12 +396,70 @@ async function loadReplayText(replayData: any) {
...
+  replay.resource_inventory = new Map();
+  for (const key of replay.all_keys) {
+    if (key.startsWith("inv:") || key.startsWith("agent:inv:")) {
+      var image: string = key;
+      image = removePrefix(image, "inv:")
+      image = removePrefix(image, "agent:inv:");
+      var color = [1, 1, 1, 1]; // Default to white.
+      for (const [colorName, colorValue] of COLORS) {
+        if (image.endsWith(colorName)) {
+          if(drawer.hasImage("resources/" + image + ".png")) {
```
daveey, line 444: this is fine, but can you turn on some linter to catch things like no-space if(...)

daveey (approved):

PR #208, opened 2025-04-24
feat: use rich logging (colors) with hydra

### LINKED ISSUE #125: create and use a `get_metta_logger` utility
# Migrate codebase to use get_metta_logger utility

## Overview
Standardize all logging across our codebase to use the custom `get_metta_logger` utility function from `util.logging` instead of the standard Python logging module.

## Motivation
- Ensure consistent logging patterns across all modules
- Leverage custom formatting and handling provided by our logging utility
- Simplify log configuration and management
- Improve ability to filter and route logs in a standardized way

## Implementation Details
Replace:
```python
import logging
logger = logging.getLogger("module_name")
```

With:
```python
from util.logging import get_metta_logger
logger = get_metta_logger("module_name")
```

Ensure all logger method calls (info, warning, error, debug, etc.) are compatible with both interfaces to avoid unexpected behavior.

rjwalters:
## Changes
- Add support for rich colored logging with millisecond precision
- Create utility to override Hydra's default logger with our rich logger
- Add time remaining estimation to epoch progress logs (shows in sec/min/hours/days)

## Why
Hydra sets up its own logger by default, but our rich logger provides better visibility with colors and precise timestamps. The time remaining estimation helps developers monitor long-running training jobs more effectively.

fixes #125


tools/play.py
```diff
@@ -9,31 +8,37 @@
...
 
-logging.basicConfig(level="INFO")
-logger = logging.getLogger(__name__)
-
-
 class PlayJob(Config):
     sim: SimulationConfig
     policy_uri: str
 
+    def __init__(self, *args, **kwargs):
```
sasmith, line 22: Seems superfluous, but okay? Is this just so there's a stub for if we need it later?

metta/rl/pufferlib/trainer.py
```diff
@@ -195,9 +198,25 @@ def train(self):
...
+                time_str = f"{remaining_time_sec / 3600:.1f} hours"
+            else:
+                time_str = f"{remaining_time_sec / 86400:.1f} days"
+
             logger.info(
-                f"Epoch {self.epoch} - {self.agent_step} "
-                f"({100.00 * self.agent_step / self.trainer_cfg.total_timesteps:.2f}%)"
+                f"Epoch {self.epoch} - {self.agent_step} [{steps_per_second:.0f}/sec]"
+                f" ({100.00 * self.agent_step / self.trainer_cfg.total_timesteps:.2f}%)"
+                f" - {time_str} remaining"
```
sasmith, line 219: nice

sasmith (approved):

tools/play.py
rjwalters: this was necessary to convince Pylance to accept

```
play_job = PlayJob(cfg.play_job)
```

as valid

PR #213, opened 2025-04-25
fix: single agent action shape

rjwalters:
this PR will add unit tests for agent/utils/distribution_utils, correct a reshaping bug, and optimize performance.

in particular we want to correct behavior when we pass in `list( torch.Size([1, n]) )`
```
logits has len 2
logits[0] tensor([[-0.7894,  3.3176, -4.2624,  2.7705, -1.6243, -1.2506, -1.4018, -2.1241,
         -1.4263]]), has shape torch.Size([1, 9])
logits_to_probs(logit): tensor([[1.0065e-02, 6.1159e-01, 3.1225e-04, 3.5389e-01, 4.3674e-03, 6.3462e-03,
         5.4559e-03, 2.6496e-03, 5.3241e-03]])
torch.multinomial(logits_to_probs(logit), 1): tensor([[1]])
torch.multinomial(logits_to_probs(logit), 1).squeeze(): 1
action has shape torch.Size([2])
inference returns action with shape torch.Size([2])
Reshaping actions from torch.Size([2]) to (1, 2)
Case 1: If we have a flat tensor with correct total elements
Step 302 - Action shape: torch.Size([1, 2])
Step 302 - Action values: tensor([[1, 0]])
Step 302 - Action type: torch.int64
Step 302 - Action array shape: (1, 2)
Step 302 - Action array values: [[1 0]]
Step 302 - Agent 0 action: [1, 0]
```

### DIFF
--- a/metta/agent/metta_agent.py
+++ b/metta/agent/metta_agent.py
@@ -140,7 +140,7 @@ class MettaAgent(nn.Module):
         self.components["_value_"](td)
         return None, td["_value_"], None
 
-    def get_action_and_value(self, x, state=None, action=None, e3b=None):
+    def get_action_and_value(self, x, state=None, action=None, e3b=None, is_inference=False):
         td = TensorDict({"x": x})
 
         td["state"] = None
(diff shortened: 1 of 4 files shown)


rjwalters:
we seem confident that the fix should be is in `metta/agent/util/distribution_utils.py`:

```python
    batch = logits[0].shape[0]

    if action is None:
        action = torch.stack([torch.multinomial(logits_to_probs(logit), 1).reshape(batch) for logit in logits])
    else:
        action = action.view(batch, -1).T
```

vs `.squeeze()`

moving patch to #206 along with comments and tests

PR #218, opened 2025-04-26
fix: optimize setup_build and checkout_and_build scripts

rjwalters:
## Changes

- Refactored `checkout_and_build.sh` with a proper caching system that:
  - Preserves build artifacts between runs
  - Avoids unnecessary rebuilds when dependencies are already installed
  - Supports skipping the build process entirely with `SKIP_BUILD=1`

- Reorganized `setup_build.sh` to:
  - Properly call the checkout_and_build script
  - Clear the dependency cache when a fresh build is needed
  - Maintain consistent environment variable usage


rjwalters:
waiting on #233 and #234

devops/checkout_and_build.sh
```diff
@@ -1,7 +1,131 @@
...
+        echo "Repository content for $repo_name"
+        ls -al
+        cd ..
+
+        echo "Moving existing repository to cache_$repo_name"
+        mv "$repo_name" "cache_$repo_name"
+    fi
+
+    echo "Cloning $repo_name into $(pwd)"
+    git clone $repo_url
```
sasmith, line 48: This is an expensive operation, and makes this command a lot more expensive than it used to be. Is there a reason we would clone the repo if we already have it? E.g., versus the pulling we were doing previously?

devops/checkout_and_build.sh
```diff
@@ -1,7 +1,131 @@
...
+# Exit immediately if a command exits with a non-zero status
+set -e
+
+# Repository URLs defined as variables
+FAST_GAE_REPO="https://github.com/Metta-AI/fast_gae.git"
+PUFFERLIB_REPO="https://github.com/Metta-AI/pufferlib.git"
+CARBS_REPO="https://github.com/kywch/carbs.git"
+WANDB_CARBS_REPO="https://github.com/Metta-AI/wandb_carbs.git"
+
+# Function to install a repository dependency
```
sasmith, line 26: Can you add a comment regarding caching? I haven't thought hard around how build systems know when to reuse their build artifacts, and a couple sentences on why the way we're caching things makes sense would help. I'd guess this is

"In case the repo already exists, this preserves build artifacts between the old and new repos. Caching for these artifacts is based on timestamps, and since timestamps are preserved in operations like `git clone`, the caching should work as expected."

devops/checkout_and_build.sh
sasmith: (I see this is moved from `setup_build`, so I guess my question retroactively applies there.)

sasmith (changes requested):
Requesting changes since I think this will slow down check_and_build by a lot.

setup_build currently takes 75s to run. checkout_and_build takes 15s, and almost all of that is because we're recompiling the cython every time (I don't know why we consistently think the files have changed).

At least a large chunk of the speed difference is in the `git clone` -- if I comment that out (and the proceeding caching that's removing the directory), the time for setup_build drop from 75s to 40s.

devops/checkout_and_build.sh
rjwalters: I think you are thinking about cache as it pertains to Github Actions?

> When your workflow runs:
> 
> GitHub Actions first checks if a cache exists with the exact key
> If found, it restores that cache to the specified path (cache hit)
> If not found, it checks for partial matches using restore-keys (partial cache hit)
> If nothing matches, it creates a new cache after the job completes (cache miss)

In the context of these shell scripts, I am just writing a local file at /deps/.built so that we can skip building in `checkout_and_build` on subsequent calls to `train.sh` or `sweep.sh`

devops/checkout_and_build.sh
rjwalters: building is slower than the github operations so I didn't think this was important. we can add an optimization if a pull is sufficient

rjwalters: I added the optimization

devops/checkout_and_build.sh
sasmith: oooooh. I'd missed the `.built`.

Okay, then I think the comment I want is a version of
"""
The caching here is primarily focused on what happens in github actions. In the context of normal development, we rely on touching the `.built` file and skipping all of this.
"""

devops/checkout_and_build.sh
```diff
@@ -1,7 +1,130 @@
...
+    # Run the build command
+    echo "Building with command: $build_cmd"
+    eval $build_cmd
+    
+    cd ..
+    echo "Completed installation of $repo_name"
+}
+
+
+SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
```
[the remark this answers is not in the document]
sasmith: `readlink -f` appears to exist on my machine.

sasmith (approved):

devops/checkout_and_build.sh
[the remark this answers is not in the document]
rjwalters: me too

PR #222, opened 2025-04-26
Migration of SimSam map generation from mettagrid

Dmitry273:


Dmitry273:
this PR adds Simplex Sampler or SimSam map generation method along with few .yaml configs that work with it
here's the previous PR to mettagrid:
https://github.com/Metta-AI/mettagrid/pull/79

deps/mettagrid/mettagrid/resolvers.py
```diff
@@ -228,6 +232,7 @@ def register_resolvers() -> None:
     """
     OmegaConf.register_new_resolver("if", oc_if, replace=True)
     OmegaConf.register_new_resolver("uniform", oc_uniform, replace=True)
+    OmegaConf.register_new_resolver("triangular", oc_triangular, replace=True)
```
rjwalters, line 235: can you add a unit test for this new resolver?

Dmitry273: done

deps/mettagrid/mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,184 @@
...
+        terrain = (terrain + 1) / 2  # changes range from [-1,1] to [0,1]
+        terrain = terrain.reshape((self._height, self._width))
+        terrain = (
+            terrain**layer.saturation
+        )  # saturates pattern with walls. Helpful since base noise is balanced to be 50/50
+        # Saturation 0 makes neutral terrain with 0 walls, saturation ~>10 fills ~everything with walls.
+
+        return terrain
+
+    def fix_map(self, terrain: np.ndarray) -> None:
```
sasmith, line 108: This is much too generic of a name for this function. I believe what you're doing here is ensuring that all of the empty tiles are connected, and trying to do so by carving the least energetic path through the walls -- is that right?

sasmith (commented):
Sorry for the slow work in reviewing. Not done yet, but figured I should send what I have.

deps/mettagrid/mettagrid/map/scenes/simplex_sampler.py
Dmitry273: Yes, "connect_empty_regions" would be a good name?

deps/mettagrid/mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,184 @@
...
+    def __init__(
+        self,
+        children: list[Any],  # it has something to do with Slava's implementation of recursive generation
+        # probably parameters for deeper layers, he should know better
+        layers: list[Layer],  # layers of sampling functions each of which dictate how generated noise is sampled.
+        # All of them are then combined to produce the end result
+        seed: MaybeSeed = None,
+        cutoff: float = 0.27,  # final step parameter that dictates at which point the whole map will be cut off to
+        # walls and empty spaces. Can be between 0 and 1
+        force_seed: int = 0,  # Option to use fixed seed.
```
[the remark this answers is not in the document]
Dmitry273: I've used it to debug or to generate gifs of the same-seed maps but with slowly changing sampling function. I can create a benchmark set of maps that will slowly change from the one where agents work properly to the one where they fail. Maybe it could help training team to troubleshoot. That seemed potentially useful, so I've added this as an explicit parameter

deps/mettagrid/mettagrid/map/scenes/simsam_functions.py
```diff
@@ -0,0 +1,297 @@
...
+def squeezed_noise(
+    x: int,
+    y: int,
+    width: int,
+    height: int,
+    zoom: float = 0.1,
+    squeeze: float = 1.5,
+    angle_theta: float = 0.25,
+) -> tuple[float, float]:
+    shift_from_center_for_x = 0
```
sasmith, line 75: These are zero, so let's clean up the equations below?

deps/mettagrid/mettagrid/map/scenes/simsam_functions.py
```diff
@@ -0,0 +1,297 @@
...
+    sn = math.sin(alpha)
+
+    xi = (+cs * (x - (0.5 + shift_from_center_for_x) * width) + sn * (y - (0.5 + shift_from_center_for_y) * height)) + (
+        0.5 + shift_from_center_for_x
+    ) * width
+    yi = (-sn * (x - (0.5 + shift_from_center_for_x) * width) + cs * (y - (0.5 + shift_from_center_for_y) * height)) + (
+        0.5 + shift_from_center_for_y
+    ) * height
+
+    xi = (xi - 0.5 * width) * zoom**2 / squeeze
```
sasmith, line 88: The zoom being squared is hard to understand, especially as it breaks the convention set in `xy_noise` of being linear.

deps/mettagrid/mettagrid/map/scenes/simsam_functions.py
```diff
@@ -0,0 +1,297 @@
...
+    y: int,
+    width: int,
+    height: int,
+    zoom: float = 0.1,
+    squeeze: float = 1.5,
+    angle_theta: float = 0.25,
+) -> tuple[float, float]:
+    shift_from_center_for_x = 0
+    shift_from_center_for_y = 0
+    alpha = 2 * math.pi * angle_theta
```
sasmith, line 77: Let's have people just pass the value in radians.

deps/mettagrid/mettagrid/map/scenes/simsam_functions.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+# function used in "noise" generator
+# generates additional noise, but scaled globally, tilted arbitrary at angle theta and squeezed
+def squeezed_noise(
+    x: int,
+    y: int,
+    width: int,
+    height: int,
+    zoom: float = 0.1,
```
sasmith, line 71: I don't think we should change it now, but my instinct is that we'd want to have functions that do more minimal things (e.g., rotate, squeeze) and then compose them, rather than having a single function that combines all of these at once.

deps/mettagrid/mettagrid/map/scenes/simsam_functions.py
sasmith, line 1: For the most part, I'm not sure how to review this -- I'm not used to reviewing semi-arbitrary math that ends up making neat looking patterns.

I think this is largely understandable now, or at least I think we've invested the right about in that direction. So I think we should be merging this to let people play around.

deps/mettagrid/tests/test_resolvers.py
```diff
@@ -89,6 +90,26 @@ def test_uniform_resolver():
             assert min_val <= val <= max_val
 
 
+def test_triangular_resolver():
```
sasmith, line 93: I see we have a test for the uniform distribution, but I think this unit test seems sort of low value -- we're effectively unit testing numpy's random number generation. I'd be satisfied with just ensuring that we're calling numpy's random number generator. Technically that would be more fragile, but 🤷 .

I'd also be comfortable with no unit test for this.

deps/mettagrid/mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,200 @@
...
+        # layers of sampling functions each of which dictate how generated noise is sampled.
+        # All of them are then combined to produce the end result
+        layers: list[Layer],
+        seed: MaybeSeed = None,
+        # Every square will get a value dictating how empty it is.
+        # If the value is above this cutoff, we'll leave it empty. Otherwise we'll fill it with a wall.
+        # So, e.g., cutoff = 0 => no walls; cutoff = 1 => all walls
+        cutoff: float = 0.27,
+        # Option to use fixed seed.
+        # Useful for generating slowly changing variants of one benchmark. 0 will use random seed.
```
sasmith, line 45: Can we make `None` mean "don't use" instead of 0? Mostly to make this symmetric with `seed`.

deps/mettagrid/tools/map/interactive_visualization.ipynb
sasmith, line 1: It's hard for me to review this just by reading, but I like the idea of giving a demo notebook.

deps/mettagrid/tools/map/mapgen_manual_drawing.py
```diff
@@ -0,0 +1,58 @@
+from omegaconf import OmegaConf
+from PIL import Image
+
+from mettagrid.map.mapgen import MapGen
+
+# import numpy as np
```
sasmith, line 6: Can you clean up unneeded comments?

sasmith (changes requested):
This looks close.

I didn't heavily review the .yaml files of the notebook -- I think it's reasonable for folks to get a sense of those through usage.

deps/mettagrid/tests/test_resolvers.py
Dmitry273: @rjwalters asked me to add this, I was a bit surprised, but did it anyway since it wasn't hard
here's the original convo: https://github.com/Metta-AI/metta/pull/222#pullrequestreview-2797109176

deps/mettagrid/mettagrid/map/scenes/simsam_functions.py
Dmitry273: It's hard to make a lot of variety while making samplers that could be easily understood: you usually end up with trivial examples

I think mostly because it's a bit counterintuitive to work with this: you need to look for derivatives and isolines, not absolute values

That's why I've made a notebook that allows to just play with them or with any new ones in the future. Those functions are closer to "discovered good cases", not some functions to use anywhere else outside of this terrain generation context

deps/mettagrid/mettagrid/map/scenes/simplex_sampler.py
Dmitry273: Done

deps/mettagrid/mettagrid/map/scenes/simsam_functions.py
Dmitry273: It makes it slightly more annoying to then initiate it in yaml configs since it requires to work with pi constant there instead of in my section. It was mostly the reason behind using those units instead of radians

PR #231, opened 2025-04-28
fix: OmegaConf is not a type

rjwalters:
# Fix OmegaConf Type Annotations

OmegaConf itself is not a type that should be used in type annotations. OmegaConf creates objects that are instances of either `DictConfig` or `ListConfig`.

## Changes

- Replace instances of `OmegaConf` in type annotations with `DictConfig | ListConfig`
- Fix related type hints to match actual usage
- Remove unnecessary multi-line formatting in several files
- Update type handling in `Config` class to better handle string keys


metta/util/config.py
```diff
@@ -23,23 +18,28 @@ class Config(BaseModel):
...
         # allow `Config(DictConfig)` or `Config(dict)` as shorthand for .from_dictconfig(...)
         if len(args) == 1 and not kwargs and isinstance(args[0], (DictConfig, dict)):
             raw = args[0]
             data = OmegaConf.to_container(raw, resolve=True) if isinstance(raw, DictConfig) else dict(raw)
-            try:
-                super().__init__(**data)
-            except ValidationError:
-                # re-raise so traceback points here
-                raise
+            # Ensure data is a proper dict with string keys
```
jacklionheart, line 32: have you considered using an assert( x isInstance) to satisify pylance?

jacklionheart (approved):

PR #238, opened 2025-04-29
feat: map previews

rjwalters:
This adds a map preview link to mettascope during training.


metta/sim/map_preview.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    logger.info("Building map preview...")
+
+    sim = cfg.sim
+    env_cfg = config_from_path(sim.env)
+
+    try:
+        vecenv = make_vecenv(env_cfg, sim.vectorization, num_envs=1, render_mode="human")
+
+        # Check if vecenv was created successfully
```
daveey, line 33: I don't think we need this. if you are worried about make_vecenv failing, add the check in there

metta/sim/map_preview.py
```diff
@@ -0,0 +1,104 @@
...
+            "action_names": env.action_names(),
+            "object_types": env.object_type_names(),
+            "inventory_items": env.inventory_item_names(),
+            "map_size": [env.map_width, env.map_height],
+            "num_agents": vecenv.num_agents,
+            "max_steps": 1,
+            "grid_objects": list(env.grid_objects.values()),
+        }
+
+        # Clear any actions
```
daveey, line 51: I don't think grid_objects should have any actions, if they do, let's figure out why and fix it

metta/sim/map_preview.py
```diff
@@ -0,0 +1,104 @@
...
+            if "action" in obj:
+                del obj["action"]
+
+        # Compress data with deflate
+        preview_data = json.dumps(preview)  # Convert to JSON string
+        preview_bytes = preview_data.encode("utf-8")  # Encode to bytes
+        compressed_data = zlib.compress(preview_bytes)  # Compress the bytes
+
+        # Create directory and save compressed data
+        preview_path = f"{cfg.run_dir}/replays/replay.0.json.z"
```
daveey, line 62: how about we use tempfile

metta/sim/map_preview.py
```diff
@@ -0,0 +1,104 @@
...
+        os.makedirs(os.path.dirname(preview_path), exist_ok=True)
+        with open(preview_path, "wb") as f:
+            f.write(compressed_data)
+
+        # Upload to S3
+        try:
```
daveey, line 68: i don't see the point of try/catch here. we don't need to clean up vecenv, let's just crash

metta/sim/map_preview.py
```diff
@@ -0,0 +1,104 @@
...
+            # Check if running in CI environment - abort upload if in CI
+            is_ci = os.environ.get("CI", "").lower() in ("1", "true", "yes")
+            if is_ci:
+                logger.info("Running in CI environment, skipping S3 upload but keeping local file")
+                return
+
+            preview_url = f"replays/{cfg.run}/replay.0.json.z"
+            s3_client = boto3.client("s3")
```
daveey, line 76: can we move this into util/s3_client.py:upload_file() to get the magic constants out of here

metta/sim/map_preview.py
daveey, line 62: i don't want this to conflict with other replays. either use a tempfile or put it into rundir/maps/

metta/sim/map_preview.py
```diff
@@ -0,0 +1,104 @@
...
+        wandb_run: Weights & Biases run object for logging
+        logger: Logger object for logging messages
+    """
+    logger.info("Building map preview...")
+
+    sim = cfg.sim
+    env_cfg = config_from_path(sim.env)
+
+    try:
+        vecenv = make_vecenv(env_cfg, sim.vectorization, num_envs=1, render_mode="human")
```
daveey, line 31: why make a vecenv instead of passing it in? the trainer already has a vecenv

tools/train.py
```diff
@@ -42,6 +43,12 @@ def train(cfg, wandb_run, logger: Logger):
     train_job = TrainJob(cfg.train_job)
 
     policy_store = PolicyStore(cfg, wandb_run)
+
+    upload_map_preview(cfg, wandb_run, logger)
+
+    if "dry_run" in cfg and cfg.dry_run:
```
daveey, line 49: i dont think we want this

daveey (changes requested):
thanks for doing this!

metta/sim/map_preview.py
rjwalters: this code runs before the trainer is created

The alternative to creating a vecenv here would be to call this from inside the init phase of PufferTrainer @ https://github.com/Metta-AI/metta/blob/8079c1718f3b8bc8a1ae29d4efa6ac1413240d3e/metta/rl/pufferlib/trainer.py#L79

I think it is cleaner to just create a second vecenv since it makes this more self contained

metta/sim/map_preview.py
rjwalters: the idea here is to provide a better error message if something fails; this seems like a code style nit since this isn't performance critical code?

metta/sim/map_preview.py
rjwalters: yeah I need to unpack some of these zipped records and figure out why rendering stopped working - see also #271

metta/sim/map_preview.py
rjwalters: I think we are safe putting this at "epoch 0" since no replay will ever be calculated until after some training has been computed... we may benefit from having a consistent name pattern in mettascope vs creating a new convention for maps. On the other hand if we end up storing collections of multiple maps (like multiple random draws as an example) then adding a new map storage location makes sense. 

I think my preference is to leave it at epoch zero until we expand the scope of what we're trying to accomplish

metta/sim/map_preview.py
rjwalters: I think if we are creating a dedicated vecenv to generate the map rather than using the one in trainer it feels like a good idea to clean it up rather than possibly have it around while training is running

tools/train.py
rjwalters: I have used this to check the map preview without training -- I can move this inside upload_map_preview if you prefer?

metta/sim/map_preview.py
rjwalters: https://app.asana.com/1/1209016784099267/project/1209041170403492/task/1210139187163834?focus=true

for now we will fix this in mettascope

metta/sim/map_preview.py
daveey: Yeah, more reason to just create an env. I'm really not a fan of try/catch in python code

daveey: And this is being called from main, so it's fine to just crash

rjwalters: this is at /sim/map_preview and may be called from other places in the future. I don't see the harm in trying to be robust in the face of errors -- and failing to upload a preview doesn't seem like something that we would always want to force an exit in all contexts.

I don't see any downside to handling the errors in this way?

metta/sim/map_preview.py
rjwalters: why is creating a vecenv directly a problem? if I create a MettaGridEnv then we will just create a vecenv in it's init

tools/train.py
rjwalters: ok - I will remove it

rjwalters:
@daveey the original concept for an "Epoch zero map preview" that was mettascope compatible was to provide me with a quick way to debug map generation. There are many other ways I can accomplish this (e.g. ascii or matplotlib output) that bypass mettascope.

I agree with you that we will want a map viewer that shows the entire variety of maps used in a training curriculum but I think that is a larger scope and requires some additional design work.

I think our best options are either to merge this as is as a temporary solution or to instead just **close this PR**

In either case I think we should task @treeform to design an improved map viewing functionality in mettascope.

metta/sim/map_preview.py
daveey: false

metta/sim/map_preview.py
daveey: code complexity for a future use case that we may not do. i agree this is just aesthetics

daveey (changes requested):
We can discuss this in person. I don't

metta/sim/map_preview.py
rjwalters: ok - when I wrote this I started top down from the sim scripts and stopped when things worked. I did not dig into exactly what "make_venenv' was doing in the context of preparing the map preview. Looking more closely, `make_venenv` was basically just calling hydra instatiate which then creates a MettaGridEnv (and not the other way around). We can instead just create the MettaGridEnv object directly using hydra.instantiate -- since we pass in an env_cfg that derives from configs/env/mettagrid/mettgrid with 
``
_target_: mettagrid.mettagrid_env.MettaGridEnv
```

This entire approach is still just resolving one sampled map, and then somewhat abusing the Mettascope interface to display as a zero step length replay. So lots of room for improvement.

daveey: just change it to 

MettaGridEnv(env_cfg, render_mode="human")

make_vecenv calls pufferlib.vector.make which does more than instantiate. i don't understand why you are resisting this change. it's simple, reduces complexity, and makes me happy.

daveey (changes requested):
just vecenv -> MettaGridEnv and it'll be ready

rjwalters:
now with 100% less hydra

PR #242, opened 2025-04-29
Fix success action tracking.

treeform:
* Fix success action tracing in C++ and Cython. !!BIG!!
* Fix the fact that some vars need to be pre simulation step others post simulation.
* Update trace following and resource bar touches other resource bars - visual changes.


daveey (approved):

deps/mettagrid/mettagrid/grid_env.pyx
```diff
@@ -173,6 +173,8 @@ cdef class GridEnv:
         for p in range(self._max_action_priority + 1):
             for idx in range(self._agents.size()):
                 action = actions[idx][0]
+                # Clear the success flag, as we can continue early.
+                self._action_success[idx] = False
```
sasmith, line 177: This looks wrong to me -- I think it needs to be at a higher level. There are two ways of thinking about this, but they're basically the same.

First way: not all places we continue are indicative of failure. Some of them are just "not applicable".

Second, more concrete: We run this loop for every priority, which means we'll end up clearing the action multiple times. So, e.g., if an agent successfully attacks (attacks have the earliest priority), we'll later clear the success flag.

I assume the right solution is something like

self._action_success[:] = False

before we get into the loop. But maybe you'll need to actually set it in a loop.

sasmith: If you're able to add a test, that would be great. I haven't tried testing grid_env, and it's in cython, so this may be a hard ask and isn't a requirement.

deps/mettagrid/player/data/trace/get_output.png
sasmith, line 1: I think this and other visual parts of the PR could / should have been separated from the "fix action_success" part. In this case, the total review isn't that big, so it's not that big of a deal; but it would be easier to review if all the pieces of the PR were more alike.

deps/mettagrid/mettagrid/actions/get_output.hpp
```diff
@@ -43,14 +43,18 @@ class GetOutput : public ActionHandler {
...
-      actor->update_inventory(static_cast<InventoryItem>(i), converter->inventory[i]);
-      converter->update_inventory(static_cast<InventoryItem>(i), -converter->inventory[i]);
+      // Only take resources if the converter has some.
+      if (converter->inventory[i] > 0) {
+        // The actor will destroy anything it can't hold. That's not intentional, so feel free
+        // to fix it.
+        actor->stats.add(InventoryItemNames[i], "get", converter->inventory[i]);
+        actor->update_inventory(static_cast<InventoryItem>(i), converter->inventory[i]);
+        converter->update_inventory(static_cast<InventoryItem>(i), -converter->inventory[i]);
+        return true;
```
sasmith, line 53: This is wrong in the case that there's more than one type of resource we're getting, since it will return on the first resource that's present. The intention is that if a converter produces A and B, we'll get both.

I think your intention was to make it so that if you try to `get`, and there are no resources to get, then it's a failure -- is that right? That seems reasonable, and consistent with it being a failure to attack empty terrain. We just need to track whether the agent got anything in a variable versus returning early.

configs/user/me.yaml
```diff
@@ -1,6 +1,7 @@
 # @package __global__
```
sasmith, line 1: nitpick: consider adding a comment at the top, explaining this is your file. The name isn't obvious.

sasmith (changes requested):

deps/mettagrid/mettagrid/grid_env.pyx
treeform: Thats what I did at first too but:

self._action_success[:] = False

is not valid as its not a numpy array but a C++ vector. I had to get an ugly for loop there that was identical to this for loop. I could revert it back if you like.

deps/mettagrid/player/data/trace/get_output.png
treeform: Ok I will figure out how to work with graphite and break it up in multiple PRs.

deps/mettagrid/mettagrid/actions/get_output.hpp
treeform: Good point i'll revisit this. So far nothing returns multiple resources though.

configs/user/me.yaml
treeform: Ok will add a comment saying `# Hey this is my file`

PR #243, opened 2025-04-29
Torch profile (trace) generation in trainer

Al-does:
Added a torch profile class that can be used as context in trainer. Set profiler_interval_epochs to 0 to turn off.

- Unless profiler_interval_epochs is set to 0, the profiler will always run at 50 epochs (hardcoded) to 1) give early feedback on a run while allowing the overall epoch window to be very long for perf considerations and 2) to give plenty of time for torch's warm up runs. We may want to reconsider this number if we start compiling under max-autotune.
- Profiles are saved as json.gz files locally in train_dir/<your_run>/torch_traces/ and in S3 at torch_traces/<your_run>/
- A link to download the S3 file is dropped into wandb under "torch traces".
- To view traces, go to chrome://tracing (or arc://tracing if you happen to use that browser which is fine) and select load. Navigate traces using WASD on your keyboard.

Future work could include support for TensorBoard.


daveey (approved):

daveey:
looks good but failing tests

PR #255, opened 2025-04-30
Fix some broken config targets

sasmith:
It looks like Slava's diff to merge our config trees, and George's diff to add new training envs, passed each other. Right now we don't have tests to confirm that all configs are valid.

I'd feel better about myself if I were adding a test here, but I think the right path forward is more along the lines of Jack's (and Slava?)'s work to improve our configuration in general.


sasmith:
* **#255** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/255?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/255?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jacklionheart (approved):
Re testing: My personal beliefs are:

(1) We need to get to a point where we can test training a process *without using the configs/ directory* 
(2) Separately, we need to get to a point where we can "validate" every config file in confgs/ on all commits

PR #260, opened 2025-04-30
Convert observation_encoder to cpp

sasmith:
Converts ObservationEncoder to cpp, including making is a cpp class instead of a Python class. This included a small amount of pushing on boundaries (e.g., where is the ObservationEncoder created), but all of the boundaries felt sort of arbitrarily placed, and so it seemed reasonable to shift them in order to let this be a fully cpp class.

Tested by compiling and smoke-test training on my laptop.


deps/mettagrid/mettagrid/observation_encoder.hpp
```diff
@@ -0,0 +1,82 @@
...
+        _offsets.resize(ObjectType::Count);
+        _type_feature_names.resize(ObjectType::Count);
+
+        _type_feature_names[ObjectType::AgentT] = Agent::feature_names();
+        _type_feature_names[ObjectType::WallT] = Wall::feature_names();
+
+        // These are different types of Converters. The only difference in the feature names
+        // is the 1-hot that they use for their type. We're working to simplify this, so we can
+        // remove these types from code.
+        for (auto type_id : {ObjectType::AltarT, ObjectType::ArmoryT, ObjectType::FactoryT,
```
daveey, line 27: might be a good candidate for next cleanup

deps/mettagrid/mettagrid/observation_encoder.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+        // Set the offset for each feature, using the global offsets.
+        for (size_t type_id = 0; type_id < ObjectType::Count; ++type_id) {
+            for (size_t i = 0; i < _type_feature_names[type_id].size(); ++i) {
+                _offsets[type_id].push_back(features[_type_feature_names[type_id][i]]);
+            }
+        }
+    }
+
+    void init(unsigned int obs_width, unsigned int obs_height) {
```
daveey, line 53: we can move this to the constructor if we create the encoder inside grid_env instead of passing it in.

daveey (approved):
Looks good. I think you can move the construction into grid_env, rather than passing it in. that'll let you remove some code.

deps/mettagrid/mettagrid/observation_encoder.hpp
sasmith: This is one of those "yes, and" things.

It turns out we don't even use `obs_width` and `obs_height`, so we could delete this all together. I figured I'd leave it alone for the moment since I think on balance we want to have more of this in the observation encoder.

So, agree we can clean this up, but I'm not going to do it in this diff, since I'm not sure where we should land.

sasmith: Actually, I'll remove it now and add something back once we need it.

deps/mettagrid/setup.py
```diff
@@ -40,6 +40,7 @@ def build_ext(srcs, module_name=None):
     build_ext(["mettagrid/actions/get_output.pyx"]),
     build_ext(["mettagrid/objects/agent.pyx"]),
     build_ext(["mettagrid/objects/constants.pyx"]),
+    build_ext(["mettagrid/objects/has_inventory.pyx"]),
```
sasmith, line 43: This started being needed for compilation. I didn't dig in far, but assume it has to do with how other requirements were changed.

metta.code-workspace
```diff
@@ -144,7 +144,6 @@
 		"files.exclude": {
 			"train_dir": true,
 			"wandb": true,
-			"deps": true,
```
sasmith, line 147: Maybe this should have been a separate diff, but cursor was ignoring things in deps, including deps/mettagrid. I don't know why it wasn't happening before.

PR #261, opened 2025-04-30
EFS proxy opentofu configs

berekuk:
# Tailscale

This PR (`tailscale_acl.acl` resource in `tailscale.tf`) **replaces the Tailscale ACL**. Be careful when applying configs, you probably will need to update `tailscale.tf` first.

# OpenTofu vs Terraform
I've used OpenTofu but I expect both OpenTofu and Terraform would work identically.

The lockfile is generated with OpenTofu though, and Terraform might try to reformat it.

(My impression is that OpenTofu is doing well, but I haven't followed the fork story closely in the last year.)

# Performance

Latency is high. Copying a single 300KB file to the local share takes 12 seconds for me. OTOH, it's fast when I mount it on EC2, so it might be better for others.

# Installing on macs

Not automated yet; this PR has Terraform configs for AWS. I'll add a script for macs soon.


berekuk:
* **#261** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/261?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/261?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

devops/tf/efs-proxy/.gitignore
```diff
@@ -0,0 +1,3 @@
+.terraform
+*.tfstate
+*.tfstate.backup
```
berekuk, line 3: I've used local state, when I did tests, for now.

If we're going to adopt Terraform, we should either:
- use Spacelift or something like it
- or just an S3 bucket

I've tried both options in the past; S3 bucket is ok there's a single admin. Spacelift is better when there's some shared responsibility over devops.

devops/tf/efs-proxy/efs.tf
```diff
@@ -0,0 +1,37 @@
+# Create EFS - will be imported
+resource "aws_efs_file_system" "efs" {
+  tags = {
+    # Name = "MettaTest"
+  }
+}
+
+import {
+  to = aws_efs_file_system.efs
+  id = "..." # Enter EFS ID here
```
berekuk, line 10: This needs to be done once, so it's not in variables. Maybe could be in variables, I'm not sure.

Ideally, if we're starting to use TF, we should import EFS here and then treat it as managed by TF (move away from click-ops). Then maybe this folder should be `devops/tf/efs`, not `devops/tf/efs-proxy`?

The `efs-proxy` dir name is because I'm treating this folder as a "stack" (see e.g. https://docs.spacelift.io/concepts/stack); monolith terraform configurations, even with modules, quickly become too slow, so it's good to separate them. But I'm not sure if this stack should be "EFS proxy only" or "everything EFS related", or something even larger. And it also manages Tailscale ACL, which can't be done in parts...

devops/tf/efs-proxy/tailscale.tf
```diff
@@ -0,0 +1,54 @@
...
+        "::/0" = ["tag:efs-proxy"],
+      }
+    },
+    acls : [
+      {
+        action = "accept",
+        src    = ["autogroup:member"],
+        dst = [
+          # covers the proxy’s EFS routes
+          # note: app connector docs recommend "autogroup:internet:*" here, but it doesn't work, I think because EFS drive is on the local network
```
berekuk, line 29: I spent 3 hours banging my head against this, Tailscale documentation doesn't mention this, ChatGPT also didn't know what to do.

berekuk:
Scripts for macbooks are in `devops/sandbox` - `setup_machine.py` installs EFS tools (by calling `setup_efs.sh`), and `mount_efs.sh` does the actual mount.

devops/sandbox/mount_efs.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+# TODO - update with the actual EFS endpoint
+ENDPOINT="fs-0959f49bf925f0023.efs.us-east-1.amazonaws.com"
```
berekuk, line 4: this needs to be replaced with the actual EFS domain

daveey: can we read this from ~/.metta.env

daveey (approved):
Looks great!

devops/tf/efs-proxy/.gitignore
daveey: I haven't used spacelift, but why not just check into github?

devops/tf/efs-proxy/efs.tf
daveey: down to try spacelift. i gave you credentials to our infra account. wanna create spacelift free account and add me as a user?

devops/tf/efs-proxy/.gitignore
berekuk: tfstate files are very sensitive, can often contain passwords, secret env variables and so on.
Also, bad things will happen if you run terraform with a stale state file.

devops/tf/efs-proxy/efs.tf
berekuk: I tried to sign up for a new spacelift account with github, it requested access to metta repo that you need to confirm. (I could sign up with google email instead, but github integration will be useful down the road.)

devops/sandbox/mount_efs.sh
berekuk: Ideally we should be reading it from Terraform outputs. But end users of this script won't have access to terraform state.
o3 says I can put it to AWS SSM Parameter Store, I'll try that.

devops/sandbox/Brewfile
```diff
@@ -1,13 +1,14 @@
 # Developer tools
 tap "withgraphite/tap"
+brew "amazon-efs-utils"
```
sasmith, line 3: From https://github.com/aws/efs-utils:
"""
Note that this will ONLY work on EC2 instances ... not local Mac computers.
"""

Going to delete this.

sasmith:
`setup_machine.py` is for people's Macs, and the EFS code doesn't appear to work there. Can you take another swing at it?

devops/sandbox/Brewfile
berekuk: It worked for me on my mac without issues though.

I think the note is about lack of connectivity, not about something like "AWS patches their macOS", and this PR fixes that.

berekuk:
> `setup_machine.py` is for people's Macs, and the EFS code doesn't appear to work there. Can you take another swing at it?

Have you tried it? If so, how exactly did it fail?

PR #278, opened 2025-05-01
Bug in policy selector if policy doesn't have eval scores

daphnedemekas:
Rob ran into a bug with the policy selector which I fixed here! If the policy doesn't have eval_scores in the metadata, this will not break anymore


metta/agent/policy_store.py
```diff
@@ -114,10 +114,11 @@ def _policy_records(self, uri, selector_type="top", n=1, metric: str = "score"):
             return [random.choice(prs)]
 
         elif selector_type == "top":
-            if metric not in prs[0].metadata:
+            metadata = prs[0].metadata
```
sasmith, line 117: Seems worth a comment about why we're just looking at the latest when doing this check.

metta/agent/policy_store.py
```diff
@@ -137,7 +138,7 @@ def _policy_records(self, uri, selector_type="top", n=1, metric: str = "score"):
             logger.info(f"{'Policy':<40} | {metric:<20}")
             logger.info("-" * 62)
             for pr in top:
-                logger.info(f"{pr.name:<40} | {pr.metadata.get(metric, 0):<20.4f}")
+                logger.info(f"{pr.name:<40} | {policy_scores[pr]:<20.4f}")
 
             return top[-n:]
```
sasmith, line 143: `top` is already defined as being only `n` elements, so we don't need the slice here.

metta/agent/policy_store.py
```diff
@@ -114,10 +114,11 @@ def _policy_records(self, uri, selector_type="top", n=1, metric: str = "score"):
             return [random.choice(prs)]
 
         elif selector_type == "top":
-            if metric not in prs[0].metadata:
+            metadata = prs[0].metadata
+            if metric not in metadata and "eval_scores" in metadata:
                 # check if the metric is in eval_scores
-                if "eval_scores" in prs[0].metadata and metric in prs[0].metadata["eval_scores"]:
-                    policy_scores = {p: p.metadata["eval_scores"].get(metric, None) for p in prs}
+                if metadata["eval_scores"] is not None and metric in metadata["eval_scores"]:
```
sasmith, line 120: Consider using `metadata.get("eval_scores", {}).get(metric)` here as well, since you're using it below. I don't think it's necessarily better, but it's helpful to do things in a consistent way.

sasmith: Consider for later:

I've been staring at this for a while, since it feels like it could be more straightforward.

I think it could be useful to write out in English what we expect to do about getting the metric. E.g., it's not clear to me why `metric` would be in one place or another, and when we should expect it to be arbitrarily missing. I could also see a scenario where the metric is _normally_ in `eval_scores.metric`, but happens to be missing from the latest run. Then we'll try to read it from `metric` for all runs, and find it completely missing, instead of trying to read it from `eval_score.metric` for all runs and finding it most of the time.

Is it plausible to change where we trying to read `metric` from, so we're either always reading from `metric` or `eval_scores.metric`?

sasmith (approved):
My comments are all "considers", and are very reasonable to address as follow ups, and shouldn't block this PR.

metta/agent/policy_store.py
```diff
@@ -114,10 +114,13 @@ def _policy_records(self, uri, selector_type="top", n=1, metric: str = "score"):
             return [random.choice(prs)]
 
         elif selector_type == "top":
-            if metric not in prs[0].metadata:
+            metadata = prs[
+                0
+            ].metadata  # check the latest metadata, to see whether this policy has the metric in metadata
```
sasmith, line 119: I think that comment is more tactical than would be helpful. It's describing what's happening, but not why.

I think the comment I'm looking for would look like
"""
Most metrics are stored directly at the metadata level, but metrics that come from evaluations are stored under the "eval_scores" key.

At this level, we don't know which of these places will have the metrics, so we check both. We do this using the most recent policy as an example.

If the metric name appears in both places, or in neither, we'll check the base-level location for the rest of the policies.
"""

But [to repeat myself] I think it would be better if we had some unambiguous way to indicate where the metric should be, and then the comment could be shorter.

PR #335, opened 2025-05-03
feat: Add Transformer layer as LSTM alternative

gabe00122:
Introduces a Transformer core layer designed as a drop-in substitute for the existing LSTM layer.

This implementation assumes that the backpropagation through time (BPTT) window matches the episode length. This requires careful batch length configuration.

Example Configuration:
---------------------
forward_pass_minibatch_target_size = num_agents * num_workers batch_size = forward_pass_minibatch_target_size * context_size minibatch_size = context_size # Or any multiple
bptt_horizon = context_size
game.max_steps = context_size - 1

vectorization = serial
---------------------

To utilize the Transformer, episode timesteps are passed as input to the policy. The expected shapes are:
- Inference: (batch, 1)
- Training: (batch, context_size)

Positional indices must align correctly with the input dimensions for proper functioning.

Note: Causal masking during training currently assumes positions increase monotonically from left to right.

### DIFF
--- a/metta/sim/simulation.py
+++ b/metta/sim/simulation.py
@@ -167,12 +167,16 @@ class Simulation:
         logger.info(f"Simulation settings: {self._config}")
 
         obs, _ = self._vecenv.reset()
+        time_steps = torch.zeros((self._vecenv.num_agents, 1), dtype=torch.long, device=self._device)
         policy_state = PolicyState()
         npc_state = PolicyState()
 
         game_stats = []
         start = time.time()
 
+        self._policy_pr.policy().eval()
+        if self._npc_pr is not None:
+            self._npc_pr.policy().eval()
         # Track environment completion status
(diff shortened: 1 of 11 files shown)


metta/sim/simulation.py
```diff
@@ -171,12 +171,16 @@ def simulate(self, epoch: int = 0, dry_run: bool = False):
...
 
         obs, _ = self._vecenv.reset()
+        time_steps = torch.zeros((self._vecenv.num_agents, 1), dtype=torch.long, device=self._device)
         policy_state = PolicyState()
         npc_state = PolicyState()
 
         game_stats = []
         start = time.time()
 
+        self._policy_pr.policy().eval()
```
daveey, line 181: what does this do?

daveey (approved):
could you run normal training (defaults) and make sure that things still train? otherwise, good to go

PR #383, opened 2025-05-05
feat: add useful benchmarks

rjwalters:
This PR adds:

* unit tests and performance benchmarks for the `sample_logits` function in `agent/util/distribution_utils.py`. 
* benchmarks for the CI python training and replay smoke tests


tests/agent/util/test_distribution_utils.py
```diff
@@ -0,0 +1,190 @@
...
+    small_vocab_size = 10
+    small_batch = torch.randn(small_batch_size, small_vocab_size)
+
+    # Medium batch
+    medium_batch_size = 32
+    medium_vocab_size = 50
+    medium_batch = torch.randn(medium_batch_size, medium_vocab_size)
+
+    # Large batch
+    large_batch_size = 128
```
Al-does, line 40: Heads up that the batch size under training is much bigger than during rollout. The batch size will grow by the batch size used during rollout _times_ the `bptt_horizon` value (which can be found in trainer/puffer.yaml). Right now, it's set to `64` so batch size grows tremendously.

tests/agent/util/test_distribution_utils.py
```diff
@@ -0,0 +1,190 @@
+import pytest
+import torch
+
+from metta.agent.util.distribution_utils import sample_logits
+
+
+# Test fixtures
+@pytest.fixture
+def sample_logits_data():
```
Al-does, line 9: Am I correct in reading these as single-discrete logits? If so, great! Our MettaAgents are in a single discrete space so would be good to test the hell out of that case. However, would also be cool to test the pufferlib version of `sample_logits()` as a comparison to see how we're doing but certainly just a nice-to-have.

Al-does (approved):
Nice!

tests/agent/util/test_distribution_utils.py
rjwalters: should we make these a lot bigger in the test?

tests/agent/util/test_distribution_utils.py
rjwalters: yes, the tests assume single discrete logits. we can compare to pufferlib with the caveat that they also handle List[Tensor] as an input - so it might be more work to get to an apples to apples comparison test. I was planning to treat your code as our starting point when looking for additional performance improvement

PR #393, opened 2025-05-06
MultiLeftOrRight scene; order_by_seed feature

berekuk:
As requested by Lars.

I had to copy-paste left_or_right yaml config to Python. The better solution would be to move it to a scene and reuse, but scenes root is currently in mettagrid, so can't do that right now.

This PR creates a precedent of `metta.map.scenes.*` files; I'd argue that we should move all mapgen code to `metta.map.*` and leave mettagrid minimal/agnostic to map builder implementations.

### DIFF
--- a/configs/env/mettagrid/maps/diversity/multi_left_or_right.yaml
+++ b/configs/env/mettagrid/maps/diversity/multi_left_or_right.yaml
@@ -1,0 +1,22 @@
+# Multi-room version of `left_or_right.yaml`.
+
+defaults:
+  - /env/mettagrid/mettagrid@
+  - _self_
+
+game:
+  num_agents: 32
+  # Enough time to pick one direction, but not to dilly-dally.
+  max_steps: 25
+
+  map_builder:
+    _target_: mettagrid.map.mapgen.MapGen
+
+    width: 200
+    height: 100
+    border_width: 1
+
+    root:
+      _target_: metta.map.scenes.multi_left_and_right.MultiLeftAndRight
+      rows: 4
+      columns: 8
--- a/metta/map/scenes/multi_left_and_right.py
+++ b/metta/map/scenes/multi_left_and_right.py
@@ -1,0 +1,85 @@
+from numpy import random
+
+from mettagrid.map.scene import Scene
+from mettagrid.map.scenes.random import Random
+from mettagrid.map.scenes.room_grid import RoomGrid
+
+
+class MultiLeftAndRight(Scene):
+    """
+    Produce multiple left-or-right maps in a grid, with agents assigned randomly
+    to teams, and rooms all identical otherwise.
+    """
+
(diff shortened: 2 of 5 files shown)


berekuk:
* **#393** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/393?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/393?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sasmith (approved):
How much do you think of this as a "one off to unblock Lars" vs the way things should be done?

I see this as more of the former, in which can I think we should have a header on the file giving that context. I think there are a couple of "right" versions, and I haven't thought through them enough to be sure how they pan out. E.g., the pattern of "copy this template 100 times, and have X be the same for everything, but have Y be different" feels like something we could probably support at the yaml level; and if we can't, we can clearly support it at the Python level. But having Python that has a baked-in yaml config inside it smell pretty off from where we want to be.

Unblocking since I think this is okay to check in, but want to fast-follow on either labeling this more clearly as not-a-clean-example, or figuring out what the clean version of this would be.

berekuk:
That's a pretty sizable discussion, and I don't have all the answers yet.

The main reason I did this in Python is not even because of agent groups, but because of seeds. I needed to make a random map, but keep the seed identical across sub-scenes. When I considered how a generic API for this would look in yaml, it was clear that these seeds are like equality edges between nodes in the tree of scene definition, and it's not clear how to do this in yaml at all. (I guess I could apply the same trick I used for locks, use "placeholder values" that would be replaced with random numbers? but I didn't think of this last night).

The bigger topic is that recently I've started to think that yaml/hydra/omegaconf are worse than configuration-as-code. I don't expect this to be very convincing yet, and haven't thought through all consequences of it myself (and this PR is not an intentional move in that direction).

> "copy this template 100 times, and have X be the same for everything, but have Y be different" feels like something we could probably support at the yaml level

The general form of this is something like... definitions that are not "dicts" but "functions that take parameters and produce dicts". We could invent a DSL in yaml that would do that somehow, maybe with resolvers that are context-aware (I don't know how we could pass context though, global variables?). Maybe there's some easier solution that I'm not seeing.

This observation that we want configs to be functions that take parameters is the reason why I moved to configuration-as-code camp, personally. I don't know how to express that in yaml clearly. (The thing that we actually want feels quite similar to how JSX and React components work, btw - a tree of components that are functions, take props and return trees with those parameters; then sometimes we also want access to global context, which coincidentally also exist in React).

---

One more thing: after the scenes move to the main metta, this could be cleaned up at least partially. Instead of copy-pasting the original left-and-right scene, I could load it with OmegaConf, deepcopy, inject seeds and monkey-patch two copies to belong to two teams.

PR #398, opened 2025-05-06
MultiLeftOrRight scene

berekuk:
MultiLeftOrRight scene

move mapgen code and tests to metta.map

tools.map.gen supports (some) env configs in place of map configs


berekuk:
broken PR, graphite misunderstood me

PR #405, opened 2025-05-06
feat: optimize distribution utils

rjwalters:
This PR attempts to optimize the `sample_logits` function. Part of the goal here is to demonstrate that our CI benchmarks catch any performance changes.

## Changes
- Improved type annotations with Optional and Tuple imports
- Added torch.jit.script decorator for better performance
- Updated docstring with more precise input/output specifications
- Removed unnecessary reshaping operations

We get about a 20% improvement on smaller batches:
image


metta/agent/util/distribution_utils.py
```diff
@@ -1,83 +1,67 @@
...
-    # Shape: [B*T]
-    logits_entropy = entropy(log_sftmx_logits)
+    # Step 4: Compute entropy directly without a loop
+    # Entropy formula: -sum(p * log(p))
+    joint_entropy = -torch.sum(softmaxed_logits * normalized_logits, dim=-1)
 
-    # Return shapes: [B*T], [B*T], [B*T], [B*T, A]
-    return action, act_logprob, logits_entropy, log_sftmx_logits
+    # Return action with shape [batch_size] to match expected output
+    return output_action.squeeze(-1), joint_logprob, joint_entropy, normalized_logits
```
Al-does, line 67: I thought we were trying to get away from squeeze's ie for when batch size is 1 it would squeeze out the batch dim.

Al-does:
Thanks Robb. Hopefully this adds stability.

Al-does (approved):

PR #411, opened 2025-05-06
Add group reward to infos and sim stats

daphnedemekas:
Here's code that gets the reward-per-group and adds it to infos to be used by eval stats. 

No need to approve, feel free to just use this in whatever way makes sense


mettagrid/mettagrid/mettagrid_env.py
```diff
@@ -48,6 +49,20 @@ def _reset_env(self):
             f"Number of agents {self._env_cfg.game.num_agents} does not match number of agents in map {map_agents}"
         )
 
+        # Count number of agents per team
```
jacklionheart, line 52: we should pick 1 of "team" or "group"

mettagrid/mettagrid/mettagrid_env.py
```diff
@@ -48,6 +49,20 @@ def _reset_env(self):
...
+        team_counts = {t: None for t in self._team_names}
```
jacklionheart, line 53: as i mentioned live, it seems to me that having team --> agent_ids map (which could then be `len()`ed to get counts) would be nice

jacklionheart (approved):

mettagrid/mettagrid/mettagrid_env.py
daphnedemekas: Ok cool, I have updated to have this. In terms of following through to be able to use those metrics on the analysis side, I'm feeling like I should wait for your changes? Lmk

PR #415, opened 2025-05-07
fix: debug python CI

rjwalters:
We need to include the files we built in deps in the cache so that e.g. pufferlib is available for CI.

### DIFF
--- a/.github/workflows/py-checks.yml
+++ b/.github/workflows/py-checks.yml
@@ -337,18 +337,7 @@ jobs:
       - name: Check deps
         run: |
           source venv/bin/activate
-          echo "content of repo"
-          ls -al
-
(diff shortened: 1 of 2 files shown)


.github/workflows/py-checks.yml
```diff
@@ -86,12 +86,12 @@ jobs:
       - name: Set cache paths
         id: set-cache-paths
         run: |
-          PATHS="mettagrid/build
+          PATHS="venv
           mettagrid/**/*.so
           metta/**/*.so
-          /build
```
berekuk, line 92: maybe `/build` is necessary too? metta got some cython code now (`metta/rl/fast_gae/fast_gae.pyx`)

PR #416, opened 2025-05-07
agent readme outlining revisioning process

Al-does:
# Add guidelines for MettaAgent versioning and backward compatibility

### TL;DR

Added documentation for maintaining backward compatibility when evolving MettaAgent components.

### What changed?

- Created a new README.md in the metta/agent directory
- Established a versioning system for layer classes using the format `ClassNameRevMM-DD-XX`
- Defined a deprecation process for superseded classes
- Provided guidance on config file management
- Added notes on future considerations for model saving approaches

### Why make this change?

To ensure that older trained agents can still run inference and training with the exact class implementations they were trained with, even as the codebase evolves with newer versions of those classes. This maintains backward compatibility while allowing for continuous improvement of the agent architecture.


Al-does:
* **#416** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/416?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/416?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

daveey:
i dont love this approach, lemme think about it some more before you spend too much time implimenting

sasmith (commented):
Same thought as Dave -- let's chat in person?

Some considerations, in no clear order:
* I'd want to base this plan on breakages that we're actually causing, and expect to cause. E.g., if we expect to break things rarely post-observation changes, then maybe we can punt on this. Maybe we have a list of examples of issues we're trying to address, though?
* There are lots of non-model ways to have non-backwards compatibility. E.g., environmental changes. We should make sure our solution works for as wide a class as it reasonably can, since solving only one part of these issues will add cost without adding a lot of benefit.
* A lot of versioning can be done via git. I don't think this solve all issues, as we presumably want to be able to use old models on the new viewer (e.g.), but maybe there's something here.
* I don't think we want to guarantee people can run against exactly the same code -- just equivalent code. So, e.g., we should be able to make minor, backwards compatible changes without needing to make new layer classes or anything.
  * We should have appropriate testing to be able to pretty-much guarantee backwards compatibility. Sometimes we'll ship bugs.

daveey:
I hope we can come up with something better, but one solution i had is version packaging

metta.agent.lib.obs_shaper.v1.ObsShaper
can then be replaced by
metta.agent.lib.obs_shaper.v2.ObsShaper

daveey (changes requested):
See if my suggestions make sense, and discuss with @relh

relh:
My thoughts:

- use @daveey's version packaging suggestion for versions of observation/action space.
- include this version in metadata for the MettaAgent.
- add thunking if loading a model with an old version.
- we don't version layers because weights_only=False and torch.jit.script don't need to find the class to load and run.
- we CAN support copying old weights out but we can just set strict=False and allow it to be partial/break if layers have changed
- we just accept that because we can still do kickstarting to guide new models.

Al-does:
I think we're all aligned here and Richard has a great plan that we discussed on #policy-store in Discord so let's go forward with that. I'll close this out.

relh:
https://github.com/Metta-AI/metta/compare/richard-versioning-initial?expand=1

Here is an initial version. I think we can keep things very simple. Basically, we add the observation and action space information to the metadata and then if we're loading a policy we check if it matches the metadata. 

Next, if it doesn't we can add thunking layers / adapters to go from v0 to v1.

But when creating an agent we just create it at the current version of the code. So we only care about loading.

PR #420, opened 2025-05-07
Make feature offsets deterministic

Barraketh:


rjwalters (approved):

rjwalters:
You're observing different behavior between Linux and Mac platforms:

On Linux: The features.size() is incremented first
On Mac: The indexing happens first

This is a classic case of undefined behavior in C++. When you have an expression like:

map[key] = map.size();


The C++ standard doesn't guarantee the order of evaluation between:

Evaluating map.size()
Actually assigning the value to map[key]

Why this happens
The issue occurs because features[key] will create a new entry if the key doesn't exist. So if you're using the current size as an index value, the behavior depends on whether:

features.size() is evaluated before the new entry is created (giving the "expected" value)
features.size() is evaluated after the new entry is created (giving a value 1 higher than expected)

How to fix it
Here's a corrected version that ensures deterministic behavior across platforms:

// Instead of:
// features[_type_feature_names[type_id][i]] = features.size();

// Do this:
size_t current_size = features.size();
features[_type_feature_names[type_id][i]] = current_size;

sasmith:
Ugh. I definitely see what's happening here, and it also definitely violates my mental model.

Barraketh:
Yeah, my conclusion from this was that we should almost never use the [] operator on maps - it's totally unintuitive

PR #422, opened 2025-05-08
add LP directly into mettagrid

aadharna:
There are two classes in here the Bidirectional LP class and an Env Wrapper class. The LP class is an `online` variant that doesn't require large evaluation phases between each training step to update the LP distribution.  The LP class directly computes and maintains a learning progress calculation over x buckets (split evenly into a discrete space). 

* The LP expects data for each bucket/key to be sent in through the `collect_data(self, infos)` function that takes in an infos dict where there is a corresponding dict-key of the form `tasks/{TASK_ID}: [values...]`. The LP expects that incoming data will be from the levels/tasks it has sampled using the LP distribution. If the task is not from the expected levels, we'll still log it, but it won't iterate on the counters that determine when to refresh the LP distribution. 
* The LP (after you've sent in enough data) will allow you to calculate a new sampling distribution and provide some sampled levels from that distribution. There's also an exploration question embedded here (because the algorithm is online, so I also added some random exploration of the task space so that we don't just sample the same level/task forever. 
* There's also an env wrapper that shows a potential example usecase at the bottom of the file that would overload the reset function as defined in e.g., `mettagrid_env.py`. The wrapper interacts with the LP when we've reached the end of an episode or when you call the `notify` function that triggers an update call to the LP calculations. 

The other way to think about this API is the form of almost an evolutionary algorithm's `ask` and `tell` API. The `notify`/`calculate_dist` functions `ask` you to sample from given tasks and then you `tell` me the answers (one by one, here) until we've collected enough data to trigger a recompute / new `ask` call. 

----

I've been using this inside of Pufferlib (which is currently going through a large refactor), and so the forms here are all malleable. 

I haven't tested it yet with the learning code directly in `metta` in part because of the large refactor going on on both sides of the project, but I did update my fork of metta here to what's in master as of 7:30pm on May 7th for the PR. 

The mettagrid version I've been working off of was the following commit: 
```
commit 6aa7f863f397d2b0c5baf6b6eb2ec45fb5dfd43c (HEAD -> main, origin/main, origin/HEAD)
Author: georgejwdeane <101122217+georgejwdeane@users.noreply.github.com>
Date:   Mon Apr 14 15:13:37 2025 -0700

George navigation training (#76)
...
```

As such, I don't expect tests to necessarily pass yet

----

Learning curves looking at agent/heart.*

image

The bottom curve is the default behaviour when integrated with Pufferlib using the default metta domain randomization. The higher curve is with the LP integrated into pufferlib. 

You can find my yaml file for the sim [here](https://github.com/aadharna/PufferLib/blob/metta_lp/pufferlib/environments/metta/dr_metta.yaml). I took the yaml already in pufferlib and added back in some of the sampling params and tied together a few of the cooldowns.

The raw data is [here](https://app.neptune.ai/o/aadharna/org/metta-LearningProgress/runs/compare?viewId=standard-view&dash=charts&sortBy=%5B%22sys%2Fcreation_time%22%5D&sortFieldType=%5B%22datetime%22%5D&sortFieldAggregationMode=%5B%22auto%22%5D&sortDirection=%5B%22descending%22%5D&groupBy=%5B%22sys%2Fgroup_tags%22%5D&groupByFieldType=%5B%22stringSet%22%5D&groupByFieldAggregationMode=%5B%22auto%22%5D&suggestionsEnabled=false&lbViewUnpacked=true&compareChartsFilter=environment%2Fagent%2Fheart&compare=KwZgNOCMbbNgJnrIA) and be sure to average grouped runs. 

---
  - https://app.asana.com/0/0/1209208922660407


aadharna:
Not sure why it autoassigned it to sasmith, but it won't let me also assign it to David.

PR #432, opened 2025-05-09
fix: improve obs_shaper error msgs

rjwalters:
These changes make debugging easier by providing developers with precise information about what went wrong when a shape mismatch occurs. Instead of just seeing "Invalid input tensor shape", developers will now see exactly what shape was expected and what they provided.

This PR enhances the error messages in the `ObsShaper` class to provide more detailed and helpful information when 
- Replaced generic error messages with detailed diagnostics that clearly show:
  - The actual input tensor shape
  - The expected shape format
  - What specifically is wrong with the input
  
- Added better formatting to make error messages more readable
- Split error handling into two distinct cases:
  - Shape dimension mismatch (improved error message for when dimensions don't match)
  - Tensor dimensionality errors (improved error message when tensor has wrong number of dimensions)


metta/agent/lib/obs_shaper.py
```diff
@@ -5,6 +5,7 @@
 
 class ObsShaper(LayerBase):
     def __init__(self, obs_shape, num_objects, **cfg):
+        # Note that __init__ is NOT called when we reload a policy from a checkpoint file
```
Al-does, line 8: This comment could be on every single layer class as well as the MettaAgent. Maybe "Note that the __init__ of any layer class and the MettaAgent are only called when the agent is instantiated and never again. I.e., not when it is reloaded from a saved policy."

I'm also realizing that this could use a docstring. Maybe "This class does the following: 1) permutes input observations from [B, H, W, C] or [B, TT, H, W, C] to [..., C, H, W] 2) inspects tensor shapes, ensuring that input observations match expectations from the environment and 3) inserts batch size, TT, and B * TT into the tensor dict for certain other layers in the network to use if they need reshaping.

Al-does (approved):
Looks great to me. Thank you for doing this.

metta/agent/lib/obs_shaper.py
rjwalters: ok I added similar documentation on the other layers -- this is mostly llm so it could use a thorough review

PR #435, opened 2025-05-09
fix: update rl/pufferlib

rjwalters:
This PR updates the code in `metta/rl/pufferlib` to improve code quality, fix bugs, and add validation checks.

### In `policy.py`:
- Improved the formatting of `max_vec` tensor initialization for better readability
- Added debugging info to print weight shapes during policy loading
- Fixed the type annotations

### In `trainer.py`:
- Added proper import for `MettaAgent` class
- Improved policy loading logic with better error messages
- Added validation checks to ensure policy observation shape matches environment shape
- Fixed references to use `metta_grid_env` consistently throughout the code
- Fixed the GAE import to use the function directly instead of the module
- Added a null check for `self.losses` before accessing it

### In `trainer_checkpoint.py`:
- Improved type hints with proper Optional annotations
- Added return type annotations
- Fixed logger name to be more specific


metta/rl/pufferlib/policy.py
```diff
@@ -109,10 +78,19 @@ def load_policy(path: str, device: str = "cpu"):
     num_actions, hidden_size = weights["policy.actor.0.weight"].shape
     num_action_args, _ = weights["policy.actor.1.weight"].shape
     cnn_channels, obs_channels, _, _ = weights["policy.network.0.weight"].shape
+
+    # TODO -- fix magic number 11
```
rjwalters, line 82: I have tried to catch this case in the debugger so that i can figure out where we can read the obs_shape but I haven't been able to actually trigger a breakpoint... @Al-does do you know how to make this code run?

metta/rl/pufferlib/trainer.py
```diff
@@ -79,46 +80,45 @@ def __init__(
         self._weights_helper = WeightsMetricsHelper(cfg)
         self._make_vecenv()
 
+        self.metta_grid_env: MettaGridEnv = self.vecenv.driver_env  # type: ignore
```
daveey, line 83: i prefer we don't hold on to this in the class. make it local for the init() and then save self._agents_per_env rather than holding onto the driver_env

daveey (approved):
Looks great, other than the trainer.metta_grid_env. Feel free to merge after fixing

metta/rl/pufferlib/trainer.py
rjwalters: this is essentially a type hint because we are assuming that mettagridenv fields are available later in the file.. if we want to only access PufferEnv fields from driver_env (the parent class) then we should not typecast here and instead we should figure out why pylance is complaining

PR #437, opened 2025-05-09
Refactor simulation config to support single environment simulations

daveey:
# Refactor Simulation Config to Support Single Environment Simulations

This PR refactors the simulation configuration system to better support single environment simulations. It introduces a new `SingleEnvSimulationConfig` class that inherits from `SimulationConfig` and simplifies the replay job workflow.

Key changes:
- Added `SingleEnvSimulationConfig` class for simpler single-environment configurations
- Created `sim_single.yaml` and `sim_suite.yaml` to separate single and suite configurations
- Removed the "Play Metta" launch configuration from VSCode
- Updated simulation classes to accept device and vectorization parameters explicitly
- Simplified replay generation in the trainer
- Fixed terrain loading in `terrain_from_numpy.py` to improve logging
- Updated tools like `replay.py` and `sim.py` to work with the new configuration structure
- Removed unnecessary configuration parameters from various YAML files

This change makes it easier to configure and run simulations with a single environment, reducing boilerplate and improving clarity in the configuration system.


jacklionheart (approved):
Godspeed.

Barraketh (approved):

Barraketh:
Approved, but one question: why remove the 'dry_run' configuration?

configs/sim_job.yaml
```diff
@@ -1,20 +1,14 @@
 defaults:
   - common
   - wandb: metta_research
-  - sim: simple
+  - /sim/sim_suite@replay_job.simulation_suite
```
sasmith, line 4: I think this should be `/sim/sim_suite@sim_job.simulation_suite`.

If that's right, can you fix it and try to write a test that would have could it?

This PR also broke run_evals.sh and similar scripts. They were previously using `sim=navigation`. I can get this to work by changing this to `/sim/navigation@sim_job.simulation_suite`, but I can't figure out how to get that to work on the command line.

PR #401, opened 2025-05-06
Add StatsDb to replace EvalStatsDb

jacklionheart:
This change introduces `StatsDb`, a new datastore for simulation stats.

Our previous implementation was an aggregation of json files across all versions of a wandb artifact. where each file was a separate simulation. This has several limitations, including:  

* It is difficult to delete data (because data across all versions is always included)
* There is no schema to see what kind of data is there or how to write queries on top of it
* There is no way to launch a CLI tool to manually inspect and query the data
* It is difficult to re-use this approach in the training context where we need to track each many orders of magnitude of “simulations” and associate them with different policy versions
* It is circuitous to materialize views on top of this data because it is separated by file rather than schema.  

This new implementation addresses all of the issues through the following approach: 
* Data is now written using SQL rather than JSON with clear schema
* Each `MettaGridEnv` gets its own `StatsWriter` to a unique file on disk using duckdb.
* `Simulation` now merges each of those into a unified duckdb.
* `StatsDB.export` now explicitly merges data from the current run with any data already stored at the export URI (rather than relying consumer to merge versions).
* There is now a script `stats_duckdb_cli` that launches a CLI to interact with (a local copy of) a database   

This change deploys this only for Evaluation stats, but a subsequent diff will add a similar setup for the training data.

PR #439, opened 2025-05-09
Add StatsDb to replace EvalStatsDb (#401)

jacklionheart:
Add StatsDb to replace EvalStatsDb (#401)
This change introduces `StatsDb`, a new datastore for simulation stats.

Our previous implementation was an aggregation of json files across all
versions of a wandb artifact. where each file was a separate simulation.
This has several limitations, including:  

* It is difficult to delete data (because data across all versions is
always included)
* There is no schema to see what kind of data is there or how to write
queries on top of it
* There is no way to launch a CLI tool to manually inspect and query the
data
* It is difficult to re-use this approach in the training context where
we need to track each many orders of magnitude of “simulations” and
associate them with different policy versions
* It is circuitous to materialize views on top of this data because it
is separated by file rather than schema.  

This new implementation addresses all of the issues through the
following approach: 
* Data is now written using SQL rather than JSON with clear schema
* Each `MettaGridEnv` gets its own `StatsWriter` to a unique file on
disk using duckdb.
* `Simulation` now merges each of those into a unified duckdb.
* `StatsDB.export` now explicitly merges data from the current run with
any data already stored at the export URI (rather than relying consumer
to merge versions).
* There is now a script `stats_duckdb_cli` that launches a CLI to
interact with (a local copy of) a database   

This change deploys this only for Evaluation stats, but a subsequent
diff will add a similar setup for the training data.


jacklionheart:
* **#439** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/439?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/439?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Barraketh (approved):

PR #443, opened 2025-05-09
experimental code for a fully robust agent - obs and actions

Al-does:
### TL;DR

Added a new robust observation processing architecture for MettaAgent for testing purposes. This is experimental code and will change significantly going forward. It is fully backwards compatible with the existing agents.

### What changed?

- Added a new agent configuration `configs/agent/robust_obs.yaml` that implements a more sophisticated observation processing pipeline
- Created new observation encoding modules in `metta/agent/lib/obs_encoding.py`:
  - `ObsEmbedder`: Processes grid observations by embedding object types and attributes. This layer runs pretty slowly but I think I've implemented all of the algorithmic low hanging fruit to speed things up as it needs to process roughly 12,000 data points (the unique points in obs space) per batch. 
     It uses an ego-centric coordinate mapping.
     It scales embeddings (same direction, different magnitude) based on the distance from the agent and by the value in the in the pixel itself. I think this has potential but could be improved: 1) I'd like to see if leaving the scaling in the coords part of the feat vector is enough and 2) need to address the problem that for certain attributes like cooldown, smaller is 'better'. As an alternate, we could just have the pixel value as another vector in the coord but I'd prefer for the relationship to be multiplicative (scaling bilinearly) vs linear (additive). 
  - `ObsAttn`: Implements an attention mechanism over the embedded observations. I think this architecture could be significantly better and would like to experiment in the future. One approach would be using an attention context window to the LSTM.

### DIFF
--- a/configs/hardware/macbook.yaml
+++ b/configs/hardware/macbook.yaml
@@ -1,6 +1,6 @@
 # @package __global__
 
-device: cpu
+device: mps
 vectorization: serial
 
 trainer:
--- a/configs/hardware/macbook.yaml
+++ b/configs/hardware/macbook.yaml
@@ -1,6 +1,6 @@
 # @package __global__
 
-device: mps
+device: cpu
 vectorization: serial
 
 trainer:
--- a/configs/agent/robust_obs.yaml
+++ b/configs/agent/robust_obs.yaml
@@ -1,0 +1,92 @@
+# see reference_design.yaml for explanation of components
+_target_: metta.agent.metta_agent.MettaAgent
+
+observations:
+  obs_key: grid_obs
+
+clip_range: 0 # set to 0 to disable clipping
+analyze_weights_interval: 300
+l2_init_weight_update_interval: 0
+
+components:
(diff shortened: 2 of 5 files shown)


Al-does:
* **#443** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/443?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/443?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

daveey (approved):
Not reviewing in any detail since you plan on rewriting / cleaning it up.

configs/hardware/macbook.yaml
```diff
@@ -1,6 +1,6 @@
 # @package __global__
 
-device: cpu
```
daveey, line 3: dont check this in, it breaks people

Al-does:
Rewriting in another branch with Alex S' specific buffer structure in mind.

PR #459, opened 2025-05-12
feat: add debug utils for saving c data

rjwalters:
## Debug Utilities for C/C++ Testing

This PR adds utilities for saving Python objects and step function results in formats that are easily consumable by C/C++ tests. The primary goal is to facilitate testing of the C++ implementation against the Python reference implementation.

### What's included:

- New `debug.py` utility module with functions to serialize complex Python objects for C/C++ testing
- `save_mettagrid_args` function to save environment configuration and map data
- `save_step_results` function to save the complete output from the step function
- Added flags and hooks in `MettaGridEnv` to enable data serialization during runtime
- Safety mechanisms to prevent excessive file creation (max 10 saves per session)

### Usage:

To save constructor arguments during environment initialization:
```python
env = MettaGridEnv(env_cfg, env_map)
env._save_mettagrid_args = True
```

To save step results during environment execution:
```python
env = MettaGridEnv(env_cfg, env_map)
env._save_step_results = True
```

The saved files can be used as test fixtures in C/C++ tests to verify the C++ implementation behaves identically to Python.

### Implementation details:

- Handles complex, nested Python objects including NumPy arrays, dictionaries, and custom classes
- Filters out unnecessary metadata to keep test fixtures clean
- Prevents circular references and infinite recursion with object tracking
- Multiple output formats (pickle, JSON, TXT) for different use cases
- Customizable output directory and file naming


mettagrid/mettagrid/util/debug.py
```diff
@@ -0,0 +1,503 @@
+# metta/util/debug.py
```
sasmith, line 1: Am I seeing comments you're leaving to LLMs? I don't think we need to put file names at the top of files.

mettagrid/mettagrid/util/debug.py
sasmith, line 1: Can you add a comment / example of usage here? I haven't read all the logic carefully, and for debugging tools, I don't feel a strong need to (at least while this is "scaffolding"), but I think documentation on basic usage is necessary for folks to be able to use this.

I imagine this would be "suppose you want to debug xyz. Here's how to use this tool to aid with that: ..."

sasmith (approved):

PR #461, opened 2025-05-12
feat: Add Parallel Mac Hardware Configuration

rjwalters:
This PR splits the existing `macbook.yaml` configuration into two separate files:
- `mac_serial.yaml`: Contains the original serial vectorization configuration
- `mac_parallel.yaml`: Adds a new parallel configuration using multiprocessing

For backward compatibility, `macbook.yaml` now includes `mac_serial.yaml` via Hydra's defaults.

## Changes
- Added `mac_serial.yaml` with the original macbook configuration (serial vectorization)
- Added `mac_parallel.yaml` with multiprocessing vectorization and optimized settings:
  - Changed vectorization from serial to multiprocessing
  - Increased num_workers from 1 to 4
  - Increased forward_pass_minibatch_target_size from 2 to 512
- Modified `macbook.yaml` to include `mac_serial.yaml` for backward compatibility

## Motivation
This change allows researchers to easily switch between serial and parallel processing modes on Mac hardware without modifying their existing code. The parallel configuration provides performance improvements for multi-core MacBooks.

## Testing
Tested both configurations on MacBook Pro with M2 chip, confirming:
- Existing code using `macbook.yaml` continues to work without changes
- Parallel configuration shows improved training throughput on supported hardware


sasmith (approved):
I haven't tried multiprocessing yet. I think this is an advantage at the simulation layer, versus using more GPU power. Is that right?

Either way, can you add a comment to the configs about the difference?

PR #466, opened 2025-05-13
Remove deps directory and use git dependencies instead

daveey:
# Remove Local Dependencies in Favor of Git Dependencies

This PR removes the local `deps` directory and replaces it with direct Git dependencies in `requirements.txt`. The changes include:

- Added direct Git dependencies to `requirements.txt` for PufferLib, wandb_carbs, and carbs
- Moved version constraints to a separate `constraints.txt` file
- Removed all code that managed the local `deps` directory in build scripts
- Updated Docker configuration to use simplified build process
- Removed deps-related paths from VSCode workspace configuration
- Simplified CI workflow by removing deps-related file hashing and caching
- Updated `pyproject.toml` to remove pufferlib from package directories

These changes simplify the development environment setup and dependency management by using standard Python package installation methods.


daveey:
* **#466** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/466?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/466?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

daveey (commented):
I realize that there is more cleanup to do, and i dont love requirements.txt duplicating constraints.txt. but i think that can be cleaned up later.

devops/setup_build.sh
```diff
@@ -54,7 +54,7 @@ bash "$SCRIPT_DIR/checkout_and_build.sh"
 
 # ========== INSTALL PACKAGES BEFORE BUILD ==========
 echo -e "\n\nInstalling main project requirements...\n\n"
-pip install -c requirements.txt -r requirements.txt
+pip install -c constraints.txt -r requirements.txt
```
rjwalters, line 57: why not just reuse the requirements file?  I think that may be better for dependabot and less prone to getting out of sync. particularly if you anticipate contraints being a duplicate file rather than a shorter list of locked down versions

daveey: metta2) ➜  metta git:(daveey-nodeps) pip install -c requirements.txt -r requirements.txt 

DEPRECATION: Constraints are only allowed to take the form of a package name and a version specifier. Other forms were originally permitted as an accident of the implementation, but were undocumented. The new implementation of the resolver no longer supports these forms. A possible replacement is replacing the constraint with a requirement. Discussion can be found at https://github.com/pypa/pip/issues/8210
ERROR: Unnamed requirements are not allowed as constraints

daveey: I suppose we could use requirements.txt and source_requirements.txt maybe that's better?

rjwalters: hmm.. TIL. I guess if pip is changing the rules then dependabot will probably support a constraints file. I think the problem I was trying to solve here was pufferlib thinking it needed older versions of libraries to run. We can probably just patch our forks and get around it that way.

Also maybe we switch to UV someday

daveey: yes, i patched pufferlib to update gymnasium/pettingzoo. and yes on uv, i want to do that next.

rjwalters (approved):

PR #472, opened 2025-05-13
fix: improve resolver test suite

rjwalters:
Add tests for edge cases in scale


mettagrid/mettagrid/resolvers.py
```diff
@@ -109,6 +109,17 @@ def oc_scale(
     Numeric
         The scaled value in the output range
     """
+
+    # Ensure in_min <= in_max
+    assert in_min <= in_max
+    assert out_min <= out_max
+
+    if in_min == in_max:
+        return (out_min + out_max) / 2
```
sasmith, line 118: This seems wrong, or at least dicey. Can we just assert `in_min < in_max`?

sasmith (approved):

mettagrid/mettagrid/resolvers.py
rjwalters: yeah, we can define it that way instead

PR #484, opened 2025-05-14
fix: action-to-logit index computation in MettaAgent

rjwalters:
This PR fixes a bug in `_convert_action_to_logit_index()` where action type IDs were erroneously being added twice when computing flat logit indices from `(action_type, action_param)` pairs. This caused out-of-bounds errors during inference when `sample_logits()` attempted to index logits.

#### Changes

* ✅ Rewrote `_convert_action_to_logit_index()` to compute `logit_index = offset + action_param` instead of `action_type + offset + param`
* ✅ Added a validation check in `forward()` to raise an informative error if `action_logit_index` is out of range
* 🧼 Removed unused variables and improved clarity in comments

#### Impact

* Fixes `RuntimeError: index N is out of bounds for dimension 1 with size A` in TorchScript when provided actions exceed the logits’ range
* Ensures model behavior is consistent across training, evaluation, and export
* Maintains TorchScript compatibility and shape flexibility

### DIFF
--- a/metta/agent/metta_agent.py
+++ b/metta/agent/metta_agent.py
@@ -217,19 +217,19 @@ class MettaAgent(nn.Module):
         return action, logprob_act, entropy, value, log_sftmx_logits
 
     def _convert_action_to_logit_index(self, action):
-        """Convert action pairs (action_type, action_param) to single discrete action logit indices using vectorized
-        operations"""
+        """
+        Convert (action_type, action_param) pairs to discrete action indices
+        using precomputed offsets.
+        Assumes `cum_action_max_params` maps action types to start indices.
+        """
         action = action.reshape(-1, 2)
 
-        # Extract action components
-        action_type_numbers = action[:, 0].long()
-        action_params = action[:, 1].long()
+        action_type = action[:, 0].long()
+        action_param = action[:, 1].long()
 
-        # Use precomputed cumulative sum with vectorized indexing
-        cumulative_sum = self.cum_action_max_params[action_type_numbers]
-
-        # Vectorized addition
-        action_logit_index = action_type_numbers + cumulative_sum + action_params
+        # Only use offset + parameter — NOT + action_type
+        offset = self.cum_action_max_params[action_type]
+        action_logit_index = offset + action_param
 
         return action_logit_index.reshape(-1, 1)
 


metta/agent/metta_agent.py
```diff
@@ -137,28 +137,63 @@ def _setup_components(self, component):
...
+        # Store parameters
         self.device = device
-        self.actions_max_params = action_max_params
+        self.action_names = action_names
+        self.action_max_params = action_max_params
+
         self.active_actions = list(zip(action_names, action_max_params, strict=False))
 
         # Precompute cumulative sums for faster conversion
-        self.cum_action_max_params = torch.cumsum(torch.tensor([0] + action_max_params, device=self.device), dim=0)
```
Al-does, line 147: this line offsets the action_max_params by 1 (with the torch.tensor[0]) at the beginning which is what allows the algorithm to work correctly. Also note that action_max_param starts counting at 0. So actions with only 1 param are a 0 and contribute nothing to this cumulative sum.

metta/agent/metta_agent.py
```diff
@@ -217,19 +252,19 @@ def forward(self, x, state: PolicyState, action=None):
...
-        # Use precomputed cumulative sum with vectorized indexing
-        cumulative_sum = self.cum_action_max_params[action_type_numbers]
+        action_type = action[:, 0].long()
+        action_param = action[:, 1].long()
 
-        # Vectorized addition
-        action_logit_index = action_type_numbers + cumulative_sum + action_params
+        # Only use offset + parameter — NOT + action_type
+        offset = self.cum_action_max_params[action_type]
+        action_logit_index = offset + action_param
```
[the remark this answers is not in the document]
Al-does: For this to work, you would have to modify some of the code above. In particular, you would have to take self.action_max_params and add a vector of 1s to it so it starts counting at 0. Then generate self.cum_action_max_params.

Al-does (changes requested):

metta/agent/metta_agent.py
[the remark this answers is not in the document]
rjwalters: after additional head scratching I think we both have concluded the original math was fine

metta/agent/metta_agent.py
```diff
@@ -231,7 +239,7 @@ def _convert_action_to_logit_index(self, action):
         # Vectorized addition
         action_logit_index = action_type_numbers + cumulative_sum + action_params
 
-        return action_logit_index.reshape(-1, 1)
+        return action_logit_index.view(-1)  # shape: [B] or [B*T]
```
rjwalters, line 242: @Al-does can you weigh in here? I think this needs to be matched to what sample_logits expects

Al-does: I'm sorry, Robb. Just saw this. Will take a look in the morning.

Al-does (approved):

PR #485, opened 2025-05-14
Fix observations at map edge

Barraketh:


sasmith (approved):
Overflow is still technically a risk, but I think we don't need to worry much about maps of size > 1B yet.

PR #505, opened 2025-05-15
Object use evaluation code

daphnedemekas:
Currently commenting out the group metric code as that is breaking some of the evals. 
Updated the object use evals so the rooms are smaller, and added in additional objects into the training environment.


georgejwdeane (approved):

metta/eval/dashboard/heatmap.py
sasmith, line 1: I think this change can be checked in on its own. There's not a whole lot of complexity here, but it's changing a bunch of white space along with removing colorscales; and moreover, it stands on its own.

sasmith (commented):
I think you're already splitting it up, so these comments seem somewhat redundant.

PR #506, opened 2025-05-15
feat: add MettaGridConfig class

rjwalters:
This PR moves parts of mettagrid around to make room for the `MettaGridConfig` class, which provides a cleaner way to handle the configuration data needed by MettaGrid environments. The class encapsulates both the environment configuration (OmegaConf config) and map data (numpy array) needed to initialize the MettaGrid C++ backend.

## Key Changes
- Added new `MettaGridConfig` class that:
  - Manages both config and map data in a unified interface
  - Handles map generation and validation
  - Provides proper type conversion for C++ compatibility
  - Offers a convenient `to_c_args()` method for MettaGrid initialization
- Updated `MettaGridEnv` to use the new config class
- Relocated room modules from `mettagrid.config.room.*` to `mettagrid.room.*` to improve organization
- Moved utility functions to dedicated namespace

## Benefits
- Cleaner separation of concerns and improved code organization
- Simplified environment initialization
- Better type safety with proper casting
- Easier validation of environment configuration
- Groundwork for upcoming curriculum learning implementation

## Future Work
This PR sets the foundation for curriculum learning support, which will be implemented in a subsequent PR.


configs/env/mettagrid/bases.yaml
```diff
@@ -6,7 +6,7 @@ game:
   num_agents: 24
 
   map_builder:
-    _target_: mettagrid.config.room.room_scene.RoomScene
+    _target_: mettagrid.room.room_scene.RoomScene
```
sasmith, line 9: I think your branch had more than intended, given that it's called "move-room".

rjwalters: yeah I need to move stuff out of the way. I am also fixing the makefile in this PR

mettagrid/mettagrid/config.py
```diff
@@ -0,0 +1,93 @@
...
+        env_map: Optional[np.ndarray] = None,
+    ):
+        """
+        Initialize a MettaGridConfig.
+
+        Args:
+            env_cfg: OmegaConf configuration for the environment
+            env_map: Optional pre-generated map. If None, a map will be generated using the map_builder from env_cfg
+        """
+        self._env_cfg = OmegaConf.create(copy.deepcopy(env_cfg))
```
sasmith, line 31: Probably worth a comment on why we're copying here. I think a reasonable answer is "we copy this pre-resolving since the config is re-used, and we'll want random resolvers to have different values when the re-resolve".

mettagrid/mettagrid/config.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @property
+    def env_cfg(self) -> DictConfig:
+        """Get the environment configuration."""
+        return self._env_cfg
+
+    @property
+    def env_map(self) -> Optional[np.ndarray]:
+        """Get the environment map if it exists."""
+        return self._env_map
```
sasmith, line 46: I kind of feel like the point of having a property is so we can auto-generate a map if needed. Otherwise, why not just expose `_env_map` directly?

mettagrid/mettagrid/config.py
```diff
@@ -0,0 +1,93 @@
...
+        map_agents = np.count_nonzero(np.char.startswith(env_map, "agent"))
+        assert self._env_cfg.game.num_agents == map_agents, (
+            f"Number of agents {self._env_cfg.game.num_agents} does not match number of agents in map {map_agents}"
+        )
+
+        self._env_map = env_map
+        return env_map
+
+    @property
+    def map_builder(self):
```
sasmith, line 73: Seems like this doesn't need to be exposed.

sasmith (approved):
I think this would be better if the UV changes were split from the config changes. I think this could actually be 3 changes, but ymmv. But, for example, I think the chance of us needing to revert the UV changes are much higher than the chance of needing to revert the config changes, so it seems better to decouple them.

PR #514, opened 2025-05-16
chore: add types to MettaGridEnv

rjwalters:
(This is #4 in a series of PRs that will incrementally cherry pick changes from https://github.com/Metta-AI/metta/pull/404. I will include a pufferbox run comparing to main for each PR.)

This PR adds proper type annotations to the MettaGridEnv class, improving code readability and IDE support through static type checking.


mettagrid/mettagrid/mettagrid_c.cpp
```diff
@@ -407,8 +409,8 @@ void MettaGrid::validate_buffers() {
 }
 
 void MettaGrid::set_buffers(py::array_t<unsigned char, py::array::c_style>& observations,
-                            py::array_t<bool, py::array::c_style>& terminals,
-                            py::array_t<bool, py::array::c_style>& truncations,
+                            py::array_t<unsigned char, py::array::c_style>& terminals,
```
daveey, line 412: @sasmith ran into issues here. the types don't match the passed in type, so new arrays are allocated rather than using the buffers

mettagrid/mettagrid/mettagrid_c.hpp
```diff
@@ -26,12 +26,15 @@ class MettaGrid {
...
+  unsigned short obs_width;
+  unsigned short obs_height;
+
   // Python API methods
   py::tuple reset();
   py::tuple step(py::array_t<int> actions);
   void set_buffers(py::array_t<unsigned char, py::array::c_style>& observations,
-                   py::array_t<bool, py::array::c_style>& terminals,
-                   py::array_t<bool, py::array::c_style>& truncations,
+                   py::array_t<unsigned char, py::array::c_style>& terminals,
```
sasmith, line 36: Why are you changing these from being bools?

mettagrid/mettagrid/mettagrid_c.cpp
```diff
@@ -177,19 +179,6 @@ MettaGrid::MettaGrid(py::dict env_cfg, py::array map) {
       }
     }
   }
-
-  // Initialize buffers. The buffers are likely to be re-set by the user anyways,
```
sasmith, line 181: Removing this goes beyond adding types, so even if it's correct, let's not have it in this PR.

As for correctness: I had the same instinct as you to remove it. @daveey pointed out that there are various ways that this gets when we aren't calling `set_buffer` externally. I can't immediately say what those are, so for now: caveat emptor.

mettagrid/mettagrid/mettagrid_env.py
```diff
@@ -36,15 +61,59 @@ def __init__(
...
+        buf_obj = SimpleNamespace(**buf)
+        super().__init__(buf_obj)
+
         self.labels = self._env_cfg.get("labels", None)
-        self.should_reset = False
+        self._should_reset = False
 
-        self._reset_env()
-        super().__init__(buf)
+        # check on the buffer shapes
```
sasmith, line 92: All of these assertions live in the cpp layer. I think we can remove them from here?

mettagrid/tests/test_leaks.py
```diff
@@ -1,35 +1,30 @@
...
+    cfg = get_cfg("benchmark")
 
-def test_mettagrid_env_reset(cfg):
-    """Test that the MettaGridEnv can be reset multiple times without memory leaks."""
     env = MettaGridEnv(env_cfg=cfg, render_mode=None)
     # Reset the environment multiple times
     for _ in range(10):
-        observation = env.reset()
-        assert observation is not None, "Reset should return a valid observation"
+        obs, infos = env.reset()
```
sasmith, line 25: Oh, that's a definite fix.

mettagrid/tests/test_leaks.py
```diff
@@ -38,50 +33,114 @@ def get_memory_usage():
...
-    memory_usage = []
+    np.random.seed(TEST_SEED)
+    random.seed(TEST_SEED)
 
-    for i in range(num_iterations):
-        # Create the environment
+    cfg = get_cfg("benchmark")
+
+    print("Pre-warming phase:")
+    for _i in range(20):  # Reduced from 20 to 10 to speed up the test
```
sasmith, line 53: Comment seems wrong

mettagrid/tests/test_leaks.py
sasmith, line 1: Can this be a separate PR?

sasmith (changes requested):
This is marked as `wip`, but is assigned to me. I'm sort of uncertain what's expected under such a circumstance, so I guess I made a `wip` review? As in, I've reviewed some-but-not-all.

This PR seems to do a lot more than what it says on the tin, which makes it harder to review. Understandable if it's a wip. Otherwise, I'll keep banging the drum of wanting pull requests to contain a single focused idea (or maybe "pull requests should be sized to make git bisect easier").

mettagrid/mettagrid/mettagrid_c.cpp
```diff
@@ -421,12 +413,20 @@ py::tuple MettaGrid::step(py::array_t<int> actions) {
...
     }
   }
 
-  return py::make_tuple(_observations, _rewards, _terminals, _truncations, py::dict());
+  std::vector<ssize_t> obs_shape = {static_cast<ssize_t>(_agents.size()),
+                                    static_cast<ssize_t>(obs_height),
+                                    static_cast<ssize_t>(obs_width),
+                                    static_cast<ssize_t>(_grid_features.size())};
+  py::array_t<c_observations_type> obs_view(obs_shape, _observations);
+  py::array_t<c_rewards_type> rewards_view(static_cast<ssize_t>(_rewards_size), _rewards);
```
sasmith, line 425: I expect this would be more correct as `... reward_view({ static_cast<ssize_t>(_rewards_size) }, ...`. This may be handled automatically by templating, but it would be nice to match the way we're providing a shape to observations.

sasmith: (nit towards numpy, etc) It feels silly for us to need to cast like this all the time. Can the tensors really have negative dimensions?

mettagrid/mettagrid/mettagrid_c.cpp
```diff
@@ -452,10 +452,10 @@ py::dict MettaGrid::grid_objects() {
     for (uint8_t i = 0; i < offsets.size(); i++) {
       offsets[i] = i;
     }
-    unsigned char obj_data[type_features.size()];
+    std::vector<unsigned char> obj_data(type_features.size());
```
sasmith, line 455: Thanks for cleaning up the warning.

mettagrid/mettagrid/mettagrid_c.hpp
```diff
@@ -81,20 +89,24 @@ class METTAGRID_API MettaGrid {
...
 
-  // We'd prefer to store these as more raw c-style arrays, but we need to both
-  // operate on the memory directly and return them to python.
-  py::array_t<unsigned char> _observations;
-  py::array_t<bool> _terminals;
-  py::array_t<bool> _truncations;
-  py::array_t<float> _rewards;
-  py::array_t<float> _episode_rewards;
+  // Pointers to external buffers - these are required and must be set
+  c_observations_type* _observations;
```
sasmith, line 97: Nice. Holding on to less python stuff seems cleaner; assuming we don't get burned in the recreating-the-objects translation layer.

mettagrid/tests/test_basic.py
```diff
@@ -1,83 +1,249 @@
...
+    assert environment._c_env is not None
+
+
+def test_grid_features(environment):
+    """
+    Test that the grid features from the C++ environment match the expected features list.
+    Ensures that the features are in the correct order and that all expected features are present.
+    """
+    # Expected features list based on the provided order
+    expected_features = [
```
sasmith, line 164: This seems brittle. I'd normally object at like a 2/5 or 3/5; but instead I'll note that we're going to switch to tokens, and so this will need to be reconsidered more thoroughly at that time.

sasmith (changes requested):
This seems like a good set of 5 PRs!

mettagrid/mettagrid/mettagrid_env.py
```diff
@@ -178,51 +233,43 @@ def process_episode_stats(self, infos: Dict[str, Any]):
...
+    @override
     def close(self):
         pass
 
     @property
-    def _max_steps(self):
-        return self._env_cfg.game.max_steps
-
-    @property
+    @required
```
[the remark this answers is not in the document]
rjwalters: this is a custom annotation that we wrote to help devs understand that some functions are expected by pufferlib without actually being overrides. seems weird to me too.
https://github.com/PufferAI/PufferLib/blob/9e1c5c3c2e1c288b673aaee6dbb42e0d6b154e11/pufferlib/environment.py#L34

metta/agent/policy_store.py
```diff
@@ -54,6 +54,13 @@ def policy(self) -> nn.Module:
             self._local_path = pr.local_path()
         return self._policy
 
+    def policy_as_metta_agent(self) -> Union[MettaAgent, DistributedMettaAgent]:
```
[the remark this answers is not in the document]
rjwalters: that is just a shorthand expression for Union I think?

metta/sim/simulation.py
```diff
@@ -96,15 +107,24 @@ def __init__(
...
+
         # Let every policy know the active action-set of this env.
-        action_names = self._vecenv.driver_env.action_names()
-        max_args = self._vecenv.driver_env._c_env.max_action_args()
-        self._policy_pr.policy().activate_actions(action_names, max_args, self._device)
+        action_names = metta_grid_env.action_names
+        max_args = metta_grid_env.max_action_args
+
+        metta_agent: MettaAgent | DistributedMettaAgent = self._policy_pr.policy_as_metta_agent()
+        assert isinstance(metta_agent, (MettaAgent, DistributedMettaAgent)), metta_agent
```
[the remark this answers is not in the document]
rjwalters: I think we need this for runtime safety

sasmith:
Fwiw, unito is having trouble syncing this. I believe this is an issue of the PR having comments that are >64k in size, and this being too large for Asana comments.

I think this is _not_ causing secondary problems -- just issues for this PR.

PR #517, opened 2025-05-16
eval smoke test

sasmith:
Add a smoke test for evals.

This refactors the run_evals script and sim.py to allow us to indicate we're running a smoke test. The smoke test runs on a much stripped down number of evals and simulations.

This also updates the policies we're using in run_evals to match https://github.com/Metta-AI/metta/blob/daphne-evaluation/run_evals.sh.

### DIFF
--- a/.github/workflows/py-checks.yml
+++ b/.github/workflows/py-checks.yml
@@ -316,7 +316,7 @@ jobs:
       - name: Run evals smoke test
         id: eval_smoke_test
         env:
-          WANDB_API_KEY: WANDB_API_KEY: ${{ secrets.WANDB_API_KEY }}
+          WANDB_API_KEY: ${{ secrets.WANDB_API_KEY }}
           AWS_ACCESS_KEY_ID: set_but_not_used
           AWS_SECRET_ACCESS_KEY: set_but_not_used
         run: |
--- a/.github/workflows/py-checks.yml
+++ b/.github/workflows/py-checks.yml
@@ -316,7 +316,7 @@ jobs:
       - name: Run evals smoke test
         id: eval_smoke_test
         env:
-          WANDB_API_KEY: set_but_not_used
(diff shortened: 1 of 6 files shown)


sasmith:
This PR also includes me testing the test in github.

Also: I make run_evals be part of the test, versus running `python3 -m tools.sim ...` more directly, since I wanted to test the actual script we're using as much as possible.

.github/workflows/py-checks.yml
```diff
@@ -313,6 +313,35 @@ jobs:
           retention-days: 1
           if-no-files-found: warn
 
+      - name: Run evals smoke test
+        id: eval_smoke_test
+        env:
+          WANDB_API_KEY: set_but_not_used
```
sasmith, line 319: To run an eval against an actual policy, we either need to provide an API key, or have the policy in our codebase, or something similar.

My intention is to provide an API key, but will give this a shot tomorrow.

sasmith:
@rjwalters This is ready for review, and afaik ready to be merged. I renamed "Training Smoke Test" to just "Some Tests", so we'll need to adjust the required tests when this gets merged.

tools/sim.py
```diff
@@ -73,6 +78,16 @@ def simulate_policy(
             vectorization=cfg.vectorization,
         )
         results = sim.simulate()
+
+        if sim_job.smoke_test:
```
rjwalters, line 82: can we avoid having this here? maybe we could move it to the github CI workflow or to a dedicated script that validates the envs that the smoke test uses

sasmith: Did you change your mind on this one? My intention was to try to exercise as much of the same code paths as evaluation as practical, and this seemed like the right way to do it. Happy to iterate based on this goal.

rjwalters (approved):

PR #519, opened 2025-05-16
fix: fix old_approx_kl loss term by reverting shape change

rjwalters:
This reverts commit dda5a0eb87b9ceb88f8c2f54d727d05c84ef98eb.

Changes in `old_approx_kl` loss term were caused by this block:

```
        # Create b_obs with explicit shape for minibatches
        if len(raw_obs.shape) == 4:  # [num_minibatches, minibatch_rows, bptt_horizon, *obs_shape]
            # Reshape to [num_minibatches, minibatch_size, *obs_shape]
            self.b_obs = raw_obs.reshape(self.num_minibatches, self.minibatch_size, *raw_obs.shape[3:])
        else:
            # If the shape is already appropriate, just use it directly
            self.b_obs = raw_obs
```

shape is provided by the sorted indices and we do not want to reshape here


daveey (approved):
approving without reading since it's a revert. if you want someone to review it, assign to @Al-does

PR #523, opened 2025-05-16
Change reset logic for Mettagrid environment

dominik3141:
Optimize buffer reset in MettaGrid::reset for improved performance by replacing loop-based clearing with direct memory operations using std::fill.


dominik3141:
My understanding of C++ is not great, I'm more of a pure C person, so what I wanted to do speaking in C is something like
```
memset(terminals, 0, n * sizeof(bool));
memset(truncations, 0, n * sizeof(bool));
memset(episode_rewards, 0, n * sizeof(float));
memset(rewards, 0, n * sizeof(float));
memset(observations, 0, obs_elem_count * sizeof(uint8_t));
```

sasmith (approved):
Cool. This is a fast-follow I'd intended, but not done.

PR #530, opened 2025-05-18
Fix pylance errors in config creation

berekuk:
`pyright tools metta | tail -1`: 314 -> 308 errors.

### TL;DR

Added `__init__ = Config.__init__` to Config subclasses to fix Pylance type checking issues.

### What changed?

Added the line `__init__ = Config.__init__` (or equivalent parent class's `__init__`) to all Config subclasses across the codebase. This includes:
- `AnalysisConfig`
- `DashboardConfig`
- `SimulationConfig` and `SingleEnvSimulationConfig`
- `ReplayJob`
- `SimJob`
- `TrainJob`

Also added a comment in the base `Config` class explaining why this pattern is necessary:
```python
# Sub-classes of Config class should use the `__init__ = Config.__init__` trick to satisfy Pylance.
# Without this, Pylance will complain about 0 positional arguments, because it looks up Pydantic's
# __init__ method, which takes no positional arguments.
```

### How to test?

1. Verify that Pylance no longer shows type errors when instantiating Config subclasses
2. Run existing tests to ensure functionality is preserved
3. Try creating instances of the modified Config subclasses to confirm they still work as expected

### Why make this change?

This change fixes type checking issues with Pylance. Without this fix, Pylance incorrectly complains about 0 positional arguments when instantiating Config subclasses because it looks up Pydantic's `__init__` method instead of the custom one in the `Config` class. This explicit inheritance of the `__init__` method ensures proper type checking while maintaining the same functionality.

### DIFF
--- a/metta/eval/analysis_config.py
+++ b/metta/eval/analysis_config.py
@@ -3,6 +3,7 @@ from metta.util.config import Config
 
 
 class AnalysisConfig(Config):
+    __init__ = Config.__init__
     # Policy URI to analyze
     policy_uri: str | None = None
     policy_selector: PolicySelectorConfig = PolicySelectorConfig()
--- a/metta/eval/dashboard/dashboard_config.py
+++ b/metta/eval/dashboard/dashboard_config.py
@@ -4,6 +4,7 @@ from metta.util.config import Config
 
 
 class DashboardConfig(Config):
+    __init__ = Config.__init__
     # Output options
     num_output_policies: int | Literal["all"] = 20
     metric: str = "reward"
--- a/metta/sim/simulation_config.py
+++ b/metta/sim/simulation_config.py
@@ -10,6 +10,8 @@ from metta.util.config import Config
 class SimulationConfig(Config):
     """Configuration for a single simulation run."""
 
+    __init__ = Config.__init__
+
     # Core simulation config
     num_episodes: int
     max_time_s: int = 120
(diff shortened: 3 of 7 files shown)


berekuk:
* **#530** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/530?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/530?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

metta/eval/analysis_config.py
```diff
@@ -3,6 +3,7 @@
 
 
 class AnalysisConfig(Config):
+    __init__ = Config.__init__
```
sasmith, line 6: Please add a comment about why we have these, since otherwise they're confusing noop cruft.

sasmith (approved):

metta/eval/analysis_config.py
berekuk: This is explained in the comment in `Config` class; we have a lot of Config subclass instances (most of them are small) so commenting on each one would be too verbose, I think?

Like, they're part of current `Config` class API. The API is awkward, but unless we find a better solution, it's going to become familiar to  everyone who uses it. Then it'd be like `call_awkward_api(...) # standard comment on how API is awkward`...

PR #531, opened 2025-05-19
Type-safe WandbContext; tests

berekuk:


berekuk:
* **#531** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/531?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/531?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

metta/util/wandb/wandb_context.py
```diff
@@ -52,25 +77,44 @@ def check_wandb_version() -> bool:
...
-        self.cfg = cfg
-        self.resume = resume
-        self.name = name or cfg.wandb.name
-        self.run_id = cfg.wandb.run_id or self.cfg.run or wandb.util.generate_id()
+    """
+    Context manager for Wandb.
+
+    Usually initialized in the following way:
+
+        with WandbContext(cfg.wandb, cfg) as wandb_run:
```
berekuk, line 85: This is intentionally more explicit - `WandbContext` shouldn't care about `cfg` shape specifics, IMO.
So all lookups under `cfg` were moved to `metta_research.yaml`.
(I don't really like resolvers but at least all those references to global vars are consolidated in one location)

metta/util/wandb/wandb_context.py
```diff
@@ -52,25 +77,44 @@ def check_wandb_version() -> bool:
 
 
 class WandbContext:
-    def __init__(self, cfg, job_type=None, resume=True, name=None, run_id=None, data_dir=None, timeout=30):
-        self.cfg = cfg
-        self.resume = resume
-        self.name = name or cfg.wandb.name
-        self.run_id = cfg.wandb.run_id or self.cfg.run or wandb.util.generate_id()
```
berekuk, line 59: > `wandb.util.generate_id()`

AFAICT we always have `run_id` (i.e., top-level `run`), please tell me if you can think of exceptions where it was needed.

metta/util/wandb/wandb_context.py
```diff
@@ -52,25 +77,44 @@ def check_wandb_version() -> bool:
 
 
 class WandbContext:
-    def __init__(self, cfg, job_type=None, resume=True, name=None, run_id=None, data_dir=None, timeout=30):
-        self.cfg = cfg
-        self.resume = resume
```
berekuk, line 57: Replaced with hardcoded `resume=True` but easy to get it back if necessary, through config.

configs/wandb/metta_research.yaml
```diff
@@ -1,7 +1,12 @@
...
 entity: metta-research
-group: ${.name}
-run_id: null
+group: ${run}
+name: ${run}
+run_id: ${run}
+
+data_dir: ${run_dir}
+
+job_type: ${cmd}
```
berekuk, line 12: This is the only intended change in behavior in this PR. Right now only train run get `job_type`, now all runs will have it.

tests/util/wandb/test_wandb_context.py
```diff
@@ -0,0 +1,150 @@
...
+    def fake_finish():
+        nonlocal finished
+        finished = True
+
+    monkeypatch.setattr(wandb, "finish", fake_finish)
+
+    ctx = WandbContext(cfg_on, global_cfg=OmegaConf.create({}))
+    _ = ctx.__enter__()
+    ctx.__exit__(None, None, None)
+    assert finished, "wandb.finish should be called on exit"
```
berekuk, line 150: (This file was mostly generated with ChatGPT)

tools/sweep_init.py
```diff
@@ -79,12 +79,16 @@ def create_run(sweep_name: str, cfg: DictConfig | ListConfig, logger: Logger) ->
...
     os.makedirs(run_dir, exist_ok=True)
     cfg.run = run_name
+    cfg.run_dir = run_dir
 
     def init_run():
-        with WandbContext(cfg, data_dir=run_dir) as wandb_run:
-            wandb_run_id = wandb_run.id
-            wandb_run.name = run_name
-            wandb_run.tags += (f"sweep_id:{sweep_cfg.wandb_sweep_id}", f"sweep_name:{sweep_cfg.sweep}")
+        with WandbContext(cfg.wandb, cfg) as wandb_run:
```
berekuk, line 85: This change was the most dangerous in this PR, since it's not easy to fully `sweep_init.py` locally, but I ran it & dumped the config from `WandbContext` to make sure that `cfg.wandb` doesn't get resolved too early (because then `cfg.run_dir` wouldn't affect `wandb.data_dir`), and it did the right thing.

metta/util/wandb/wandb_context.py
```diff
@@ -10,13 +11,37 @@
...
 logger = logging.getLogger(__name__)
 
 # Alias type for easier usage (other modules can import this type)
 WandbRun = wandb.sdk.wandb_run.Run
 
 
+class WandbConfigOn(Config):
+    enabled: Literal[True] = True
+
+    track: bool
```
daveey, line 27: in practice, we never run with track=false but enabled=true. feel free to remove this to simplify

metta/util/wandb/wandb_context.py
daveey: this may not be true in the sweep code, but i don't remember. we can fix it if we encounter it

tools/sweep_init.py
```diff
@@ -79,12 +79,16 @@ def create_run(sweep_name: str, cfg: DictConfig | ListConfig, logger: Logger) ->
...
+            if wandb_run:
```
daveey, line 86: we can't really do sweeps without wandb, so i would assert here rather than check if run

daveey (approved):
feel free to merge after fixes

PR #544, opened 2025-05-19
Supporting Externally Trained Policies Part 1

relh:
This PR adds simple support for external python classes used as policies, demo-ing with PufferAgentWrapper.

We add a top-level parameter to the `common.yaml` config called "policy_class", e.g.:
`policy_class: metta.agent.external.example.Recurrent`

We added two external models, one is the PufferLib default model in `agent/external/example.py` the other is a custom lstm_transformer in `agent/external/lstm_transformer.py`

`policy_store.py` now checks for `policy_class` being not null, and if it's set it tries to import the class and use it. 

There are still magic numbers in the `rl/pufferlib/policy.py` file for now.

`rl/pufferlib/trainer.py` is modified to accept MettaAgent/DistributedMettaAgent OR PufferAgentWrapper, though next step will be to try to roll PufferAgentWrapper into MettaAgent, where experiments either use a MettaAgentBuilder or the PufferAgentWrapper.

### DIFF
--- a/configs/common.yaml
+++ b/configs/common.yaml
@@ -8,7 +8,6 @@ dist_cfg_path: null
 data_dir: ./train_dir
 run_dir: ${data_dir}/${run}
 policy_uri: file://${run_dir}/checkpoints
-policy_class: null
 
 torch_deterministic: true
 vectorization: multiprocessing
--- a/configs/common.yaml
+++ b/configs/common.yaml
@@ -8,6 +8,7 @@ dist_cfg_path: null
 data_dir: ./train_dir
 run_dir: ${data_dir}/${run}
 policy_uri: file://${run_dir}/checkpoints
+policy_class: null
 
 torch_deterministic: true
 vectorization: multiprocessing
--- a/configs/sim/simple.yaml
+++ b/configs/sim/simple.yaml
@@ -1,9 +1,5 @@
 defaults:
-  - sim_suite
+  - sim_single
   - _self_
 
-name: simple 
-
-simulations:
-  simple/simple:
-    env: /env/mettagrid/simple
+env: /env/mettagrid/simple
--- a/configs/sim/simple.yaml
+++ b/configs/sim/simple.yaml
@@ -1,5 +1,9 @@
 defaults:
-  - sim_single
+  - sim_suite
   - _self_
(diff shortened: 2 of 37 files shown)


configs/common.yaml
```diff
@@ -8,6 +8,7 @@ dist_cfg_path: null
 data_dir: ./train_dir
 run_dir: ${data_dir}/${run}
 policy_uri: file://${run_dir}/checkpoints
```
daveey, line 10: we probably want to remove this from common. this should live under agent:

metta/agent/policy_store.py
```diff
@@ -484,7 +485,17 @@ def _make_codebase_backwards_compatible(self):
                     modules_queue.append(submodule_name)
 
     def _load_from_puffer(self, path: str, metadata_only: bool = False) -> PolicyRecord:
-        policy = load_policy(path, self._device)
+        if self._cfg.policy_class is not None:
```
daveey, line 488: this seems fine, but maybe it'll be easier to do with hydra.instanciate()?

daveey (changes requested):
good direction

configs/common.yaml
```diff
@@ -8,6 +8,7 @@ dist_cfg_path: null
...
+policy_class: null
```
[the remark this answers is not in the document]
relh: yup!

configs/sim/simple.yaml
```diff
@@ -1,5 +1,9 @@
 defaults:
-  - sim_single
+  - sim_suite
```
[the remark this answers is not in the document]
relh: yeah, I only changed it cause sim_single seemed broken but i'll pull it out

configs/common.yaml
relh: on it

metta/agent/policy_store.py
relh: trying now

Al-does (approved):
Nice!

relh:
I think I can neatly change the instantiate to a hydra instantiate with some configs and will do that

relh:
sigh I'll fix this all soon please ignore

PR #554, opened 2025-05-19
Fix map labels

daphnedemekas:
Map labels are being loaded before map_builder, so the plots for reward per label over time are not showing up on wandb


sasmith (approved):
Happy for this to be fixed, but if this is needed, it seems really fragile. I would expect `to_c_args` and `map_labels` to be read only operations, and so the order wouldn't matter.

Is there an assertion we could add to mettagrid_config or something?

daphnedemekas:
"I would expect to_c_args and map_labels to be read only operations, and so the order wouldn't matter."

I'm not sure what you mean by this.  to_c_args calls generate_map which loads self._map_builder, and map_labels uses self._map_builder. So if the order is switched self._map_builder is None and thus there are no labels

PR #556, opened 2025-05-19
Fix bug with replay urls

daphnedemekas:
If the length of replay URLs is zero, training crashes, so we now check for that.


metta/rl/pufferlib/trainer.py
```diff
@@ -640,12 +640,14 @@ def _generate_and_upload_replay(self):
             if self.wandb_run is not None:
                 replay_url = results.stats_db.get_replay_urls(
```
sasmith, line 641: Can we change this to `replay_urls`, to make it more obvious it's a list?

sasmith (approved):

PR #565, opened 2025-05-20
Revert a change to sim replay_dir that's breaking smoke tests

sasmith:
This reverts a change in https://github.com/Metta-AI/metta/pull/505 that causes smoke tests to fail, by requiring them to access s3.

### DIFF
--- a/configs/sim_job.yaml
+++ b/configs/sim_job.yaml
@@ -13,6 +13,6 @@ sim_job:
   simulation_suite: ${sim}
   stats_dir: ${run_dir}/stats
   stats_db_uri: ${run_dir}/stats.db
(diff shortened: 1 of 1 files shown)


sasmith:
@daphnedemekas -- at first glance, it's not clear if the change made to the replay dir was intentional. If so, we should update the smoke tests to be able to deal with it.

daphnedemekas:
> @daphnedemekas -- at first glance, it's not clear if the change made to the replay dir was intentional. If so, we should update the smoke tests to be able to deal with it.

Nice find, sorry about this. It was intentional, the replay dir needs to link to that in order for replays to appear on our dashboards.

PR #567, opened 2025-05-20
token based observation encoding

sasmith:
# Add token-based observation format

This PR adds a new observation format using tokens instead of a dense grid. Each token consists of a location, feature ID, and value. This allows for more efficient representation of observations, especially in sparse environments.

Key changes:
- Define `ObservationToken` struct to represent a single token (3 bytes)
- Add `obs_tokens` method to `GridObject` and implement for various object types
- Update `MettaGrid` to support both traditional and token-based observations
- Add configuration options `use_observation_tokens` and `num_observation_tokens`
- Update observation encoding logic to handle token-based format
- Add tests for token-based observations

This is guarded behind a flag for the moment, so shouldn't impact normal observations.

What this doesn't do:
- This doesn't update the observation shape in a way that's communicated to puffergrid. I.e., in actual training, we get our observation buffer via `set_buffer`, and this doesn't correctly advertise what shape that should be. So this doesn't work for training.
- This leaves the feature_ids as dependent on the order in which we read feature names. This means the feature_ids will be inconsistent over a range of changes to mettagrid.


mettagrid/mettagrid/objects/agent.hpp
```diff
@@ -106,6 +106,33 @@ class Agent : public MettaObject {
     return this->frozen;
   }
 
+  virtual size_t obs_tokens(ObservationTokens tokens, const std::vector<unsigned char>& feature_ids) const override {
+    // TODO: ablate 1-hot type_id encoding. Not ablating it now, since it should be easier to ablate once
+    // we've switched to the new observation format.
+    // Remove hp. Not removing it now for the same reason.
+    vector<uint8_t> basic_token_values = {1, group, hp, frozen, orientation, color};
+    size_t max_basic_tokens = tokens.size() > basic_token_values.size() ? basic_token_values.size() : tokens.size();
```
daveey, line 114: std::min()

mettagrid/mettagrid/objects/converter.hpp
```diff
@@ -133,6 +133,27 @@ class Converter : public HasInventory {
     this->maybe_start_converting();
   }
 
+  size_t obs_tokens(ObservationTokens tokens, const std::vector<unsigned char>& feature_ids) const override {
+    vector<uint8_t> basic_token_values = {1, this->hp, this->color, this->converting || this->cooling_down};
+    size_t max_basic_tokens = tokens.size() > basic_token_values.size() ? basic_token_values.size() : tokens.size();
+    size_t max_total_tokens = tokens.size() > max_basic_tokens + InventoryItem::InventoryCount
+                                  ? max_basic_tokens + InventoryItem::InventoryCount
```
daveey, line 140: don't love that this is duplicating agent.hpp. should we make inventory.hpp to encapsulate this?

mettagrid/mettagrid/objects/wall.hpp
```diff
@@ -18,6 +18,18 @@ class Wall : public MettaObject {
     this->_swappable = cfg["swappable"];
   }
 
+  size_t obs_tokens(ObservationTokens tokens, const std::vector<unsigned char>& feature_ids) const override {
```
daveey, line 21: can this go into MettaObject.obs_tokens(), and we just call super(basic_tokens)?

daveey (approved):

daveey (approved):

sasmith:
Rewrote to give each class a method of `vector<PartialObservations> obs_features()`, and then to have the base class that copies from the vector into the memory buffer, limited by the size.

I _think_ this does result in an unreasonable amount of copying the features around, but feel free to take another look.

Merging, since this is behind a flag.

PR #568, opened 2025-05-20
fix: update mettagrid makefile to track header files

rjwalters:
**Fix incremental build system by adding proper dependency tracking**

### Problem
The Makefile required `make clean` before every build due to missing header dependency tracking. Changes to `.h`/`.hpp` files wouldn't trigger recompilation of dependent `.cpp` files, leading to stale builds and linking errors.

### Solution
1. **Added automatic dependency generation** using GCC's `-MMD -MP` flags
2. **Implemented build configuration tracking** to detect environment changes (Python version, compiler flags, etc.)
3. **Removed duplicate compilation rules** that were overriding dependency-aware versions
4. **Added ccache integration** for faster rebuilds

### Key Changes
- **Dependency files**: Generated `.d` files now track header dependencies automatically
- **Build config tracking**: `build/.build_config` file triggers full rebuild when environment changes
- **Clean rule hierarchy**: Single compilation rule per target type with proper dependency tracking
- **Debug targets**: Added `debug-deps` and `debug-build-state` for troubleshooting

### Result
- **Header changes** now correctly trigger recompilation of affected files only
- **Source changes** rebuild just the modified file + linking
- **Environment changes** trigger full rebuild when needed
- **No changes** result in near-instant builds
- **Eliminates need for `make clean` on every build** while maintaining correctness


berekuk (approved):
I haven't read the Makefile carefully to check whether it does the right thing, but I assume you've tested this.
(we discussed in the DMs how what we really want is some other build system)

PR #572, opened 2025-05-20
Use pip install -e instead of setup.py; simplify setup_build; call uv commands directly in cloud jobs

berekuk:
Tested:
- locally (both with `--clean` and without)
- on AWS Batch
- on Skypilot

(graphite description below)

# Simplify Build Process and Improve Environment Setup

This PR simplifies our build and installation process by:

- Replacing manual `setup.py build_ext` calls with direct `pip install -e .` commands
- Streamlining the `setup_build.sh` script to create and manage virtual environments more reliably
- Updating Docker, AWS Batch, and SkyPilot configurations to use the simplified installation approach
- Improving the Makefile by removing redundant build targets and clarifying installation steps
- Updating the mettagrid setup to use pybind11's built-in helpers for more reliable extension building

These changes make our development environment setup more consistent across different platforms and simplify the onboarding process for new developers.


berekuk:
* **#572** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/572?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/572?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

devops/setup_build.sh
```diff
@@ -22,9 +23,9 @@ done
 
 # check if we're in docker
 if [ -f /.dockerenv ]; then
-  export IS_DOCKER=true
+  export IS_DOCKER=1
 else
-  export IS_DOCKER=false
```
berekuk, line 27: `false` value meant that dev installs (vscode, mettascope, aws) were never called

devops/setup_build.sh
```diff
@@ -72,9 +73,7 @@ if ! command -v uv &> /dev/null; then
   fi
 fi
 
-# Required Python version
-REQUIRED_PYTHON_VERSION="3.11.7"
```
berekuk, line 76: not necessary, `uv venv` automatically picks up `.python-version`

devops/setup_build.sh
```diff
@@ -91,168 +90,83 @@ if [ "$CLEAN" -eq 1 ]; then
...
     echo "Checking for conda environments associated with this project..."
-    PROJECT_NAME=$(basename "$PROJECT_DIR")
+    PROJECT_NAME=$(basename "$(pwd)")
     CONDA_ENVS=$(conda env list | grep "$PROJECT_NAME" | awk '{print $1}')
     if [ -n "$CONDA_ENVS" ]; then
       echo "⚠️  Found the following conda environments that might be related to this project:"
       echo "$CONDA_ENVS"
       echo "⚠️  You may want to manually remove these if they're no longer needed (conda env remove -n ENV_NAME)"
     fi
   fi
```
berekuk, line 134: I'm not sure conda check is useful, it'll be safe to remove it after another week or two.

devops/setup_build.sh
```diff
@@ -91,168 +90,83 @@ if [ "$CLEAN" -eq 1 ]; then
...
-    exit 1
-  fi
-}
+uv pip install -r requirements.txt
 
-# ========== BUILD FAST_GAE ==========
-echo -e "\nBuilding FastGAE..."
-make build
+echo -e "\nInstalling Metta..."
+uv pip install -e .
```
berekuk, line 143: This is why I started on this PR: `pip install -e` always calls `build-backend` from `pyproject.toml`, so we don't ever have to run `setup.py` ourselves.

Note that the logs are less verbose in this case (C++ build commands are not logged), but they will still show up if something goes wrong. And it's possible to enable verbose logs with `uv pip install -v`.

There might be a way to never even call `pip install -e mettagrid` (can we put `-e mettagrid` in top-level `requirements.txt`?), but I couldn't make it to work yet.

devops/skypilot/install.sh
```diff
@@ -12,4 +8,8 @@ if [ -z "$SERVER" ]; then
   exit 1
 fi
 
+# Install SkyPilot directly into the .venv
+# This is not in requirements_pinned.txt because want to make sure that users are using our remote API server
+uv pip install skypilot==0.9.2 --prerelease=allow
```
berekuk, line 13: Changed the order - first acquire the server endpoint, to make sure we have access to AWS. Then install skypilot.

So there's less chance that someone will run jobs from their default local API server.

mettagrid/setup.py
```diff
@@ -59,6 +49,6 @@ def build_extensions(self):
     version="0.1.6",
     packages=["mettagrid"],
     ext_modules=ext_modules,
-    cmdclass={"build_ext": CustomBuildExt},
+    cmdclass={"build_ext": build_ext},
```
berekuk, line 52: Context: at some point I had issues with `pip install` -> `setup.py` call, setup wasn't called. ChatGPT suggested that I should update this module with a modern configuration, something about PEP-517 and PEP-660.

Then somehow it became ok even without this change, I'm not sure why.

Because this change simplifies setup code a bit, I decided to keep it in, but it's not necessary.

devops/setup_build.sh
```diff
@@ -91,168 +90,83 @@ if [ "$CLEAN" -eq 1 ]; then
...
-cd mettagrid
-make install-dependencies
-make build
-cd ..
-
-# ========== INSTALL SKYPILOT ==========
-echo -e "\nInstalling Skypilot..."
-uv tool install skypilot --from 'skypilot[aws,vast,lambda]'
+echo -e "\nInstalling MettaGrid..."
+uv pip --directory mettagrid install -e .
```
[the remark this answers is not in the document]
rjwalters: check this

rjwalters (approved):
lgtm

devops/setup_build.sh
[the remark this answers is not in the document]
berekuk: This is fine, `--directory` is a global option

berekuk:
I'm not merging yet because I realized Dockerfile could fail (it's missing `uv venv`, for one thing), so I want to test it first.

berekuk:
Tested by building an image on a tmp skypilot cluster, the image with `uv venv` works.

PR #573, opened 2025-05-20
feat: add workflow for launching AWS training jobs

rjwalters:
This PR adds a new flag to our AWS launch script that allows a job to self-terminate after a certain number of minutes. 

The intention is to use this in a new workflow that will train for ~30 min on env/simple for every push to main so that we can spot real world performance problems earlier.

### DIFF
--- a/.github/workflows/aws-train.yml
+++ b/.github/workflows/aws-train.yml
@@ -1,0 +1,99 @@
+name: Launch AWS Training Job
+
+# This workflow can be manually triggered from the GitHub UI
+on:
+  workflow_dispatch:
+    inputs:
+      timeout_minutes:
+        description: 'Job timeout in minutes (auto-termination)'
+        required: true
+        default: '60'
+        type: number
+      trainer_env:
+        description: 'Training environment configuration'
+        required: true
+        default: 'env/mettagrid/simple'
+        type: string
+      pr_number:
+        description: 'PR number (if applicable, leave empty otherwise)'
+        required: false
+        type: string
+
+
+jobs:
+  launch-batch-job:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # This is required for AWS credentials
+      contents: read # This is required to checkout the repository
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Python and uv
+        id: setup-python-uv
+        uses: ./.github/actions/setup-python-uv
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+          aws-region: us-east-1
+
+      - name: Get current commit hash
+        id: get_commit_hash
+        run: |
+          COMMIT_HASH=$(git rev-parse HEAD)
+          echo "commit_hash=$COMMIT_HASH" >> $GITHUB_OUTPUT
+          echo "Commit hash: $COMMIT_HASH"
+
+      - name: Generate run name
+        id: generate_run_name
+        run: |
+          # Get short commit hash
+          COMMIT_HASH=$(git rev-parse --short HEAD)
+          
+          # Get current timestamp in a format suitable for filenames
+          TIMESTAMP=$(date -u +"%Y%m%d_%H%M%S")
+          
+          # Get branch name
+          BRANCH=$(git rev-parse --abbrev-ref HEAD | sed 's/[^a-zA-Z0-9]/_/g')
+          
+          # Generate the run name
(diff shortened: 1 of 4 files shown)


.github/workflows/aws-train.yml
```diff
@@ -0,0 +1,99 @@
...
+          else
+            # Use branch name if no PR number provided
+            RUN_NAME="github.$BRANCH.$COMMIT_HASH.$TIMESTAMP"
+            echo "Using branch name '$BRANCH' in run name"
+          fi
+          
+          echo "Generated run name: $RUN_NAME"
+          echo "run_name=$RUN_NAME" >> $GITHUB_OUTPUT
+
+      - name: Launch AWS Batch job
```
sasmith, line 77: expect to need to change this for sky pilot

devops/aws/batch/entrypoint.sh
```diff
@@ -10,29 +10,185 @@ set -e
...
+# - JOB_TIMEOUT_MINUTES: Optional timeout in minutes for auto-termination
 #
 # AWS Batch environment variables used:
+# - AWS_BATCH_JOB_ID: The ID of the current job
 # - AWS_BATCH_JOB_NODE_INDEX: Index of this node in the job
 # - AWS_BATCH_JOB_MAIN_NODE_INDEX: Index of the main node
 # - AWS_BATCH_JOB_NUM_NODES: Total number of nodes in the job
 
-# Link training directory
-ln -s /mnt/efs/train_dir train_dir 2> /dev/null || true
```
sasmith, line 20: Shuffling things like this causes a headache for me as a reviewer. Please LLM, don't reorder unnecessarily!

devops/aws/batch/entrypoint.sh
```diff
@@ -10,29 +10,185 @@ set -e
...
-source ./devops/env.sh
+# Source environment variables first to set up any paths
+source ./devops/env.sh 2>/dev/null || true
 
-# Setup log directory and file
+# Setup log directory and file early
 export NODE_INDEX=${AWS_BATCH_JOB_NODE_INDEX:-0}
 export LOG_FILE="train_dir/logs/${JOB_NAME}.${NODE_INDEX}.log"
-mkdir -p $(dirname $LOG_FILE)
+mkdir -p $(dirname $LOG_FILE) 2>/dev/null || true
```
sasmith, line 27: The `|| true` means we'll continue even if this fails. That seems like the wrong behavior. Or at least, it's not clear why we'd want to.

devops/aws/batch/entrypoint.sh
```diff
@@ -10,29 +10,185 @@ set -e
...
+
+# Function to log timeout-related messages to both stdout and our timeout log
+function timeout_log() {
+    local message="$1"
+    local timestamp=$(date +"%Y-%m-%d %H:%M:%S")
+    echo "[$timestamp] TIMEOUT: $message" | tee -a "train_dir/logs/${JOB_NAME}.timeout.log" 2>/dev/null || echo "[$timestamp] TIMEOUT: $message"
```
sasmith, line 33: This feels dangerously fancy, but looks right. Or at least, I understand the purpose of the fanciness.

devops/aws/batch/entrypoint.sh
```diff
@@ -10,29 +10,185 @@ set -e
...
+}
+
+# Set up the auto-termination feature if JOB_TIMEOUT_MINUTES is set
+if [ ! -z "$JOB_TIMEOUT_MINUTES" ] && [ "$NODE_INDEX" = "0" -o "$AWS_BATCH_JOB_NODE_INDEX" = "0" ]; then
+    # Create timeout log file
+    mkdir -p "train_dir/logs" 2>/dev/null || true
```
sasmith, line 39: Same as above -- seems more correct to fail if we can't create this.

devops/aws/batch/entrypoint.sh
sasmith: So I guess, decide on how defensive we want to be, and then add a comment about what we're doing, and why. I think that will be important to human and LLM consumers alike.

By default, I think the right behavior is "fail on error". I think defensiveness could be right if this is "we want to try to tear down even if the job is in bad shape".

sasmith (changes requested):
This seems fine, though it's worth considering other ways to terminate the jobs (e.g., a lambda watchdog, which would conveniently live outside of the job).

What I'd want more than this is some reference to why we're running these jobs. This would help inform things like
* Is 60m the right amount of time?
* What monitoring / alerting should we have?
* Do we want to have 1 training per commit, or to have some throttling so we have 1 training per hour?

sasmith (approved):

berekuk:
In my understanding we'll get rid of AWS Batch in favor of Skypilot in the next week or two.

PR #579, opened 2025-05-20
Updated prioritized env set that also samples across parameters

daphnedemekas:
Previosuly, our prioritized environment replay class only sampled across environment configs, constructing a dsitribution over the configs based on how much reward the agents were getting over time in each config (we upsample environments that get low reward over time). This is being done with an importance sampling, to correct the bias introduced for harder environments, and a tunable parameter of how sharply we should be doing this prioritization. 

Now, we are also doing importance sampling within each environment config, over all parameters that are being sampled (for example if we include $sampling over width and height. This is done by splitting the given range into bins, and again starting with a uniform distribution over bins, we sample a bin at every timestep, update the distribution with that range, and then keep track of performance per parameter bin, so that we can construct bin-level prioritized within configs.


metta/env/mettagrid_env_set.py
```diff
@@ -163,41 +160,153 @@ def get_env_stats(self):
...
+        self._epsilon = epsilon
+
+        # Env-level trackers
+        self._env_priorities = np.ones(self._num_envs, dtype=np.float32)
+        self._env_visits = np.zeros(self._num_envs, dtype=np.int32)
+
+        # Sampling bins
+        self._num_bins = num_bins
+
+        # Discover sampling macros and bin them
```
sasmith, line 202: "Discover `${sampling(...)}` macros in each environment and bin them, so we can do prioritized replay across them."

(Mostly, it took me a while to parse this, and I'm trying to figure out what nudges I can do to the comments to help a future me.)

metta/env/mettagrid_env_set.py
```diff
@@ -163,41 +160,153 @@ def get_env_stats(self):
...
+            # inject via bracket-path
+            OmegaConf.update(raw, path_expr, bins[i], merge=False)
+
+        # resolve macros
+        resolved = OmegaConf.to_container(raw, resolve=True)
+        return OmegaConf.create(resolved)
+
+    def reset(self, *args, **kwargs):
+        if hasattr(self, "_last_episode_reward"):
+            err = abs(-self._last_episode_reward) + self._epsilon
```
sasmith, line 285: The `abs` and `-` seem sort of weird together. Why take a negative if you're then taking abs?

metta/env/mettagrid_env_set.py
```diff
@@ -163,41 +160,153 @@ def get_env_stats(self):
...
+            prio = err**self._alpha
+            e = self._current_choice["env"]
+            # update env
+            self._env_priorities[e] = prio
```
sasmith, line 289: It's interesting to use the same weight for the env as for each of the parameters you just used. Then the priority of getting that env and those same parameters is like (prio**2).

metta/env/mettagrid_env_set.py
```diff
@@ -163,41 +160,153 @@ def get_env_stats(self):
...
+        num_bins: int = 7,
+        **kwargs,
+    ):
+        # Base configs
+        self._env_cfgs = env_cfg.envs
+        self._num_envs = len(self._env_cfgs)
+
+        # Prioritized replay parameters
+        self._alpha = alpha
+        self._beta = beta
```
sasmith, line 191: I see this get set and updated, but not used. Can you say more?

I also see this is used in `MettaGridEnvSet`, but only seems to be used in `_get_importance_sampling_weight`, which doesn't appear to be called.

metta/env/mettagrid_env_set.py
```diff
@@ -163,41 +160,153 @@ def get_env_stats(self):
...
+            OmegaConf.update(raw, path_expr, bins[i], merge=False)
+
+        # resolve macros
+        resolved = OmegaConf.to_container(raw, resolve=True)
+        return OmegaConf.create(resolved)
+
+    def reset(self, *args, **kwargs):
+        if hasattr(self, "_last_episode_reward"):
+            err = abs(-self._last_episode_reward) + self._epsilon
+            prio = err**self._alpha
```
sasmith, line 286: I'd love some comment on how to think about this. Seems like we're strictly prioritizing maps were we got better scores. Deprioritizing maps where we're doing really badly makes sense, since we may be too far away from learning anything. But I would have expected some down-weighting of maps that we're acing.

I can see justification in a lot of directions; but would love to have something written up about the intention here.

sasmith (changes requested):
Sorry for the delay.

LMK if you'd like to go over some of the math with me on a whiteboard.

PR #581, opened 2025-05-20
Adding performance benchmarks for environment

dominik3141:
Adding two new benchmarks:
1. Agent steps per second
2. Environment reset time

Both metrics are measured and logged to bencher.dev on every new PR.


.github/workflows/py-checks.yml
```diff
@@ -465,6 +465,26 @@ jobs:
 
           print("Successfully combined benchmark results.")
 
+          # Extract agent_steps_per_second and write Bencher Metric Format file
```
rjwalters, line 468: are you sure we need to modify CI at all? I think the new benchmarks should just be picked up by bencher automatically...

mettagrid/benchmarks/test_mettagrid_env_benchmark.py
```diff
@@ -47,3 +47,91 @@ def run_step():
         rounds=10,  # Number of rounds to run
         warmup_rounds=0,  # Number of warmup rounds to discard
     )
+
+
+def test_step_performance_no_reset(benchmark, environment, single_action):
```
rjwalters, line 52: these are great! - can we delete some of the old ones?

.github/workflows/py-checks.yml
rjwalters: reading up on this more I see that we only get latency results automatically and that we have to transform the data into a bmf file one way or another if we want to report a throughput... I think this approach is fine for now, but long term we should probably combine this with the `save-benchmarks` action and streamline how we are building these json files

rjwalters (approved):
looks good to me! we should consolidate the benchmarks json file stuff with the custom action and try to just upload one generated bmf file

mettagrid/benchmarks/test_mettagrid_env_benchmark.py
dominik3141: The only reason I didn't delete the `test_step_performance` function (without reset) was to not give up on a metric for which we already have some historical data. The benchmarking is so cheap that deletion seems to have a rather weak incentive.

.github/workflows/py-checks.yml
dominik3141: We can definitely do that! I also think we should see if there are more interesting metrics to collect in bencher. It is cheap and really helpful. I don't have anything particular in mind right now, but I think it would be a good task for CodeX or jules.

PR #583, opened 2025-05-21
navigation sequence evals

georgejwdeane:
These are navigation evals with the mine --> generator --> ore sequence

### DIFF
--- a/configs/env/mettagrid/navigation_sequence/evals/all.yaml
+++ b/configs/env/mettagrid/navigation_sequence/evals/all.yaml
@@ -1,0 +1,96 @@
+defaults:
+  - /env/mettagrid/mettagrid@
+  - _self_
+
+game:
+  num_agents: 14
+  max_steps: 250
+
+  map_builder:
+    _target_: mettagrid.room.multi_room.RoomList
+    rooms:
+      - _target_: mettagrid.room.ascii.Ascii
+        border_width: 1
+        uri: "configs/env/mettagrid/maps/navigation/corridors.map"
+      - _target_: mettagrid.room.ascii.Ascii
+        border_width: 1
+        uri: "configs/env/mettagrid/maps/navigation/cylinder.map"
+      - _target_: mettagrid.room.mean_distance.MeanDistance
+        width: 25
+        height: 25
+        mean_distance: 15
+        border_width: 3
+        agents: 1
+        objects:
+          altar: 3
+      - _target_: mettagrid.room.mean_distance.MeanDistance
+        width: 35
+        height: 35
+        mean_distance: ${sampling:20,30,30}
+        border_width: 3
+        agents: 1
+        objects:
+          altar: 3
+      - _target_: mettagrid.room.mean_distance.MeanDistance
+        width: 12
+        height: 12
+        mean_distance: 10
+        border_width: 3
+        agents: 1
+        objects:
+          altar: 3
+      - _target_: mettagrid.room.cognitive_evals.labyrinth.LabyrinthMaze
+        width: 35
+        height: 35
+        corridor_width: 3
+        onlyhearts: true
+      - _target_: mettagrid.room.ascii.Ascii
+        border_width: 1
+        uri: "configs/env/mettagrid/maps/navigation/obstacles0.map"
+      - _target_: mettagrid.room.ascii.Ascii
+        border_width: 1
+        uri: "configs/env/mettagrid/maps/navigation/obstacles1.map"
+      - _target_: mettagrid.room.ascii.Ascii
+        border_width: 1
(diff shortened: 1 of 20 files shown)


sasmith:
Hello from the present!

If you assign this PR to someone, it should show up as an Asana task for them. For this reason / clarity, I (as the Github AoR holder) prefer for people to assign PRs to folks.

configs/env/mettagrid/maps/diversity/left_or_right.yaml
sasmith, line 1: I can't see why these were deleted. Can you say more?

sasmith:
> Hello from the present!
> 
> If you assign this PR to someone, it should show up as an Asana task for them. For this reason / clarity, I (as the Github AoR holder) prefer for people to assign PRs to folks.

Oh, this also makes it more clear that you're ready for a code review. Sometimes people aren't.

configs/env/mettagrid/maps/object_use/shoot_out_sequence.map
sasmith, line 1: Not clear why this was deleted. If this should go, can you move it to a separate PR?

sasmith: If it's "this was removed because of the changes to how maps are encoded", then this makes sense; although my main request would be "split that off first, then add all the nav_sequence files".

configs/env/mettagrid/navigation_sequence/corridors.yaml
```diff
@@ -0,0 +1,23 @@
+defaults:
+  - /env/mettagrid/navigation_sequence/defaults@
+  - _self_
+
+game:
+  num_agents: 1 #how many agents are in the map x2
```
sasmith, line 6: The comment here doesn't make sense to me.

configs/env/mettagrid/navigation_sequence/corridors.yaml
```diff
@@ -0,0 +1,23 @@
...
+  max_steps: 600
+
+  agent:
```
sasmith, line 9: I think we could add a lot of these to the defaults file. E.g., heart requests, most of map_builder targets.

It would also let us be more consistent with things like "border_width" which seems to change sometimes.

configs/env/mettagrid/navigation_sequence/corridors.yaml
```diff
@@ -0,0 +1,23 @@
...
+    _target_: mettagrid.room.multi_room.MultiRoom
+    num_rooms: 1
+    room:
+      _target_: mettagrid.room.ascii.Ascii
+      border_width: 1
+      uri: "configs/env/mettagrid/maps/navigation_sequence/corridors.map"
+
+  objects:
+    altar:
+      cooldown: 255
```
sasmith, line 23: Can you say what the intention here is? 255 smells like "cool down as infrequently as possible". If we didn't have limitations, would you set this to something else?

configs/env/mettagrid/navigation_sequence/defaults.yaml
```diff
@@ -0,0 +1,44 @@
...
+game:
+
+  num_agents: 1
+  max_steps: 700
+
+  agent:
+    max_inventory: 100
+    freeze_duration: 0
+    hp: 10
+    mortal: False
```
sasmith, line 16: `mortal` isn't meaningful. Can you remove it?

configs/env/mettagrid/navigation_sequence/defaults.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+  objects:
+    altar:
```
sasmith, line 19: Ah! I see that a lot of things in the map files are already in defaults. Can you remove the redundant settings from the other yaml files?

mettagrid/mettagrid/room/ascii.py
sasmith, line 1: For as much as this requires a change to existing maps, can you check that change in first, and then add the navigation sequence maps as a separate PR?

daphnedemekas (approved):

configs/env/mettagrid/maps/diversity/left_or_right.yaml
georgejwdeane: anything in 'maps' should be a .map file, and these are yamls!

configs/env/mettagrid/navigation_sequence/corridors.yaml
georgejwdeane: You smelled correctly. This is necessary in these evals to cap their episode reward at 1 so they can't reuse altars. 255 is fine for these environments in longer would be good to have the option of infinite cooldown to make objects single use

PR #594, opened 2025-05-21
Add Action Buttons

treeform:
* Ads action buttons as a UI you can click on screen:

image

Still not done, normal attack selection and color selection.


rjwalters:
can we just use the arrow keys / awsd for moving in the desired direction and have the code figure out the rotations?

treeform:
@rjwalters thats how the keyboard works. Keyboard is the correct way to use it. Though I wanted to keep the gui buttons in "natural mode" of rotations and motion separate so that people know how the agents really feel. One could argue that it should be a mode or some thing, but I feel like its a separate feature request.

daveey (approved):

PR #596, opened 2025-05-21
Memory evals

georgejwdeane:
Aound 20 memory/ spatial memory/ navigation evals


daphnedemekas (commented):
Can you clean up the yaml files so we don't have any redundancy on the yaml files and on defaults

daphnedemekas (approved):

daphnedemekas (changes requested):
Need to go through the memory evals and check the times/ layout

daphnedemekas (changes requested):
This evals are great! But looking at the replays, I don't think they have enough time for a lot of them: 
https://s3.amazonaws.com/softmax-public/policydash/memory.html

If you could go through and adjust the times, that would be great. 

Also, before merging, please delete the run_memory_evals.sh file

daphnedemekas (approved):

PR #598, opened 2025-05-22
ascii update

georgejwdeane:
This is an update to the symbols for ascii maps that is compatible with all of the new configs.


daphnedemekas (approved):
I'm okay with this. I wonder if we keep adding symbols if we want to move towards a different way of doing this though

PR #599, opened 2025-05-22
Update inventory management

sasmith:
Refactors inventory management to make objects more in control of when they can and can't receive input.

This changes to make it so put and get return the actual delta of items, rather than having outside logic assume what the agent / converter can take. The main purpose of this is to make it so we can [later] have per-item inventory limits, and have the logic to manage those contained in the `update_inventory` function. As a side effect of this change, I fixed some places where items could get lost, like when you're stealing from a frozen agent.

This still makes some assumptions that entities will be able to _lose_ inventory if they have it. If we cared, we could also expose functions like "max_potential_update", rather than only exposing it through mutation.

Testing:
Added unit tests.

### DIFF
--- a/mettagrid/mettagrid/objects/has_inventory.hpp
+++ b/mettagrid/mettagrid/objects/has_inventory.hpp
@@ -3,6 +3,7 @@
 
 #include <map>
 #include <string>
+#include <iostream>
 
 #include "constants.hpp"
 #include "metta_object.hpp"
--- a/mettagrid/mettagrid/objects/has_inventory.hpp
+++ b/mettagrid/mettagrid/objects/has_inventory.hpp
@@ -1,9 +1,9 @@
 #ifndef HAS_INVENTORY_HPP
 #define HAS_INVENTORY_HPP
 
+#include <algorithm>
 #include <map>
 #include <string>
-#include <algorithm>
 
 #include "constants.hpp"
 #include "metta_object.hpp"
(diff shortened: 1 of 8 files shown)


sasmith:
* **#599** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/599?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/599?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mettagrid/mettagrid/objects/has_inventory.hpp
```diff
@@ -20,14 +20,17 @@ class HasInventory : public MettaObject {
...
-      amount = 255 - this->inventory[item];
+  virtual int update_inventory(InventoryItem item, short amount) {
+    int initial_amount = this->inventory[item];
+    int new_amount = initial_amount + amount;
+    if (new_amount > 255) {
+      new_amount = 255;
     }
-    if (amount + this->inventory[item] < 0) {
-      amount = -this->inventory[item];
+    if (new_amount < 0) {
```
daveey, line 29: max(new_amount, 0)

daveey (approved):

mettagrid/mettagrid/objects/has_inventory.hpp
sasmith: clamp(new_amount, 0, 255)

sasmith:
### Merge activity

* **May 22, 10:09 PM UTC**: @sasmith merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Metta-AI/metta/599).

PR #602, opened 2025-05-22
fix: wandb_context test

rjwalters:
This PR replaces the deprecated `pkg_resources` library with the standard `importlib.metadata` in the wandb context module, and fixes test isolation issues that were creating an unwanted config.yaml file during test runs.

**Changes:**
- Replace deprecated `pkg_resources` with `importlib.metadata` (eliminates deprecation warnings)
- Add file operation mocking to prevent tests from creating `config.yaml` files in the working directory

### DIFF
--- a/metta/util/wandb/wandb_context.py
+++ b/metta/util/wandb/wandb_context.py
@@ -2,9 +2,9 @@ import copy
 import logging
 import os
 import socket
+from importlib.metadata import PackageNotFoundError, version
 from typing import Annotated, Literal, Union, cast
 
-import pkg_resources
 import requests
 import wandb
 import wandb.errors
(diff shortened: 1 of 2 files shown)


berekuk (changes requested):
Thanks for taking care of this! I've noticed dummy `config.yaml` but didn't get around to fixing it.

I think this PR is too verbose for the amount of value it creates, though. Every new line of code brings other potential bugs and add maintenance costs and cognitive costs when the code needs to be changed again.

It's not a huge deal in tests, but recently I've been through several situations where I wanted to change some part of the codebase and most of my time was spent on reading through big chunks of code that weren't doing anything particularly useful, so I'm paying attention to that.

rjwalters:
@berekuk and I discussed this and he wants to clean things up a different way so we are closing this one

PR #604, opened 2025-05-22
Allow item-specific maximums

sasmith:
This PR enhances the inventory system to support different maximum capacities for each item type. Previously, all items shared a single maximum value defined by `max_inventory`. Now:

- Added support for item-specific maximums via `item_name_max` configuration
- Maintained backward compatibility with the existing `max_inventory` parameter
- Added a new `default_item_max` parameter as an alternative to `max_inventory`
- Updated the inventory update logic to respect item-specific limits
- Added tests to verify different maximum capacities for red and green ore

The change allows for more flexible game design where certain resources can be more limited than others.

### DIFF
--- a/mettagrid/mettagrid/objects/agent.hpp
+++ b/mettagrid/mettagrid/objects/agent.hpp
@@ -44,7 +44,8 @@ public:
     this->freeze_duration = cfg["freeze_duration"];
     this->orientation = 0;
     this->inventory.resize(InventoryItem::InventoryCount);
-    unsigned char default_item_max = cfg["max_inventory"] || cfg["default_item_max"];
+    // We can specify default_item_max two ways, for backwards compatibility.
+    unsigned char default_item_max = cfg["max_inventory"] > 0 ? cfg["max_inventory"] : cfg["default_item_max"];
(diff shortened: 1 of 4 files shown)


mettagrid/mettagrid/objects/agent.hpp
```diff
@@ -43,7 +43,16 @@ class Agent : public MettaObject {
     this->freeze_duration = cfg["freeze_duration"];
     this->orientation = 0;
     this->inventory.resize(InventoryItem::InventoryCount);
-    this->max_items = cfg["max_inventory"];
+    // We can specify default_item_max two ways, for backwards compatibility.
+    unsigned char default_item_max = cfg["max_inventory"] > 0 ? cfg["max_inventory"] : cfg["default_item_max"];
```
daveey, line 47: I appreciate the backward comp, but wonder if it would be better to just replace the old flag with the new in all the configs. That way there is less confusion

daveey (approved):

mettagrid/mettagrid/objects/agent.hpp
sasmith: Yeah -- I'll update files and kill this in a following PR.

PR #606, opened 2025-05-22
Adds new battery colors

sasmith:
Renamed `battery` to `battery.red` and added support for `battery.blue` and `battery.green` in the MettagGrid environment.

This lets us build more intricate training and evals for agents.

Downsides to consider before we merge this!
* This will change the dimensions / properties for agents, and so will invalidate existing policies.
* This increases the number of observation layers from 34 -> 38 (agents and converters have separate inventory features). Based on experimenting, I believe this increase will slow down training speed by about 6% (probably due to increased model complexity).
  * This penalty applies to our current convolution-based models, but not our future token-based models. So maybe it's fine to just take the hit for now.


sasmith:
* **#606** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/606?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/606?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rjwalters:
we could reduce the size of the vector by collapsing agent and converter inventory into the same slots...

mettagrid/tests/mettagrid_test_args_env_cfg.json
```diff
@@ -125,7 +125,7 @@
       "generator.green": {
         "hp": 30,
         "input_ore.green": 1,
-        "output_battery": 1,
+        "output_battery.red": 1,
```
daphnedemekas, line 128: I think that green generators should output green battery by default / blue generatros blue battery

mettagrid/tests/mettagrid_test_args_env_cfg.json
```diff
@@ -18,8 +18,8 @@
         "ore.red": 0.005,
         "ore.blue": 0,
         "ore.green": 0.005,
-        "battery": 0.01,
-        "battery_max": 5,
+        "battery.red": 0.01,
+        "battery.red_max": 5,
```
daphnedemekas, line 22: might be nice to set default rewards for all battery colors (like with ore)

mettagrid/tests/mettagrid_test_args_env_cfg.json
sasmith: Ah, my laziness got the better of me. Will fix up configs as well.

sasmith:
> we could reduce the size of the vector by collapsing agent and converter inventory into the same slots...

Agreed. I'm semi-resistant to doing that now since I could see us needing to spend time debugging, and this will become irrelevant when we switch to tokens. Or, rather, I'd prefer to test this sort of ablation in the world of tokens vs here.

It's also possible that collapsing these is just plan better and needs no debugging.

configs/env/mettagrid/mettagrid.yaml
```diff
@@ -25,8 +25,12 @@ game:
       ore.red: 0.005
       ore.blue: 0.005
       ore.green: 0.005
-      battery: 0.01
-      battery_max: 5
+      battery.red: 0.01
```
sasmith, line 28: In places where there was a clear pattern of rewards, I made colored batteries match plain batteries.

configs/env/mettagrid/mettagrid.yaml
```diff
@@ -73,7 +77,7 @@ game:
   objects:
     altar:
       hp: 30
-      input_battery: 3
+      input_battery.red: 3
```
sasmith, line 80: red is still the default for things "above" generators.

configs/env/mettagrid/mettagrid.yaml
```diff
@@ -120,7 +124,7 @@ game:
     generator.blue:
       hp: 30
       input_ore.blue: 1
-      output_battery: 1
+      output_battery.blue: 1
```
sasmith, line 127: made the generators color match

configs/env/mettagrid/puffer.yaml
```diff
@@ -23,8 +23,8 @@ game:
       ore.red: 0.005
       ore.blue: 0
```
sasmith, line 24: In places where the reward for ore wasn't consistent, I didn't try to extend the pattern for batteries.

mettagrid/tests/mettagrid_test_args_env_cfg.json
sasmith: Agreed, but this is a place where I'm skipping since the ore default rewards are inconsistent. Should the ore all be 0.005?

sasmith:
Smoke test failed since we changed the dimensions. So this is expected.

mettagrid/tests/mettagrid_test_args_env_cfg.json
daphnedemekas: Yeah, i don't think they are inconsistent for any particular reason

PR #607, opened 2025-05-22
_episode_rewards was not being initialized correctly and all indices …

relh:
…in the numpy array shared a view.

Here is what was happening before in metttagrid_env.py:

```
-> obs, infos = self._c_env.reset()
(Pdb) ep_rew
array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.], dtype=float32)
(Pdb) abc[3] = 2
(Pdb) abc
array([0.,  0.,  0.,  2.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0., 15., 15., 15., 15., 15., 15., 15., 15., 15., 15., 15., 15.])
(Pdb) ep_rew[3] = 2
(Pdb) ep_rew
array([2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2.], dtype=float32)
```

Here's what's happening now:

```
-> obs, infos = self._c_env.reset()
(Pdb) ep_rew
array([1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0.], dtype=float32)
(Pdb) ep_rew[3] = 2
(Pdb) ep_rew
array([1., 0., 0., 2., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0.], dtype=float32)
```

Code is a one-line change:

`_episode_rewards = py::array_t<float>(_rewards.shape(0));`

becomes

`_episode_rewards = py::array_t<float>({static_cast<ssize_t>(_rewards.shape(0))}, {sizeof(float)});`

### DIFF
--- a/metta.code-workspace
+++ b/metta.code-workspace
@@ -15,7 +15,8 @@
 		"ruff.enable": true,
 		"ruff.nativeServer": "on",
 		"files.associations": {
-			"*.pxd": "cython"
+			"*.pxd": "cython",
+			"array": "cpp"
 		},
 		"files.excludeGitIgnored": true,
 		"terminal.integrated.profiles.osx": {
--- a/metta.code-workspace
+++ b/metta.code-workspace
@@ -15,8 +15,7 @@
 		"ruff.enable": true,
 		"ruff.nativeServer": "on",
 		"files.associations": {
-			"*.pxd": "cython",
-			"array": "cpp"
+			"*.pxd": "cython"
 		},
 		"files.excludeGitIgnored": true,
 		"terminal.integrated.profiles.osx": {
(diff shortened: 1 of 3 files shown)


daphnedemekas:
Nice find!

rjwalters:
can you add a test that would fail if this patch was not applied?

sasmith:
Approving and sending back to @relh to take a look at the tests and see if there are other things he wants to add.

sasmith (approved):

relh:
Looks great! Never thought we would be checking if 1 + 1 = 2 but here we are.

sasmith:
> Looks great! Never thought we would be checking if 1 + 1 = 2 but here we are.

Seriously. But when you find out that it sometimes doesn't equal 2, you write the test.

PR #621, opened 2025-05-23
Fix S3 save_to_uri bug and add tests

daveey:
## Summary
- fix `save_to_uri` to skip writing a local file when saving to S3
- add tests verifying local and S3 save behaviors

## Testing
- `pytest tests/map/utils/test_storage.py -q`


berekuk (approved):
Oops, right.
I wonder how I missed that, I thought I ran map generation in many different configurations.

PR #632, opened 2025-05-23
Use uv for MettaGrid build

daveey:
## Summary
- delete the old MettaGrid Makefile
- update cleanup logic in the root Makefile
- adjust GitHub workflow and git-history script
- document `uv pip --directory mettagrid install -e .` for building the C++ extension

## Testing
- `pytest -V`


sasmith (changes requested):
Double check tabs vs spaces. My belief is that make files really care about this -- they use tabs for their own directives, and spaces for the scripts that they're supposed to run.

daveey:
i deleted all makefiles. also, doesn't ruff handle that?

PR #635, opened 2025-05-23
Use uv build, drop Makefiles

daveey:
## Summary
- remove Makefiles from repo
- switch cleanup script to manual steps instead of `make clean`
- build the C++ extension using `uv pip` in CI
- run Python tests after activating the virtual environment

## Testing
- `uv pip install -e ./mettagrid`
- `source .venv/bin/activate && pytest -V`
- `source .venv/bin/activate && pytest`


rjwalters:
This is going to break a bunch of stuff. I think we should postpone this conversation until after @berekuk has had some time to think about it.

Some of my concerns:
1. python setup-tools adds flags to the C compilation that we need to manage if we want assertions and errors to show up
2. it is non-trivial to build mettagrid's c++ tests and benchmarks from setup tools; if you remove the makefiles you need to restore this functionality

Overall I understand that Makefiles are annoying and brittle but the system is working in it current state.

PR #636, opened 2025-05-23
Replace Makefile with direct uv pip install

daveey:
### TL;DR

Replace Makefile-based C++ build system with direct Python package installation using `uv pip install -e ./mettagrid`.

### What changed?

- Removed the root `Makefile` and `mettagrid/Makefile` in favor of a simpler build approach
- Updated CI workflows to use `uv pip install -e ./mettagrid` instead of Make commands
- Updated documentation in `README.md` and `CLAUDE.md` with new build instructions
- Simplified test execution to use `pytest` directly instead of Make wrappers
- Removed Makefile from the list of tracked files in CI cache

### How to test?

1. Build the mettagrid extension:
```bash
uv pip install -e ./mettagrid
```

2. Run the test suite:
```bash
uv run pytest
```

3. Verify that CI workflows pass with the updated build commands

### Why make this change?

The Makefile-based build system added unnecessary complexity. Using Python's native package installation mechanism with `uv pip install -e ./mettagrid` provides a simpler, more maintainable approach that works consistently across platforms. This change reduces build complexity while maintaining the same functionality.


daveey:
* **#636** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/636?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/636?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/cpp-checks.yml
```diff
@@ -272,13 +266,12 @@ jobs:
...
-      - name: Build C++ tests
-        working-directory: mettagrid
-        run: make build-tests
+      - name: Build mettagrid extension
+        run: uv pip install --python .venv/bin/python -e ./mettagrid
 
-      - name: Run C++ tests
-        working-directory: mettagrid
-        run: make test
+      - name: Run Python tests
```
sasmith, line 272: We have cpp tests, and I don't think this will run them.

rjwalters (changes requested):
I think we should close this and ask @berekuk to think about it

berekuk (changes requested):
AFAICT this PR and your other 3 PRs on builds break C++ tests and benchmarks in mettagrid. They need to be compiled separately, mettagrid/setup.py can't do it by itself for several reasons:
- they depend on google-benchmark and gtest which need to be installed
- `pip install -e .` builds in isolated environment, so even if we compiled them somehow they won't be available for running

#630 handles all that, at the cost of some cmake complexity (~100 lines of code). So unless we're doing bazel instead, or if I missed some other piece of your plan for those, I think we should proceed with that approach instead of this (#630 is almost ready but needs new github workflows).

PR #648, opened 2025-05-24
feat: allow aws launch from subdirs

rjwalters:
This allows the launch_task script to be run from any directory in the project.

Previously, the `launch_task.py` script had to be run from the repository root directory. If executed from a subdirectory, the script would fail because it couldn't find critical files like:
- `./devops/aws/batch/entrypoint.sh`
- Configuration files in `./configs/`
- Tool scripts in `./tools/`

This created friction for developers who might be working in subdirectories and wanted to quickly launch AWS Batch jobs without manually navigating to the root.

- Fully backward compatible - existing workflows unchanged
- New `--no-auto-cd` flag available for scripts that need the old behavior
- No breaking changes to existing functionality

### DIFF
--- a/devops/aws/batch/launch_task.py
+++ b/devops/aws/batch/launch_task.py
@@ -8,6 +8,7 @@ import sys
 import boto3
 
 from metta.util.colorama import blue, bold, cyan, green, red, use_colors, yellow
+from metta.util.fs import ensure_metta_repo_root
 from metta.util.git import (
     get_branch_commit,
     get_commit_message,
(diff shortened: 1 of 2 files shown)


daveey (changes requested):
this is icky, and launch_task is going away. i don't want to add 200 lines of code just so you can launch scripts from random spots

berekuk:
+1 to David that this is bloated.

But actually I won't mind this PR if `metta.util.fs` was 10x shorter, single function that climbs up to `.git`, no logs.

I often `cd mettagrid` and switching dirs can be annoying.

---

Another idea: most of our python scripts have common code, `setup_metta_environment`, `setup_mettagrid_environment`, etc. We could have a decorator, `@metta_script`, that does all these.

daveey:
much more reasonable. fwiw, launch_task is getting deleted soon, since we are now mostly on skypilot.

daveey (approved):

PR #650, opened 2025-05-24
hotfix: fail gracefully when policies don't match env

rjwalters:
this will allow training runs to finish so that we can create new policies to use in evals

image


metta/sim/simulation_suite.py
```diff
@@ -39,27 +39,44 @@ def __init__(
...
+                merged_db.merge_in(sim_result.stats_db)
+                sim_result.stats_db.close()
+                successful_simulations += 1
+
+            except SimulationCompatibilityError as e:
+                logger.warning("Skipping simulation '%s' due to compatibility issue: %s", name, str(e))
+                continue
+            except Exception as e:
+                logger.error("Unexpected error in simulation '%s': %s", name, str(e))
+                # You can decide whether to continue or re-raise based on your needs
```
sasmith, line 75: Seems like a spurious comment.

metta/sim/simulation_suite.py
```diff
@@ -39,27 +39,44 @@ def __init__(
...
+                logger.info("=== Simulation '%s' ===", name)
+                sim_result = sim.simulate()
+                merged_db.merge_in(sim_result.stats_db)
+                sim_result.stats_db.close()
+                successful_simulations += 1
+
+            except SimulationCompatibilityError as e:
+                logger.warning("Skipping simulation '%s' due to compatibility issue: %s", name, str(e))
+                continue
+            except Exception as e:
```
sasmith, line 73: Let's not do the blanket catch. I'd rather have unexpected failures bubble to the surface.

metta/sim/simulation_suite.py
```diff
@@ -39,27 +39,44 @@ def __init__(
...
+                    replay_dir=self._replay_dir,
+                )
+                logger.info("=== Simulation '%s' ===", name)
+                sim_result = sim.simulate()
+                merged_db.merge_in(sim_result.stats_db)
+                sim_result.stats_db.close()
+                successful_simulations += 1
+
+            except SimulationCompatibilityError as e:
+                logger.warning("Skipping simulation '%s' due to compatibility issue: %s", name, str(e))
```
sasmith, line 71: Can we narrow this to only give a pass when the NPC is incompatible?

sasmith (approved):
New version makes sense, since we expect NPC policies to be bad when we're training new replacement policies. But please narrow the exceptions that we're turning into warnings.

PR #653, opened 2025-05-24
Remove grid layers

sasmith:
I wanted to make GridLayer an `enum class`, but instead decided we don't need layers in the grid. My best guess is that they came from us not being able to use dynamic_cast (or equivalent) in Cython. Let me know if there's something I'm missing. Otherwise here's the auto-gen commit message:

This PR removes the concept of layers from the grid system, simplifying the codebase by having a single object per grid cell. Key changes:

- Remove `layer` field from `GridLocation` class
- Update `Grid` class to use a 2D grid instead of 3D
- Remove `layer_for_type_id` and related layer mapping
- Update all action handlers to work with the simplified grid structure
- Fix object targeting in attack actions to use dynamic_cast for type checking
- Update benchmarks and tests to work with the new grid structure

This change improves code clarity by removing the need to manage multiple layers at each grid position, while maintaining all existing functionality through proper object type checking.

### DIFF
--- a/mettagrid/benchmarks/grid_object_benchmark.cpp
+++ b/mettagrid/benchmarks/grid_object_benchmark.cpp
@@ -10,8 +10,8 @@ class TestGridObject : public GridObject {
 public:
   TestGridObject() = default;
 
-  TestGridObject(TypeId type_id, GridCoord r, GridCoord c, Layer layer) {
-    init(type_id, r, c, layer);
+  TestGridObject(TypeId type_id, GridCoord r, GridCoord c) {
+    init(type_id, r, c);
   }
 
   virtual vector<PartialObservationToken> obs_features() const override {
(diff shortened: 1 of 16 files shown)


sasmith:
* **#653** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/653?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/653?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

daveey:
The intention for grid layers is to support agents walking onto objects, and possibly stacking objects. This way you could make a "road" or "swamp", or a "tower". 

We don't currently use this, so maybe the answer is "simplify now add back later"?

daveey (changes requested):
Let's discuss, I am not sure which way to go here. This does reduce / simplify code, and layers have previously been a source of bugs. OTOH, this does seem like a nice feature of the env, and I'd be sad to lose it. Bringing it back would run into the same bugs that have already been fixed.

Could go either way.

sasmith:
Don't feel a strong need to push this through, at least for now. This was oddly driven by wanting to switch to `enum class`, which wasn't likely to go all the way anyways.

PR #660, opened 2025-05-25
Add git-pr script

daveey:
## Summary
- add `git-pr.sh` script to run lint, tests and `gt submit`
- rebuild both packages and run MettaGrid tests

## Testing
- `uv run ruff format --diff`
- `uv run ruff check`
- `uv run pytest`
- `uv run python -m pytest mettagrid/tests`

### DIFF
--- a/git-pr.sh
+++ b/git-pr.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+set -e
+
+# Ensure not on main branch
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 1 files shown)


sasmith (approved):
Comment from Diamond seems pretty reasonable.

Consider doing building / tests and `gt submit` in parallel, if there's a good way to do that.

I don't expect everyone is using graphite. This will push people towards that. This seems fine to me, although you've managed to foot-gun a few times.

PR #662, opened 2025-05-25
Add git-pr helper

daveey:
## Summary
- rename `git-pr.sh` to `git-pr`
- reinstall both packages before running tests
- test Metta and MettaGrid

## Testing
- `uv run ruff format --diff`
- `uv run ruff check`
- `uv run pytest`
- `uv run python -m pytest mettagrid/tests`

### DIFF
--- a/git-pr
+++ b/git-pr
@@ -1,0 +1,35 @@
+#!/bin/bash
+set -e
+
+# Ensure not on main branch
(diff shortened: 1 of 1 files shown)


sasmith (approved):
I feel like this should be in a scripts directory or something.

PR #673, opened 2025-05-26
Disable smoke test

daveey:
## Summary
- disable the smoke test job in CI
- remove smoke-test dependency from benchmark job
- ignore missing artifacts when downloading benchmark files

## Testing
- `uv run pytest -q`


daveey:
Merging this without CR

PR #692, opened 2025-05-27
Validate args in env init

relh:
## Summary
- validate MettaGridEnv constructor inputs
- test MettaGridEnv type validations

## Testing
- `ruff format mettagrid/mettagrid/mettagrid_env.py tests/mettagrid/test_env_input_validation.py`
- `ruff check mettagrid/mettagrid/mettagrid_env.py tests/mettagrid/test_env_input_validation.py`
- `pytest tests/mettagrid/test_env_input_validation.py -q` *(fails: ModuleNotFoundError: No module named 'omegaconf')*

### DIFF
--- a/mettagrid/mettagrid/mettagrid_env.py
+++ b/mettagrid/mettagrid/mettagrid_env.py
@@ -28,6 +28,16 @@ class MettaGridEnv(pufferlib.PufferEnv, gym.Env):
         replay_writer: Optional[ReplayWriter] = None,
         **kwargs,
     ):
+        if not isinstance(env_cfg, DictConfig):
+            raise TypeError(f"env_cfg must be an OmegaConf DictConfig, got {type(env_cfg)}")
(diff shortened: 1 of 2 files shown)


mettagrid/mettagrid/mettagrid_env.py
```diff
@@ -28,6 +28,16 @@ def __init__(
         replay_writer: Optional[ReplayWriter] = None,
         **kwargs,
     ):
+        if not isinstance(env_cfg, DictConfig):
```
sasmith, line 31: I've dragged my feet a bit since @daveey is adding a decorator that should make this easier / more concise.

relh: Yup yup, I'm closing this now. It's now handled and I checked that it's actually working by running the piece of PufferLib that used to break this

PR #695, opened 2025-05-27
Add ability to hide mini-map, controls, and info panel.

treeform:
* Add ability to hide mini-map, controls, and info panel.
* Use Shift-Z to open up attack visual even when controls are hidden.
* Help link back to readme.
* Fix a bug with mini-map not appearing with WebSocket.

image


treeform:
> The comment // Just move forward. followed by sendAction("move", 0) adds no information beyond what's already clear from the code.

Robot, thats not true. Its not clear if sendAction("move", 0) or sendAction("move", 1) is move forward. Without the comment you don't know. The comments are helpful, and // Noop is there for symmetry. It looks ugly without a comment.

daveey (approved):

PR #701, opened 2025-05-27
Add runtime arg type validation

daveey:
## Summary
- validate arg types at runtime with `validate_arg_types` decorator
- export decorator from `metta.util`
- test the decorator

## Testing
- `ruff check metta/util/validate.py tests/util/test_validate_arg_types.py metta/util/__init__.py`
- `uv run pytest -q`


sasmith (approved):
Looks great

PR #707, opened 2025-05-27
Update observation_space to work for observation_tokens

sasmith:
This exposes a different observation shape if we're using tokens, so puffer will send us the right shaped observations.

Testing:

AlexV's been developing against this for a while.


sasmith:
### Merge activity

* **May 27, 11:33 PM UTC**: @sasmith merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Metta-AI/metta/707).

PR #719, opened 2025-05-28
mapgen tests

berekuk:
This PR adds tests for all existing mapgen scenes, to make them more robust to upcoming refactorings.

It also brings back `tools.map.view` script fixes `tools.map.gen` which were broken after the curriculum PR.


metta/map/utils/ascii_grid.py
```diff
@@ -0,0 +1,102 @@
+import numpy as np
+
+from metta.map.types import MapGrid
+
+ascii_symbols = {
```
sasmith, line 5: As a follow up, do you want to see if you can reduce the number of places we store these makes and reverse maps down to one?

Also, (and this is scope creep from the current diff), why do we need multiple ways to show (for example) walls?

sasmith (approved):

PR #728, opened 2025-05-29
Implement ascii renderer

relh:
## Summary
- add simple ASCII renderer for MettaGrid
- instantiate renderer when render mode is used
- update tests for renderer initialization

## Testing
- `uv run ruff format mettagrid/tests/test_basic.py mettagrid/mettagrid/renderer/renderer.py mettagrid/mettagrid/mettagrid_env.py`
- `uv run ruff check mettagrid/tests/test_basic.py mettagrid/mettagrid/renderer/renderer.py mettagrid/mettagrid/mettagrid_env.py`
- `uv run pytest -q`

### DIFF
--- a/mettagrid/mettagrid/mettagrid_env.py
+++ b/mettagrid/mettagrid/mettagrid_env.py
@@ -58,6 +58,10 @@ class MettaGridEnv(pufferlib.PufferEnv, gym.Env):
         self._should_reset = False
 
         self._reset_env()
+        if self._render_mode is not None:
+            from .renderer.renderer import AsciiRenderer
+
+            self._renderer = AsciiRenderer(self.object_type_names)
         super().__init__(buf)
 
     def _make_episode_id(self):
@@ -99,6 +103,11 @@ class MettaGridEnv(pufferlib.PufferEnv, gym.Env):
     def reset(self, seed: int | None = None, options: dict | None = None) -> tuple[np.ndarray, dict]:
         self._env_cfg = self._get_new_env_cfg()
         self._reset_env()
+
+        if self._render_mode is not None:
+            from .renderer.renderer import AsciiRenderer
+
+            self._renderer = AsciiRenderer(self.object_type_names)
 
         self._c_env.set_buffers(self.observations, self.terminals, self.truncations, self.rewards)
 
--- a/mettagrid/mettagrid/mettagrid_env.py
+++ b/mettagrid/mettagrid/mettagrid_env.py
@@ -9,9 +9,9 @@ import gymnasium as gym
 import numpy as np
 import pufferlib
 from omegaconf import DictConfig, OmegaConf
+from pufferlib.utils import unroll_nested_dict
 from typing_extensions import override
 
-from metta.util.datastruct import flatten_config
 from mettagrid.level_builder import Level
 from mettagrid.mettagrid_c import MettaGrid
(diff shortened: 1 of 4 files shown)


sasmith:
How does this fit in with the ASCII renderer that we removed a few weeks ago?

mettagrid/mettagrid/mettagrid_env.py
```diff
@@ -58,7 +58,12 @@ def __init__(
         self._should_reset = False
 
         self._reset_env()
+
         super().__init__(buf)
+        if self._render_mode is not None:
```
sasmith, line 63: I guess this gets set by joseph's configs, and I assume `_render_mode` isn't something we can rename. Try adding a comment to explain this -- e.g., why are we treating this non-boolean parameter as a boolean.

sasmith: It smells funny for us to need to set the renderer here and in `reset`. Try to make us only need to set this in one place.

mettagrid/mettagrid/renderer/renderer.py
```diff
@@ -0,0 +1,107 @@
+from __future__ import annotations
+
+import shutil
+from typing import Dict, List
+
+
+class AsciiRenderer:
+    """Simple ASCII renderer for ``MettaGridEnv``."""
+
+    SYMBOLS = {
```
sasmith, line 10: We define versions of these in at least 4 other places (with the UTF8 characters). I don't want to scope-creep you wrangle all of these into one place
(a) consider how hard it would be to have a single config file that defines these (like maybe a yaml file that gives an ascii and UTF8 symbol for each of these objects)
(b) if not the above, add a comment to link the various places, and note what we probably want to do as a future step

mettagrid/mettagrid/renderer/renderer.py
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    def __init__(self, object_type_names: List[str]):
+        self._object_type_names = object_type_names
+        self._bounds_set = False
+        self._min_row = 0
+        self._min_col = 0
+        self._height = 0
+        self._width = 0
+        self._terminal_height = 0
```
sasmith, line 32: terminal_height and ...width aren't used here. Based on this, I assume that either you used an LLM that added garbage, or we're conforming to a puffer API. If it's the former, please fix. If the latter, please reference / comment on the fact.

sasmith (changes requested):

mettagrid/mettagrid/renderer/renderer.py
relh: don't worry about scope creep I will go hunt for the pre-existing characters. I had been using Emojis before but they weren't all the right width.

relh:
I'm going to leave the name as AsciiRenderer despite using emojis because we are getting the symbols from a file called ascii.py. 

Alternatively, we could change to actual ascii, which helps because the emojis widths are different.

I don't think this is worth it because I think what we really want is to move towards render() being more than just a debugging tool and be instead something that can hook into MettaScope so anyone who uses the environment can see @treeform 's pretty graphics! 

They won't be able to see the graphics in the terminal, but the idea could be that if you have X11 windows or w/e we launch mettascope by default as a GUI.

But for now this change is just intended to give Joseph a way to run the environment and see why his agents are performing well/badly.

sasmith (approved):

PR #736, opened 2025-05-29
feat: simplified StatsTracker for MettaGrid

rjwalters:
This PR refactors the `StatsTracker` class to provide a more flexible and feature-rich statistics tracking system for the MettaGrid environment.

### Key Changes

1. **Flexible stat types**: Stats can now be either integers or floats using `std::variant<int, float>`. The system automatically handles type conversions when needed.

2. **Automatic metadata tracking**: When an environment is set, the tracker automatically records:
   - First and last step when a stat was updated
   - Min/max values over time
   - Number of updates
   - Update rate (updates per simulation step)
   - Average value (total / number of updates)

3. **Simplified API**: Removed the overloaded methods that concatenated strings. Stats are now tracked with explicit key names like `"action.move.success"` instead of multiple string parameters.

4. **Added StatsTracker to Converter class**: Converters now have their own `StatsTracker` instance, enabling detailed tracking of:
   - Conversions started/completed
   - Items produced/consumed
   - Blocking reasons (insufficient input, output full)
   - Cooldown cycles
   - Inventory changes

5. **Enhanced stat tracking throughout the codebase**:
   - More granular tracking in attack actions (differentiating between own team/other team attacks)
   - Better inventory tracking with `.gained`, `.lost`, `.stolen` suffixes
   - Converter stats are now included in episode stats output

6. **Per-object stats**: Both agents and converters now have their own `StatsTracker` instances, allowing for detailed per-object metrics that are aggregated in `get_episode_stats()`.

### Breaking Changes

- Removed overloaded `incr()`, `add()`, and `set_once()` methods that took multiple string parameters
- Stats are now returned as `float` values in `to_dict()` for Python compatibility
- The `stats()` method has been replaced with `to_dict()`

### Testing

- Comprehensive unit tests added for the new `StatsTracker` functionality
- Tests cover basic operations, type handling, metadata tracking, and edge cases
- Tests gracefully handle the absence of an environment (for unit testing)


mettagrid/tests/test_stats_tracker.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+// Test increment on float should throw
+TEST_F(StatsTrackerTest, IncrementFloatThrows) {
+  stats.set("float_stat", 1.5f);
+  EXPECT_THROW(stats.incr("float_stat"), std::runtime_error);
+}
+
+// Test timing tracking with environment
+TEST_F(StatsTrackerTest, TimingWithEnvironment) {
+  // We can't easily test timing with a mock environment since
```
sasmith, line 102: The name vs this comment and the body of the test are confusing. I think this is just `NoEnvironmentNoTiming`?

mettagrid/mettagrid/stats_tracker.cpp
sasmith, line 1: Is there a reason to have this file?

sasmith (commented):
The config changes look good, if you want to split those off.

For the stats stuff, you're changing functionality at the same time you're refactoring the stats code, which makes some parts harder to follow. But also, a lot of this feels like it's moving pretty opposite to 8d97abdc76a1a36e59616e4943725c1368588035. That PR doesn't give a good indication as to why we want stats passed in parts and combined within library code, but I assume it's to avoid depending on callers to format stats string consistently. I'll give this to @daveey for comment.

mettagrid/mettagrid/mettagrid_c.cpp
```diff
@@ -54,8 +54,9 @@ MettaGrid::MettaGrid(py::dict env_cfg, py::list map) {
 
   _event_manager = std::make_unique<EventManager>();
   _stats = std::make_unique<StatsTracker>();
+  _stats->set_environment(this);
```
daveey, line 57: we don't have to do this now, but i think it would be nice if GridObjects took environment in the constructor, rather than having us set it later. that would let us require that StatsTracker() also take env, and avoid potential bugs where we forget to set it later.

daveey (approved):

PR #740, opened 2025-05-29
Preview any env or scene through mapgen and mapgen-scene scripts

berekuk:
Added two new command-line tools for map generation: `mapgen` and `mapgen-scene`.

`mapgen` script is an alias for `python -m tools.map.gen`; `mapgen-scene` is new.

`mapgen` script now supports any environment file, not just mapgen-based maps.

After this PR, I expect `mapgen` (script, not map class system) to become the most convenient way to preview env maps and finally get some adoption.

Upd: I've just realized that it doesn't work for maps that can'be expressed in ascii, so it fails on `simple.yaml` env because it has `block` object types. But it does work for almost all other configs. I'll address the ascii problem in another PR.

### How to test?

1. Generate a map from an environment config:
   ```
   mapgen configs/env/mettagrid/school.yaml
   ```

2. Generate a map from a scene config with specific dimensions:
   ```
   mapgen-scene scenes/test/maze.yaml 20 15
   ```

### DIFF
--- a/tools/map/gen.py
+++ b/tools/map/gen.py
@@ -13,6 +13,7 @@ from omegaconf import DictConfig, OmegaConf
 
 from metta.map.utils.show import ShowMode, show_map
 from metta.map.utils.storable_map import StorableMap
+from metta.util.config import config_from_path
 from metta.util.resolvers import register_resolvers
 
 # Aggressively exit on ctrl+c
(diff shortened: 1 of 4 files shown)


tools/map/gen.py
```diff
@@ -13,6 +13,7 @@
 
 from metta.map.utils.show import ShowMode, show_map
 from metta.map.utils.storable_map import StorableMap
+from metta.util.config import config_from_path
```
daveey, line 16: this is now in mettagrid.util.hydra

daveey (approved):

tools/map/gen.py
berekuk: from what I see right now it's copy-pasted and some code on main still uses the old version, so I'll leave that cleanup for a separate PR

PR #743, opened 2025-05-30
spotted a potential bug in lists of nodes

relh:
The Bug:
The select_areas method in node.py was supposed to always return a regular Python list. But when you asked it to randomly pick some areas (like "give me 2 random areas"), it was secretly returning a NumPy array instead.

Why it mattered:
The method promised to return a list (that's what the type hint said)
NumPy arrays look similar to lists but they're different - you can't use methods like .append() on them
This inconsistency could break code that expected a real list

The Fix:
Super simple - we just wrapped the NumPy result with list() to convert it back to a regular Python list:

So basically, NumPy was being helpful by giving us a way to randomly select items, but it was giving us back its own special array type. We just needed to convert it back to what Python normally uses - a regular list. Now the method always returns the same type no matter how you ask for the areas (random, first, or last).

### DIFF
--- a/metta/map/node.py
+++ b/metta/map/node.py
@@ -93,7 +93,7 @@ class Node:
             if order_by == "random":
                 assert offset is None, "offset is not supported for random order"
                 rng = np.random.default_rng(query.get("order_by_seed"))
-                selected_areas = rng.choice(selected_areas, size=int(limit), replace=False)  # type: ignore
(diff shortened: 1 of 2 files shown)


berekuk (approved):
lgtm, thanks

(don't forget to fix the lint before merge)

rjwalters (approved):

PR #745, opened 2025-05-30
Add checks for Skypilot git refs

sasmith:
## Summary
- check that git ref is pushed before launching a Skypilot run
- fail Skypilot setup steps on first error

## Testing
- `ruff check devops/skypilot/launch.py`
- `uv run pytest -q` *(fails: KeyboardInterrupt but shows 49 passed)*


sasmith:
Actually this doesn't work yet. `git ls-remote origin REF` only does the expected thing for named branches, not for hashes.

berekuk:
Oops, I approved and only then noticed the comment.

PR #746, opened 2025-05-30
Unify map.node and map.scene; pydantic in mapgen

berekuk:
`metta.map.Node` doesn't exist anymore; individual nodes are instances of `Scene` class. This should clean up some confusion, and generally makes Scene APIs nicer.

Removed `recursive_map_builder`. Scene classes are now specified with `type: ...` instead of `_target_`, so Hydra can't instantiate them.

Pydantic everywhere; all scene classes provide a generic `Config` parameter that's used for validation. This required a breaking change in yaml: scene params are now wrapped under `params:` field. This makes scene configs slightly more verbose, but one potential win here is that all other scene properties (seeds, children) are now standardized, which will be helpful in the future.

Because of the breaking change, the diff is quite long, sorry. I had to reimport all DCSS maps and update all env configs.

The main part to review here is the new `metta.map.scene` API.


berekuk:
> Can you, e.g., do the pydantic changes separately; and then figure out how to do the bulk yaml changes in a PR with as minimal of a Python change as possible?

This would make me redo the entire PR from scratch (and take me longer than I did on this PR) because all these changes are related:
- `params` was necessary for pydantic to work, and it forced me to update yaml
- without node/scene refactoring pydantic would be harder to do; and vice versa

I could extract some other stuff like `ChildrenAction` pydantic type, but it won't help much...

Maybe you could review individual commits on this? https://github.com/Metta-AI/metta/pull/746/commits
They're decently organized, but I can't send them separately because it would result in broken tests or configs.

sasmith:
Okay. I may be able to take a look Sunday evening or Monday.

PR #753, opened 2025-05-31
Add update epoch scaling

dominik3141:
Scale the number of update_epochs with the world size. 

As described here: [wandb writeup](https://api.wandb.ai/links/metta-research/ct31kzhg)


dominik3141:
> We talked about adding this in train.sh instead, otherwise we won't be able to override this via cmd args

Then it will not take effect if people use `train.py` instead of `train.sh`

daveey:
Yes, but when we run training jobs on the cluster, it always runs train.sh

train.py bypasses torchrun so it doesn't work with multi-gpu anyway

Al-does (approved):

daveey (approved):

PR #776, opened 2025-06-02
George navseq merge

georgejwdeane:
There's now a folder entitled 'experiments' under navigation_sequence, and multiagent. This enables anyone to experiment with the multiagent and navigation_sequence environments. there are two kinds. Explanations for the different environments noted in the defaults for each


daphnedemekas (changes requested):
Can you add a curriculum file for multiagent and navsequence that point to these experiment files

PR #779, opened 2025-06-03
feat: improved training time metrics

rjwalters:
This PR adds detailed timing instrumentation to the PufferLib trainer to better understand performance bottlenecks and optimize training efficiency.

We are keeping all the old Pufferlib metrics for now -- this just adds another set of cleaned up metrics.

### New Timing Instrumentation
- Added `Stopwatch` utility class in `metta/util/timing.py` that provides:
  - Named timers with context manager support
  - Nested timing capabilities
  - Progress tracking and rate calculations
  - Human-readable time formatting

### Enhanced W&B Metrics
- Added comprehensive timing metrics to W&B logging:
  - Steps per second based on actual training time (excluding overhead)
  - Percentage breakdown of where time is spent
  - Wall clock time tracking
- Updated training time calculation to only include rollout + train time (excluding evaluation/checkpointing overhead)
- Improved epoch logging to show timing breakdown per operation

## Testing
image

image


metta/rl/pufferlib/trainer.py
```diff
@@ -714,7 +723,20 @@ def _process_stats(self):
 
         environment = {f"env_{k.split('/')[0]}/{'/'.join(k.split('/')[1:])}": v for k, v in self.stats.items()}
 
+        # Add timing metrics to wandb
         if self.wandb_run and self._master:
+            rollout_time = self.timer.get_elapsed("_rollout")
```
daveey, line 728: Could this be a loop over all the timers in self.timer() instead?

metta/util/timing.py
```diff
@@ -0,0 +1,203 @@
+import logging
+import time
+from contextlib import contextmanager
+from typing import Any, ContextManager, Dict, Optional, Tuple
+
+
+class Stopwatch:
```
daveey, line 7: @relh can this be consolidated with your profiler code?

daveey (commented):
Added my comments, sending to @relh to approve/request changes

relh (approved):

daveey:
Sounds good! I support removing the use of the other profile either here or as a quick-followup diff.

rjwalters:
> Sounds good! I support removing the use of the other profile either here or as a quick-followup diff.

I will open a folllow up

PR #784, opened 2025-06-03
Bring mettamap to monorepo; preview mettagrid configs with it

berekuk:
## How to run

```
# in one terminal:
python -m metta.map.server

# in another:
cd mettamap
pnpm install
pnpm dev
```

## History

This code is based on what I had previously in mettamap repo's [player](https://github.com/Metta-AI/mettamap/tree/player) branch. That branch had an experimental frontend for `tools.play` that I did a month ago, but now MettaScope has this functionality so I removed that code. But the map viewer functionality (with pan & zoom) remains.

I've also added some new features.

## Configs viewer

Non-obvious feature: config name and class names can be clicked and they're immediately opened in local Cursor editor.

## Viewing stored maps from S3

We're not currently using the ability to store mapgen maps, I'm not sure if we will, but I kept the code and switched it from local files to S3, as we discussed before.

## Other notes

This is still unpolished and I know there's a lot that could be cleaned up, but this PR is pretty big as it is, so I want to send it in early.

Some especially unpolished things:
1. `metta.map.server` could be merged with `mettascope.server`
2. Code is separate from the observatory
3. "Stored Maps" is very slow (because of S3)
4. "Stored Maps" part hardcodes a single map collection on the backend, `test-collection`
5. Pan&zoom cell highlighting is a bit broken
6. Maps with blocks don't render (not an issue with this viewer, though, it's a mapgen StorableMap limitation that I'll fix separately)
7. Maps use old style sprites.

I hope to fix these soon; (1) and (2) require coordination with Pasha and Andre; (3)-(7) could be improved but I had to stop and submit somewhere. I can continue to fix in this PR or in follow-up PRs.


metta/map/server.py
```diff
@@ -0,0 +1,121 @@
...
+    async def route_stored_maps_dirs():
+        return {
+            "dirs": [
+                # TODO - list all dirs in s3://softmax-public/maps/
+                "s3://softmax-public/maps/test-collection",
+            ]
+        }
+
+    @app.get("/stored-maps/find-maps")
+    async def route_stored_maps_find_maps(dir: str, filter: str):
```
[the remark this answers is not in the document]
berekuk: Do you mean `dict` or fully-typed `TypedDict` shapes?

metta/map/utils/storable_map.py
```diff
@@ -58,3 +64,81 @@ def from_uri(uri: str) -> "StorableMap":
...
+        metadata={
+            "gen_time": gen_time,
+            "timestamp": datetime.now().isoformat(),
+        },
+        config=cfg,
+    )
+    return storable_map
+
+
+class StorableMapIndex:
```
Barraketh, line 98: add a comment about what this class does

mettamap/src/lib/api.ts
```diff
@@ -0,0 +1,146 @@
+import { z } from "zod";
```
Barraketh, line 1: We probably don't need to use zod to validate our maps, since our maps are generated by our own server and can be trusted. But that can be a cleanup for a later time

Barraketh (approved):
lgtm

mettamap/src/lib/api.ts
berekuk: I prefer to validate always as an assertion that the shapes haven't got out of sync. Doesn't cost much (typescript type gets inferred from the schema) and makes it much easier to catch bugs.

PR #794, opened 2025-06-03
Move normalization values to the mettagrid environment

sasmith:
This moves the source of truth for observation normalization values from the ObservationNormalizer to the cpp environment. This sets us up to have the normalization set in configuration.

We continue passing grid_features for now, although they're unused. We should talk about whether they're useful to ensure that the features remain stable (e.g., the order remains stable). This is a similar problem we have for tokens -- making sure the feature_ids remain appropriately stable.

Normalization values are also only passed on layer creation -- not when we save and restore. That will need to change if we want the agent to be able to deal with normalization values updating between environments.

### DIFF
--- a/configs/user/sasmith.yaml
+++ b/configs/user/sasmith.yaml
@@ -4,5 +4,5 @@ defaults:
   - _self_
 
 seed: null
-run_id: 20250524.1
+run_id: 20250603.3
 run: ${oc.env:USER}.local.${run_id}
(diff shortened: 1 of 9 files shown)


sasmith:
* **#794** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/794?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/794?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #795, opened 2025-06-03
Add env overrides to simulation suites

daphnedemekas:
We want to be able to have global level overrides to our simulation suite so that we can train and evaluate on the robust cross architecture (which requires an environment override)

### DIFF
--- a/metta/sim/simulation_suite.py
+++ b/metta/sim/simulation_suite.py
@@ -45,6 +45,8 @@ class SimulationSuite:
 
         for name, sim_config in self._config.simulations.items():
             try:
+                # merge global simulation suite overrides with simulation-specific overrides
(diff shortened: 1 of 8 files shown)


metta/sim/simulation_suite.py
```diff
@@ -45,6 +45,8 @@ def simulate(self) -> SimulationResults:
 
         for name, sim_config in self._config.simulations.items():
             try:
+                # merge global simulation suite overrides with simulation-specific overrides
```
daveey, line 48: i don't think so, but i see where you're coming from

daveey (approved):

metta/sim/simulation_suite.py
daphnedemekas: you don't think what?

PR #796, opened 2025-06-03
An improved nethack style ascii renderer

relh:
Also added a new tool for checking the renderer under tools/renderer

Also updated our README to be a bit clearer


treeform (approved):
I ran this and it works.

PR #798, opened 2025-06-03
alexv updated token obs normalizer

Al-does:


metta/agent/lib/obs_enc.py
```diff
@@ -21,16 +22,31 @@ def __init__(
...
 
     def _make_net(self) -> None:
         self._out_tensor_shape = [self.M, self._feat_dim]
 
-        self._atr_embeds = nn.Embedding(257, self._atr_embed_dim, padding_idx=0)
+        self._atr_embeds = nn.Embedding(self._max_embeds, self._atr_embed_dim, padding_idx=0)
         nn.init.uniform_(self._atr_embeds.weight, -0.1, 0.1)
 
-        self._coord_embeds = nn.Embedding(256, self._atr_embed_dim)
+        # Coord byte supports up to 16x16, so 256 possible coord values
```
sasmith, line 34: I think we'll want to use some of the coordinates to mean "self" and "global".

Al-does: The network can likely learn that on its own as long as the env reuses the same coord location to mean "self" and another for "global". I'd even expect it to figure it out if we use the same coord location for both self and global. That would be to say that maybe mapping all of those to 0,0 might be the way to go. In any case, this wouldn't require a code change here in the obs_enc as it can already handle that.

sasmith (approved):

PR #799, opened 2025-06-03
Add `--clear-uv-cache` option to `setup_build.sh` for clearing `uv` cache directory

mrazomej:
### What changed?
- Introduced a new command line argument `--clear-uv-cache` to the `setup_build.sh` script.
- Added functionality to delete the `uv` cache directory if it exists, with a confirmation prompt before proceeding.

### Why make this change?
This enhancement enables users to easily clear the `uv` cache, ensuring that cached packages are removed and forcing re-downloads on subsequent use, which can help troubleshoot and maintain a clean environment. Running this option might be necessary because the `--clean` flag still uses some cached files stored in the `~/.cache/uv` directory that can interfere with changes in mettagrid (as I experienced firsthand).

### DIFF
--- a/devops/setup_build.sh
+++ b/devops/setup_build.sh
@@ -12,10 +12,15 @@ cd "$PROJECT_DIR"
 
 # Parse command line arguments
 CLEAN=0
+CLEAR_UV_CACHE=0
 for arg in "$@"; do
   case $arg in
     --clean)
       CLEAN=1
+      shift
+      ;;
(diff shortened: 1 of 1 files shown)


daveey (approved):

berekuk:
I believe this is not needed after #805, please close unless you can reproduce cache issues on latest main.
(see the long Discord discussion last night for details)

PR #801, opened 2025-06-03
Tensor typing

relh:
This PR would add explicit typing to all the torch.tensor() calls we make throughout our code. 

I'm not sure if we want something like this, but I'm making it just to highlight WHERE all our torch.tensor calls actually are.

### DIFF
--- a/devops/lint_tensor_dtype.py
+++ b/devops/lint_tensor_dtype.py
@@ -1,0 +1,151 @@
+#!/usr/bin/env python3
+"""
+Custom linter to check for torch.tensor() calls without explicit dtype parameter.
+
+This script scans Python files and identifies torch.tensor() calls that don't
+specify a dtype parameter, which can lead to inconsistent behavior across
+different PyTorch versions and platforms.
+
+Usage:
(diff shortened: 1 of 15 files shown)


devops/lint_tensor_dtype.py
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env python3
```
daveey, line 1: let's just put it into tests/test_tensors_have_dtype.py

daveey (changes requested):
thanks for doing this! i love all the fixes this found too. 

we can think of this more as a unittest rather than a tool. so let's just put it there. consider rewriting as a pytest test, should mostly be the same code

relh:
> thanks for doing this! i love all the fixes this found too.
> 
> we can think of this more as a unittest rather than a tool. so let's just put it there. consider rewriting as a pytest test, should mostly be the same code

perfect, I'm on it!

rjwalters (approved):
I like this idea -- do we want to set up type aliases somewhere so that it is easier to change these values? (I think we are currently using torch.int64 in some places)

PR #805, opened 2025-06-04
cp

daveey:


mettagrid/pyproject.toml
```diff
@@ -19,6 +19,142 @@ keywords = [
     "environment",
     "gym",
 ]
+dependencies = [
+    "aiofiles==24.1.0",
```
daveey, line 23: we should clean these up, but for now, just copied from metta/requirements.txt

berekuk (approved):

PR #811, opened 2025-06-04
fix: add types.hpp to mettagrid c env

rjwalters:
This PR introduces proper type handling throughout the mettagrid codebase to ensure compatibility with PufferLib's expected data types and improve type safety across the Python/C++ boundary.

### Core Type System (`mettagrid/types.hpp`)
- Added centralized type definitions for all numpy arrays used in the environment
- Defined C++ types that match PufferLib's expectations:
  - `ObservationType` (uint8)
  - `ActionType` (int32) 
  - `RewardType` (float32)
  - `TerminalType`, `TruncationType`, `MaskType`, `SuccessType` (bool)
- Created Python dtype helper functions to ensure consistency

### Environment Updates (`mettagrid_env.py`)
- Replaced hardcoded numpy types with centralized dtype definitions
- Added explicit type conversions in `reset()` to ensure buffers have correct dtypes
- Improved type annotations for properties and methods
- Added proper dtype specification for observation and action spaces

### C++ Integration (`mettagrid_c.cpp`)
- Updated observation and action spaces to use the new dtype system
- Added validation for agent IDs to prevent out-of-bounds access
- Improved error handling with descriptive error messages
- Used typed min/max values for observation space bounds

### Training and Simulation
- Updated `trainer.py` to use the correct action dtype
- Modified `simulation.py` to properly cast actions to expected type
- Ensured numpy arrays are converted to the correct dtype before passing to C++

## Motivation

1. **PufferLib Compatibility**: PufferLib expects specific numpy dtypes for its vectorized environments. This change ensures our types match exactly.

2. **Type Safety**: By centralizing type definitions, we reduce the risk of type mismatches between Python and C++ code.

3. **Debugging**: The added validation and error messages make it easier to catch type-related issues early.

4. **Maintainability**: Having a single source of truth for types makes future changes easier.


relh (approved):

mettagrid/mettagrid/types.hpp
```diff
@@ -0,0 +1,53 @@
+#ifndef METTAGRID_METTAGRID_TYPES_HPP_
```
sasmith, line 1: I'm surprised this diff doesn't use more of these types. It seems like you defined them; but most actual type changes are at the python layer.

sasmith: (That said, feel free to switch the cpp to use these in a separate PR.)

sasmith (changes requested):
I see this is marked as `hold`, but I've already reviewed it! Possible tooling failure.

Requesting changes for:
* the assertions around agent pointers
* the `astype(..., copy=False)`


General PR-style preference: I would have preferred the python level changes to be separated from the cpp changes. No need to change for this PR.

sasmith (approved):
Didn't fully re-review, but the cpp parts look good.

PR #813, opened 2025-06-04
uv-related improvements and cleanups; rename setup_build.sh -> setup_dev.sh

berekuk:
`uv sync` works significantly faster than `uv pip`, and it's the officially blessed way to install things in uv. This PR uses it everywhere.

All executable scripts include `#!/usr/bin/env -S uv run` line, which makes them runnable without any configuration and without `python -m script.name` (which is more convenient because of shell autocompletion). uv will automatically create `.venv` if it doesn't exist and run sync.

`setup_build.sh` name is now incorrect, so I renamed it to `setup_dev.sh`. Also removed some code from it that didn't do anything useful. `make dev` aliases to it.


berekuk:
* **#813** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/813?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/813?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

CLAUDE.md
```diff
@@ -18,7 +19,7 @@ The codebase consists of:
 curl -LsSf https://astral.sh/uv/install.sh | sh
 
 # Run setup script (creates virtual environment automatically)
-./devops/setup_build.sh
+./devops/setup_dev.sh
```
daveey, line 22: let's move this to ./devops/sandbox/setup.sh

daveey (approved):

CLAUDE.md
berekuk: Don't people run metta locally on their laptops too? I think of sandboxes only as of temporary hosts on aws. (note that I moved `devops/sandbox` to `devops/macos` a while ago)

I'm not even sure that it would work properly on sandboxes because SSO sign-in workflow requires a browser.

daveey: sorry, i call sandboxes their laptops too, but i agree it's not a good name. i forgot we dont have devops/sandbox anymore, so this looks good as is

PR #815, opened 2025-06-04
feat: system monitor

rjwalters:
image
image
image


daveey (approved):
I'm surprised there isn't a decent system monitor available as a pip package. I don't love adding a bunch of code, but it's pretty isolated and if you think this is the way to go, it's fine. just move stats() into system monitor

sasmith:
Also, have you considered https://github.com/osquery/osquery? I haven't looked deeply, but have a general feeling of "we're trying to get lots of system metrics, and it's a pain because they're structured differently, and osquery is built to solve that problem".

PR #818, opened 2025-06-04
feat: initial map state hash

rjwalters:
This PR adds a new property `initial_grid_hash` to MettaGridEnv that provides a simple and efficient way to compare and identify unique map configurations. The hash is computed during environment initialization based on the positions and types of all objects in the grid.

## Changes
- **Added wyhash implementation** (`hash.hpp`): Integrated a minimal, high-performance hash function based on wyhash v4.2 for computing grid checksums
- **Grid hash computation**: Modified `mettagrid_c.cpp` to compute a hash during initialization by concatenating position and type information for all grid objects
- **Python API exposure**: 
  - Added `initial_grid_hash` property to the C++ class and exposed it via pybind11
  - Added corresponding Python property in `MettaGridEnv`
  - Updated type stub file (`mettagrid_c.pyi`) for proper type hints
- **Stats integration**: Added the grid hash to episode statistics attributes for tracking and analysis
- **Testing**: Added comprehensive test to verify hash functionality, including consistency checks and value validation

## Technical Details
- Uses wyhash for its excellent performance and distribution properties
- Hash is computed as a 64-bit unsigned integer from a string representation of grid state
- Format: `"r1,c1:type1;r2,c2:type2;..."` ensures deterministic ordering
- Python handles the full uint64 precision without issues
- https://github.com/rurban/smhasher/tree/master

## Use Cases
- **Reproducibility**: Quickly verify if two episodes started with the same map layout
- **Analytics**: Track performance metrics by unique map configurations
- **Debugging**: Identify specific map layouts that lead to certain behaviors
- **Caching**: Use as a key for storing map-specific data or strategies

## Testing
- Added test verifying hash is within valid uint64 range
- Confirmed hash consistency across environment resets
- Validated specific expected hash value for the benchmark configuration
- All existing tests pass

## Example Usage
```python
env = MettaGridEnv(curriculum, render_mode=None)
env.reset()
print(f"Map hash: {env.initial_grid_hash}")
# Output: Map hash: 15684112579441699572
```

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210595863659907)

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210595732802842)


mettagrid/mettagrid/mettagrid_c.cpp
```diff
@@ -162,6 +170,8 @@ MettaGrid::MettaGrid(py::dict env_cfg, py::list map) {
     }
   }
 
+  initial_grid_hash = wyhash::hash_string(grid_hash_data);
```
sasmith, line 173: What's the advantage over std::hash?

rjwalters: @claude what's the advantage of using a custom implementation of wyhash here over std:hash?

sasmith (approved):
Seems reasonable to see if this is useful.

Added https://app.asana.com/1/1209016784099267/project/1209041170403474/task/1210594297567975, so we can follow up on keeping or removing.

mettagrid/mettagrid/mettagrid_c.cpp
rjwalters: @claude open-pr 

1. add a nice summarized comment to the header of the `hash.hpp` file explaining how the hash differs from std:hash
2. add a short comment at the place where we instantiate the hash as well

rjwalters:
@claude open-pr

    add a nice summarized comment to the header of the hash.hpp file explaining how the hash differs from std:hash
    add a short comment at the place where we instantiate the hash as well

mettagrid/mettagrid/mettagrid_c.cpp
rjwalters: @claude open-pr

    add a nice summarized comment to the header of the hash.hpp file explaining how the hash differs from std:hash
    add a short comment at the place where we instantiate the hash as well

rjwalters:
@claude open-pr

    add a nice summarized comment to the header of the hash.hpp file explaining how the hash differs from std:hash
    add a short comment at the place where we instantiate the hash as well

PR #819, opened 2025-06-05
Switch test_buffer to test with tokens

sasmith:


Al-does (approved):
LGTM. I don't know when this test is triggered but I assume it won't auto fail PR's if its code in if mettagrid.yaml has 'use_observation_tokens' set to `false' and the env is generating CNN obs. I noticed you set use_observation_tokens to true in this so that should handle that concern. And totally agreed with the general motivation here!

PR #820, opened 2025-06-05
Fix L2 renorm and add weight decay to optimizer

Al-does:
Fixed the explicit L2 loss calculation methods. 

Also added weight decay param to the optimizers. Given the way the agent code has evolved, I think this method should be preferred. It has another advantage over explicit L2 loss anyway. Let's experiment with both and see.

To use the explicit path, set `l2_reg_loss_coef` to a positive value (like 0.0002) in puffer.yaml. For the built-in, optimizer path, set `optimizer.weight_decay' to some positive value. 

If the optimizer path works equally than the the explicit path then we should deprecate the explicit path as it's a lot of code overhead.


metta/agent/metta_agent.py
```diff
@@ -416,23 +416,31 @@ def _apply_to_components(self, method_name, *args, **kwargs) -> list[torch.Tenso
...
+            if result is not None:
+                results.append(result)
 
         return results
 
     def l2_reg_loss(self) -> torch.Tensor:
         """L2 regularization loss is on by default although setting l2_norm_coeff to 0 effectively turns it off. Adjust
         it by setting l2_norm_scale in your component config to a multiple of the global loss value or 0 to turn it off.
         """
+        l2_reg_loss = torch.tensor(0.0, device=self.device)
```
rjwalters, line 429:
```suggestion
```

metta/agent/metta_agent.py
```diff
@@ -416,23 +416,31 @@ def _apply_to_components(self, method_name, *args, **kwargs) -> list[torch.Tenso
...
         component_loss_tensors = self._apply_to_components("l2_reg_loss")
-        return torch.sum(torch.stack(component_loss_tensors))
+        if len(component_loss_tensors) > 0:
+            l2_reg_loss = torch.sum(torch.stack(component_loss_tensors))
+        return l2_reg_loss
```
rjwalters, line 433:
```suggestion
else:
    return  torch.tensor(0.0, device=self.device)
```

rjwalters (approved):

PR #843, opened 2025-06-06
Added three new token embedding architectures

Al-does:
These are for experimentation and, pending test results, we'll delete some of them. 

The original token embedder added attribute embeddings to coordinate embeddings. Now we have three more to experiment with:
- Concatenating attribute embeddings with coordinate embeddings
- Representing coordinates with Fourier features and concatenating them with attribute embeddings
- Rotating attribute embeddings using RoPE which is, of course, based on the coordinate value. 

I also created three new agent config files that reflect these architectures and I set their values to have the same final feature embedding dimension so we can run an apples to apples comparison.


devops/skypilot/config/sk_train.yaml
```diff
@@ -2,18 +2,18 @@ resources:
...
-      region: us-east-1
-      use_spot: true
-      accelerators: "A10G:1"
-    - cloud: aws
-      region: us-east-1
-      accelerators: "L4:1"
-    - cloud: aws
-      region: us-east-1
-      accelerators: "A10G:1"
+      use_spot: false
```
sasmith, line 5: Presumably this shouldn't be merged. At least, this looks like it falls outside of "I'm just testing things".

sasmith (approved):
I only skimmed the models. Approving to merge for experimenting, but there are parts of this diff that seem like they're outside the box of "just experimenting".

PR #848, opened 2025-06-06
feat: precalculate maps for faster reset()

rjwalters:
## Summary

This PR adds background precalculation of C++ environment construction arguments to improve reset performance in MettaGridEnv. By preparing the next environment configuration while the current episode is running, we can significantly reduce the time spent in the `reset()` method.

## Motivation

Environment reset can be a performance bottleneck in reinforcement learning, especially when:
- Map generation is computationally expensive
- Configuration processing takes significant time
- The curriculum frequently advances to new tasks

This change addresses these issues by overlapping environment preparation with episode execution.

### Core Implementation
- Added a background thread (`_precalculation_worker`) that continuously prepares environment construction arguments
- Implemented a single-buffered queue system to store precalculated data
- Modified `_initialize_c_env` to use precalculated data when available, with automatic fallback to synchronous calculation

### Key Features
- **Thread-safe design**: Uses `queue.Queue` for safe communication between threads
- **Task consistency**: Ensures the task used for precalculation is the same one used for initialization, preventing race conditions
- **Graceful degradation**: Falls back to synchronous calculation if precalculated data isn't ready
- **Resource efficient**: Single-buffered queue prevents excessive memory usage
- **Proper cleanup**: Threads are properly terminated on environment close

### Technical Details
- Extracted environment preparation logic into `_calculate_c_env_args()` for reuse
- Added `_get_c_env_construction_args()` to manage precalculated vs on-demand calculation
- Task object is now passed through the precalculation pipeline to maintain consistency
- Added proper type hints using the `Task` type from the curriculum module

## Performance Impact

Expected improvements:
- Reduced `reset()` latency when precalculated data is available
- No performance regression when falling back to synchronous mode
- Minimal overhead from the background thread (sleeps when idle)

## Testing

- [x] Verified thread starts and stops correctly
- [x] Tested fallback behavior when queue is empty
- [x] Confirmed proper cleanup on environment close
- [x] Validated task consistency between precalculation and usage

## Future Improvements

Potential follow-ups:
- Add metrics to track precalculation hit rate
- Consider multi-buffering for environments with variable reset times
- Optimize the worker thread sleep/wake patterns based on profiling


rjwalters:
I am closing this because the trainer already effectively parallelizes the environment generation -- so promising benchmark results do not translate into better real world performance. We should explore S3 caching instead.

PR #850, opened 2025-06-07
Standardized ascii encoding in maps & related cleanups

berekuk:
## Changes
- some changes in which symbols we use (see below)
- removed emoji support (because of double width I think it wasn't useful)
- simplified nethack renderer as a consequence
- refactored ascii map tests
- supported new encoding in StorableMap, which will make tools.map.gen and mettamap compatible with all maps
- simplified tests because we use `.` instead of space now and `||` boundaries are not necessary
- added scripts that can normalize maps and updated all existing map and scene configs

## Encoding decisions

- `@` for default agents instead of `A` (nethack convention)
- `.` instead of ` ` - more readable
- `_` instead of `a` for altars (nethack again)
- `#` instead of `W` - I think `#` is prettier because it's symmetrical
- `L` for lab, `S` for lasery (I made sure no ascii maps were using either)
- no `AP` and `Ap` for predators/prey (again, unused, and double width means it was always a hack)

## Future directions

I have three more ideas that I wanted to do, but this is big enough already:

1. reuse these encodings in `map_editor_ui` (by sharing a common json, probably)
2. make a breaking change again that would use lower-case for all objects, and reserve the upper-case for "aliases"; for example, you could write `generator.blue=B` at the top of your map file, and then use that `B` everywhere in the following map
3. update nethack renderer to output text in color

(1) will be good to do when we fold map_editor_ui to Observatory;
(2) would be cool but if we get a decent map editor soon enough then it doesn't matter, we won't care about ascii encodings;
(3) would be easier to do when (2) is taken care of (i.e., when we won't need to use different characters for objects of the same type but different color)

---

Tip on how to review: use file type filters, https://github.com/Metta-AI/metta/pull/850/files?file-filters%5B%5D=.json&file-filters%5B%5D=.py&show-deleted-files=true&show-viewed-files=true


rjwalters (approved):
lgtm

relh (approved):

PR #866, opened 2025-06-09
feat: dependency synchronization on each dependabot PR

rjwalters:
This PR introduces an automated system to keep `mettagrid` and `metta` dependency versions synchronized for workspace compatibility.

## 📖 Background

We previously had UV managing our requirements with a single source of truth, but **scikit-build-core requires us to keep requirements separated** in the metta and mettagrid pyproject files. This creates a problem for Dependabot:

- We could configure Dependabot to separately open PRs on both `metta/pyproject.toml` and `mettagrid/pyproject.toml`
- But it's **annoying to keep things synchronized** manually
- The `dependabot.yml` configuration options available for this create **a third file that needs to be updated** when dependency lists change
- **Version drift** between packages causes compatibility issues

**This solution runs a sync script that drives mettagrid versions from metta** on each Dependabot update, maintaining metta as the single source of truth while preserving mettagrid's curated dependency list.

### 🛠️ Build System Updates
- **Updated build requirements** to use version ranges instead of exact pins:
  - `scikit-build-core>=0.11` (was unversioned)
  - `pybind11>=2.10,<3.0` (was `==2.10.4`)
  - `numpy>=2.2.6,<2.3` in mettagrid (was `==2.2.5`)
- **Added `wheel` to root build requirements** for consistency

### 🤖 Automated Dependency Management
- **New sync script** (`.github/scripts/sync_dependencies.py`):
  - Preserves mettagrid's dependency list but syncs versions with metta
  - Handles complex dependency formats (extras, operators, etc.)
  - Robust error handling and clear logging

### 🔄 GitHub Actions Workflow
- **Replaced** `assign-dependabot.yml` with `process-dependabot.yml`
- **Automated workflow** runs 1 hour after Dependabot (10:00 UTC)
- **For each Dependabot PR**:
  1. Checks out the PR branch
  2. Runs dependency sync script
  3. Commits and pushes sync changes if needed
  4. Assigns reviewers (existing functionality)
- **Manual trigger options** for testing specific PRs


daveey (changes requested):
I think this is for @berekuk to figure out. i'm not excited about adding a giant python file for this, but he can make the call

sasmith (commented):
Oh, nevermind. It's already assigned to @berekuk . Graphite's UI makes it feel like all PRs are assigned to me.

berekuk:
My thoughts on this:
- it's not a problem by itself if `>=` versions in pyproject files get out of sync, because we have a single `uv.lock`; actually could be a feature if some of our packages are more lenient than others
- we should wait and see how dependabot handles our setup after #864, I don't mind waiting until Monday but we can switch to daily if you're curious
- if dependabot doesn't work well we should switch to renovate instead of trying to patch dependabot; the list of [open uv-related dependabot issues](https://github.com/dependabot/dependabot-core/issues?q=is%3Aissue%20state%3Aopen%20uv) is pretty long

PR #873, opened 2025-06-09
fix: refactor action priority handling

rjwalters:
This PR refactors the action processing loop in `MettaGrid::_step()` to improve priority handling efficiency, add proper error reporting, and implement early termination for successful actions.

### 1. Dynamic Priority Level Collection
**Before:**
```cpp
for (unsigned char p = 0; p <= _max_action_priority; p++) {
    // Uses: _max_action_priority - p
}
```

**After:**
```cpp
std::set<unsigned char> priority_levels;
for (const auto& handler : _action_handlers) {
    priority_levels.insert(handler->priority);
}
```

**Benefits:**
- Only processes priority levels that actually have registered handlers (e.g., priorities 1, 5, 10)
- Self-maintaining when new handlers with different priorities are added

### 2. Early Action Success Termination
**New behavior:**
```cpp
if (_action_success[agent_idx]) {
    continue;
}
```

Agents who have already successfully performed an action in the current step are skipped for subsequent priority levels. This prevents agents from performing multiple actions per step and ensures fair action distribution.

### 3. Enhanced Error Handling and Validation
**Before:**
```cpp
if (action < 0 || action >= _num_action_handlers) {
    printf("Invalid action: %d\n", action);
    continue;
}
```

**After:**
```cpp
if (action_id < 0 || action_id >= _num_action_handlers) {
    throw std::runtime_error("Invalid action ID " + std::to_string(action_id) + " for agent " +
                             std::to_string(agent_idx) + ". Valid range: 0 to " +
                             std::to_string(_num_action_handlers - 1));
}
```

**Improvements:**
- Detailed error messages including agent index, action ID, and valid ranges
- Exception-based error handling for better debugging
- Argument validation now includes action name from handler for clearer error context
- Fail-fast approach prevents silent failures

### 4. Code Clarity and Maintainability
- More descriptive variable names (`agent_idx` vs `idx`, `action_id` vs `action`)
- Explicit casting for `ActionArg`

## Behavioral Changes

**Breaking Change:** Agents are now limited to one successful action per step. Previously, an agent could potentially perform actions at multiple priority levels within the same step.

### DIFF
--- a/mettagrid/mettagrid/mettagrid_c.cpp
+++ b/mettagrid/mettagrid/mettagrid_c.cpp
@@ -321,30 +321,52 @@ void MettaGrid::_step(py::array_t<ActionType, py::array::c_style> actions) {
   current_step++;
   _event_manager->process_events(current_step);
 
-  // Process actions by priority
-  for (unsigned char p = 0; p <= _max_action_priority; p++) {
-    for (size_t idx = 0; idx < _agents.size(); idx++) {
-      int action = actions_view(idx, 0);
-      if (action < 0 || action >= _num_action_handlers) {
-        printf("Invalid action: %d\n", action);
+  // Collect unique priority levels from action handlers
+  std::set<unsigned char> priority_levels;
+  for (const auto& handler : _action_handlers) {
+    priority_levels.insert(handler->priority);
+  }
+
+  // Process actions by priority levels (highest to lowest)
+  for (auto it = priority_levels.rbegin(); it != priority_levels.rend(); ++it) {
+    unsigned char current_priority = *it;
+
+    for (size_t agent_idx = 0; agent_idx < _agents.size(); agent_idx++) {
+      // Skip agents who already successfully performed an action this step
+      if (_action_success[agent_idx]) {
         continue;
       }
 
-      ActionArg arg = actions_view(idx, 1);
-      auto& agent = _agents[idx];
-      auto& handler = _action_handlers[action];
+      // Extract action data
+      int action_id = actions_view(agent_idx, 0);
+      ActionArg action_arg = static_cast<ActionArg>(actions_view(agent_idx, 1));
+      Agent* agent = _agents[agent_idx];
 
-      if (handler->priority != _max_action_priority - p) {
+      // Validate action ID
+      if (action_id < 0 || action_id >= _num_action_handlers) {
+        throw std::runtime_error("Invalid action ID " + std::to_string(action_id) + " for agent " +
+                                 std::to_string(agent_idx) + ". Valid range: 0 to " +
+                                 std::to_string(_num_action_handlers - 1));
+      }
+
+      auto& handler = _action_handlers[action_id];
+
+      // Skip if this handler doesn't match current priority level
+      if (handler->priority != current_priority) {
         continue;
       }
 
-      if (arg > _max_action_args[action]) {
-        continue;
+      // Validate action argument
+      if (action_arg > _max_action_args[action_id]) {
+        throw std::runtime_error("Action argument " + std::to_string(action_arg) + " exceeds maximum " +
+                                 std::to_string(_max_action_args[action_id]) + " for action " +
+                                 std::to_string(action_id) + " (" + handler->action_name() + ")" + " on agent " +
+                                 std::to_string(agent_idx));
       }
 
       // handle_action expects a GridObjectId, rather than an agent_id, because of where it does its lookup
-      bool success = handler->handle_action(agent->id, arg);
-      _action_success[idx] = success;
+      bool success = handler->handle_action(agent->id, action_arg);
+      _action_success[agent_idx] = success;
     }
   }
 


mettagrid/mettagrid/mettagrid_c.cpp
```diff
@@ -321,30 +321,49 @@ void MettaGrid::_step(py::array_t<ActionType, py::array::c_style> actions) {
...
+  for (const auto& handler : _action_handlers) {
+    priority_levels.insert(handler->priority);
+  }
+
+  // Process actions by priority levels (highest to lowest)
+  for (auto it = priority_levels.rbegin(); it != priority_levels.rend(); ++it) {
+    unsigned char current_priority = *it;
+
+    for (size_t agent_idx = 0; agent_idx < _agents.size(); agent_idx++) {
+      // Skip agents who already successfully performed an action this step
```
sasmith, line 335: This doesn't feel like a savings over short circuiting if the handle priority is wrong.

mettagrid/mettagrid/mettagrid_c.cpp
```diff
@@ -321,30 +321,49 @@ void MettaGrid::_step(py::array_t<ActionType, py::array::c_style> actions) {
...
+        throw std::runtime_error("Invalid action ID " + std::to_string(action_id) + " for agent " +
+                                 std::to_string(agent_idx) + ". Valid range: 0 to " +
+                                 std::to_string(_num_action_handlers - 1));
       }
 
-      if (arg > _max_action_args[action]) {
+      auto& handler = _action_handlers[action_id];
+
+      // Skip if this handler doesn't match current priority level
+      if (handler->priority != current_priority) {
```
sasmith, line 353: Changing the loop so we can use `current_priority` instead of `_max_action_priority - p` seems like good cleanup.

sasmith (changes requested):
I'm not really sure of the motivation behind this PR. I know the LLM-generated description says it improves priority handling efficiency, but I'm not convinced it does, not that this was a problem.

Can you say more about why we should be making changes here?

rjwalters:
I think there are two problems I am trying to solve here:

1. I worry about invalid actions being converted to no-op by 
```c
     if (arg > _max_action_args[action]) {
        continue;
      }
```

2. I think looping over an amount to subtract from the priority level is unclear:
```c
for (unsigned char p = 0; p <= _max_action_priority; p++) {
```
vs
```c
for (unsigned char current_priority = max_action_priority; p != 255; p--)
```

mettagrid/mettagrid/action_handler.hpp
```diff
@@ -26,7 +26,7 @@ class ActionHandler {
     this->_grid = grid;
   }
 
-  bool handle_action(GridObjectId actor_object_id, ActionArg arg) {
+  bool handle_action(GridObjectId actor_object_id, ActionType arg) {
```
sasmith, line 29: `ActionType` seems less clear that `ActionArg`, and it collapses the distinction between actions and their arguments.

mettagrid/mettagrid/mettagrid_c.cpp
```diff
@@ -329,30 +329,40 @@ void MettaGrid::_step(py::array_t<ActionType, py::array::c_style> actions) {
...
-      int action = actions_view(idx, 0);
-      if (action < 0 || action >= _num_action_handlers) {
-        printf("Invalid action: %d\n", action);
+  // Process actions by priority levels (highest to lowest)
+  for (unsigned char offset = 0; offset <= _max_action_priority; offset++) {
+    unsigned char current_priority = _max_action_priority - offset;
+
+    for (size_t agent_idx = 0; agent_idx < _agents.size(); agent_idx++) {
+      // Skip agents who already successfully performed an action this step
+      if (_action_success[agent_idx]) {
```
sasmith, line 338: This still doesn't seem like a savings. Why not just move up the `if (handler->priority != current_priority) {` check?

mettagrid/mettagrid/mettagrid_c.cpp
```diff
@@ -329,30 +329,40 @@ void MettaGrid::_step(py::array_t<ActionType, py::array::c_style> actions) {
...
         continue;
       }
 
-      ActionArg arg = actions_view(idx, 1);
-      auto& agent = _agents[idx];
+      ActionType action = actions_view(agent_idx, 0);
+      ActionType arg = actions_view(agent_idx, 1);
+
+      if (action < 0 || action >= _num_action_handlers) {
+        throw std::runtime_error("Invalid action type " + std::to_string(action) + ". Valid range: 0 to " +
```
sasmith, line 346: You haven't convinced me we should change this behavior. I agree that I don't expect our agents to submit invalid actions; but I'm not convinced that the right behavior is to crash if they do. E.g., if we get a 3rd party submission, I don't think the game should crash if they attempt to take invalid actions. (I understand this setting doesn't exist yet -- it's mostly an example point in the bucket of "what's the correct, long term behavior here"?)

sasmith (changes requested):

mettagrid/mettagrid/action_handler.hpp
rjwalters: both are supplied to the env in the same actions array and so I think they necessarily have the same type?

rjwalters: we can move to types and match there but keep the typedef

mettagrid/mettagrid/mettagrid_c.cpp
rjwalters: @daveey do you want to weigh in on this?

rjwalters:
> Sorry for the delay when this was passed back to me.

no worries - this PR has changed purpose several times at this point - I can reopen it without all the commit history if you prefer

rjwalters:
@claude review comments

PR #890, opened 2025-06-10
Add postgres_stats_db

Barraketh:
Add postgres_stats_db

Move observatory to FE

Add server.py and call it


Barraketh:
* **#890** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/890?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/890?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mettagrid/mettagrid/postgres_stats_db.py
```diff
@@ -0,0 +1,221 @@
...
+            )""",
+        ],
+    ),
+]
+
+
+class PostgresStatsDB:
+    def __init__(self, db_url: str) -> None:
+        self.db_url = db_url
+        self.con = Connection.connect(self.db_url)
```
Barraketh, line 76: I'll probably switch this to a connection pool

mettagrid/mettagrid/stats_writer.py
berekuk, line 1: Assuming we still want to publish `mettagrid` on PyPI, we should move `StatsWriter` _implementation_ to metta, and leave only the interface (`ABC`?) here.

observatory/backend/server.py
berekuk, line 1: Bikeshedding on naming, areas of responsibility and repo layout:

I wonder if `observatory.backend` is going to be a separate python package (with its own `pyproject.toml` file) or just a module inside `metta` package. Right now it's the latter, you don't have pyproject.

But given how tightly coupled it is with metta code (same postgres DB), I think it might make sense to move this code under `metta.server`, and keep the previous observatory layout only for the frontend.

One reason why I think so is that the same server is _probably_ going to be unified with `mettascope.server` code (provide a websocket for live sessions), and maybe for other purposes, so it's not just for observatory frontend.

Barraketh: Well, it doesn't currently have it's own pyproject.toml, but I'm happy to move it under metta.server - that logic makes sense to me

mettagrid/mettagrid/stats_writer.py
Barraketh: Can you elaborate on why? I don't mind doing that, just want to understand your reasoning better

observatory/backend/server.py
```diff
@@ -0,0 +1,350 @@
...
+class GroupDiff(BaseModel):
+    group_1: str
+    group_2: str
+
+
+# Type alias for group metric
+GroupHeatmapMetric = Optional[str] | GroupDiff
+
+
+@app.get("/api/suites")
```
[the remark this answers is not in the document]
Barraketh: This should be "/observatory" - that's one of the things I would have cleaned up before opening the PR

mettagrid/mettagrid/stats_writer.py
berekuk: When I imagine someone who's trying to use PyPI mettagrid instead of this full metta project, I don't expect them to be happy to set up postgres instance just to log stats.

`MettaGridEnv` takes `stats_writer` parameter and pushes stats into it, but it doesn't care about the implementation. So it looks to me like a dependency injection pattern:
- `MettaGridEnv` can push stats if you give it a hook
- but the hook is very MettaAI-specific, it relies on other infra that we'll have only in the outer metta

(in general, I think we should try to make `mettagrid` minimal, e.g. we discussed previously how we don't want it to depend on hydra and maybe even on omegaconf, but after curriculum we've fallen back on using those)

Barraketh: Well, the stats_writer is generally optional, so you can always run without one.  And then it might be nice to have a stats writer inside of mettagrid in case you DO want to collect stats.  But i'm happy to keep mettagrid more minimal

PR #892, opened 2025-06-10
Distributed training by using global batch_size, minibatch_size and number of environments 

dominik3141:
This PR implements distributed training by keeping `batch_size`, `minibatch_size` and number of environments constant on a global level, i.e. we divide those hyperparameters by the `world_size` on each rank.
We also make sure to properly conduct all operations that need global data on a global level, i.e. all batch norms are calculated over the full global batch and the average advantage is also calculated over all ranks.
The previous implementation of distributed training by scaling the `update_epoch` is being removed. 

Here an example of the performance on the navigation curriculum:
W B Chart 6_10_2025, 10_47_34 AM


metta/rl/pufferlib/trainer.py
```diff
@@ -136,6 +142,11 @@ def __init__(
 
         self.policy.activate_actions(actions_names, actions_max_params, self.device)
 
+        if torch.distributed.is_initialized():
```
daveey, line 145: can this go into metta_agent.py:DistributedMettaAgent ?

metta/rl/pufferlib/trainer.py
```diff
@@ -700,6 +761,14 @@ def _process_stats(self):
         # Now synchronize and aggregate stats across processes
         sps = self.profile.SPS
         agent_steps = self.agent_step
+        if torch.distributed.is_initialized():
```
daveey, line 764: do we need to do that? can't we just multiply agent_steps * world_size? it's going to be the same on all GPUs

metta/rl/pufferlib/trainer.py
```diff
@@ -839,8 +909,17 @@ def _make_vecenv(self):
         if self.target_batch_size < 2:  # pufferlib bug requires batch size >= 2
             self.target_batch_size = 2
 
-        self.batch_size = (self.target_batch_size // self.trainer_cfg.num_workers) * self.trainer_cfg.num_workers
+        if torch.distributed.is_initialized():
```
daveey, line 912: if you need world_size, set self._world_size in init() rather than here.

metta/rl/pufferlib/trainer.py
```diff
@@ -839,8 +909,17 @@ def _make_vecenv(self):
...
+            world_size = torch.distributed.get_world_size()
+        else:
+            world_size = 1
+
+        self.batch_size = int(self.target_batch_size)
```
daveey, line 917: self.batch_size is not used outside of this function, let's make it local. but i think you should just change trainer. forward_pass_minibatch_target_size in train.py and not make any changes here

daveey (changes requested):
glad you got this working!

metta/rl/pufferlib/trainer.py
```diff
@@ -60,6 +60,12 @@ def __init__(
         if torch.distributed.is_initialized():
             self._master = int(os.environ["RANK"]) == 0
             self._world_size = torch.distributed.get_world_size()
+
+            # some config parameters have to be interpreted differently for distributed training
+            # i.e. as global parameters, that have to be divided across all ranks
+            self.trainer_cfg.batch_size = self.trainer_cfg.batch_size // self._world_size
```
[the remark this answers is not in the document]
dominik3141: Ok, now using _batch_size and _minibatch_size for that

metta/rl/pufferlib/trainer.py
dominik3141: Yes

metta/rl/pufferlib/trainer.py
```diff
@@ -452,7 +468,25 @@ def _train(self):
             if self.trainer_cfg.average_reward:
```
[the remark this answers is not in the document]
dominik3141: Done

metta/rl/pufferlib/trainer.py
```diff
@@ -513,7 +547,26 @@ def _train(self):
 
                     adv = adv.reshape(-1)
                     if self.trainer_cfg.norm_adv:
-                        adv = (adv - adv.mean()) / (adv.std() + 1e-8)
+                        if torch.distributed.is_initialized():
```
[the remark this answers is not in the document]
dominik3141: Done

metta/rl/pufferlib/trainer.py
dominik3141: I agree! Maybe we should even rename it so it's less confusing. I always think of this batch_size as the `vecenv.recv()` package size

dominik3141:
@daveey I implemented all your suggestions. Let me know if there is anything else you would like me to change

metta/rl/pufferlib/trainer.py
```diff
@@ -700,6 +711,10 @@ def _process_stats(self):
         # Now synchronize and aggregate stats across processes
         sps = self.profile.SPS
         agent_steps = self.agent_step
+        steps_per_update = 0.0
```
daveey, line 714: why is this a float?

metta/rl/pufferlib/trainer.py
```diff
@@ -762,6 +778,32 @@ def _process_stats(self):
...
         self.stats.clear()
 
+    def _compute_advantage(self, adv: torch.Tensor) -> torch.Tensor:
+        """Compute normalized advantages, handling distributed training synchronization."""
+        adv = adv.reshape(-1)
+        if self.trainer_cfg.norm_adv:
+            if torch.distributed.is_initialized():
+                # in the case of distributed training we have to synchronize the advantage mean
+                # and standard deviation to get the same normalized advantage on all ranks
+                local_sum = adv.sum().unsqueeze(0)
```
daveey, line 788: use einops.rearange instead of unsqueeze

metta/rl/pufferlib/trainer.py
```diff
@@ -839,20 +881,23 @@ def _make_vecenv(self):
         if self.target_batch_size < 2:  # pufferlib bug requires batch size >= 2
             self.target_batch_size = 2
 
-        self.batch_size = (self.target_batch_size // self.trainer_cfg.num_workers) * self.trainer_cfg.num_workers
-        num_envs = self.batch_size * self.trainer_cfg.async_factor
+        batch_size = int(self.target_batch_size)
```
daveey, line 884: this isn't right, if you have target_batch_size = 100 and num_workers=3, batch_size should be 99 not 100

daveey, line 884: we can call it forward_pass_batch_size

tools/train.py
```diff
@@ -45,6 +46,10 @@ def train(cfg, wandb_run, logger: Logger):
 
     policy_store = PolicyStore(cfg, wandb_run)
 
+    if torch.distributed.is_initialized():
+        world_size = torch.distributed.get_world_size()
```
daveey, line 50: what do you think about moving most (or maybe all) of the dist code from trainer.py out to here?

we can just pass world_size, is_master, device, and a modified cfg.trainer to trainer.__init__() 

then only train.py needs to know about distributed training

daveey (changes requested):
Much better. A few more changes. Also cursor() has a couple of suggestions, not sure if it's right

metta/rl/pufferlib/trainer.py
dominik3141: Thanks for clarifying, I changed that back

tools/train.py
dominik3141: Personally I think that would be an aesthetic change and the machine code at runtime (which is the thing I care about) would look approximately the same.

dominik3141:
> Also, I think we should be scaling num_workers by NUM_GPUS (not world_size) in train.sh. Otherwise when we run on 4 GPUs we make 4 vecenvs and end up with 4x the workers

What would be the problem with having 4x the workers when using 4 GPUs? Having 4 GPUs (on a single node) means we have 4x consumers of vector steps, so having 4x workers makes sense from that perspective. Am I missing something? Even if that means having more workers than cores, I don't see a problem with that. Let the CPU figure it out, they tend to have some great concurrency mechanisms. Especially with our CPUs using hyper-threading.

tools/train.py
dominik3141: Let me know if you still want me to do that, but then I would do it in a separate PR so that people can finally start training on multiple GPUs

daveey (approved):
thanks for making the changes. i'll merge this, but could you follow up with a cleanup PR and scale the num_workers?

PR #898, opened 2025-06-10
fix: remove hydra proxy from mettagrid

rjwalters:


mettagrid/mettagrid/mettagrid_env.py
```diff
@@ -99,12 +97,13 @@ def __init__(
     def _make_episode_id(self):
         return str(uuid.uuid4())
 
-    def _reset_env(self):
-        # Prepare the level
+    def _initialize_c_env(self, task: Task) -> None:
+        """Initialize the C++ environment."""
         self._task = self._curriculum.get_task()
```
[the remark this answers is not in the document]
berekuk: I think Diamond is right about this
(out of its two options, I'd prefer (1), if we have `self._task` we might as well use it everywhere instead of risking an inconsistency)

berekuk (approved):
approving, but please do fix the comment first

mettagrid/mettagrid/mettagrid_env.py
```diff
@@ -99,12 +97,12 @@ def __init__(
     def _make_episode_id(self):
         return str(uuid.uuid4())
 
-    def _reset_env(self):
-        # Prepare the level
-        self._task = self._curriculum.get_task()
```
sasmith, line 104: This logic looks wrong, since we already pick a task in __init__; so good fix (unless I'm missing something?).

mettagrid/mettagrid/mettagrid_env.py
```diff
@@ -99,12 +97,12 @@ def __init__(
...
+    def _initialize_c_env(self, task: Task) -> None:
```
sasmith, line 100: Passing the task here is confusing, given that this function can access self._task. It implies that `task` could be different from `self._task`, which it had better not be.

PR #902, opened 2025-06-10
refactor average reward

rjwalters:
# Fix Average Reward Filter Checkpoint Recovery

## Problem
When using average reward RL with checkpointing, the exponential moving average (EMA) filter for the average reward estimate was not being properly restored from checkpoints. This caused two issues:

1. **Filter State Loss**: The `_average_reward_estimate` was being reinitialized to 0.0 on every checkpoint reload, losing the accumulated EMA state
2. **Inconsistent Reward Adjustment**: This reset caused sudden shifts in the reward baseline, leading to unstable advantage computation and training instability

## Root Cause
The previous implementation used `self.average_reward` directly without proper checkpoint persistence, and the EMA filter state was not being saved/loaded with the checkpoint data.

## Solution
- **Checkpoint Integration**: Save and restore `average_reward` in checkpoint `extra_args` to maintain filter continuity
- **Robust Initialization**: Initialize `_average_reward_estimate` from checkpoint data when available, falling back to 0.0 for new training runs
- **Improved Buffer Calculation**: Use `_get_experience_buffer_mean_reward()` for more accurate batch statistics instead of raw `np.mean(rewards_np)`
- **Consistent State Management**: Use private `_average_reward_estimate` attribute with proper initialization logic

## Key Changes
1. Added checkpoint loading of `average_reward` from `extra_args`
2. Implemented lazy initialization of `_average_reward_estimate` with checkpoint recovery
3. Added `_get_experience_buffer_mean_reward()` helper for better batch mean calculation
4. Ensured EMA filter state persists across training restarts

## Testing
- [ ] Verify average reward training converges with checkpoint restarts
- [ ] Confirm EMA filter maintains proper state across save/load cycles
- [ ] Test that new training runs still initialize correctly (no regression)

This fix ensures training stability when using average reward RL with checkpointing, preventing the filter reset artifacts that could destabilize learning.

### DIFF
--- a/metta/rl/pufferlib/trainer.py
+++ b/metta/rl/pufferlib/trainer.py
@@ -93,6 +93,10 @@ class PufferTrainer:
         os.makedirs(cfg.trainer.checkpoint_dir, exist_ok=True)
         checkpoint = TrainerCheckpoint.load(cfg.run_dir)
 
+        self._checkpoint_avg_reward = None
+        if checkpoint.policy_path and "average_reward" in checkpoint.extra_args:
+            self._checkpoint_avg_reward = checkpoint.extra_args["average_reward"]
+
         policy_record = None
         load_policy_attempts = 10
         while policy_record is None and load_policy_attempts > 0:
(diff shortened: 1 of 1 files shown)


relh (approved):
Yeah, I'm not sure we actually use this average_reward setting? I was talking about this with Joseph the other day. I think the else branch is our default because of trainer.yaml.

I'm in general skeptical of complexity, I tried running this and it runs fine. I think we probably want better long term tests. I also think we might be able to roll average_reward and _average_reward_estimate into one variable.

Checkout https://github.com/Metta-AI/metta/compare/robb/0610-refactor-avg-reward..richard-simplified-rewards?expand=1 if you want to merge in that simplification.

PR #914, opened 2025-06-11
feat: skypilot launch precheck

rjwalters:
This PR enhances the SkyPilot launch script with validation checks and better user experience, focusing on the two most common launch mistakes.

### What's Changed

**New validation checks:**
- ✅ **Git state validation** - Warns when uncommitted changes won't be reflected in cloud jobs
- ✅ **Config file validation** - Checks that referenced config files actually exist (e.g. `agent=dqn` → `./configs/agent/dqn.yaml`)
- ✅ **Helpful suggestions** - Shows available config files when typos are detected

**Enhanced user experience:**
- 📋 **Job summary display** - Shows what will be launched before submission
- 🤝 **Confirmation prompt** - Prevents accidental launches  
- 🏷️ **Better copy naming** - Multi-copy jobs now get proper names (`run_1`, `run_2`, etc.)

**New command line options:**
- `--skip-git-check` - Skip git validation for automation
- `--skip-validation` - Skip confirmation prompt

### Example Usage

```bash
# Normal usage with validation and confirmation
./launch.py train my_experiment agent=dqn trainer=basic

# Skip validations for CI/automation  
./launch.py train automated_run --skip-git-check --skip-validation

# Preview what would be launched
./launch.py train test_config --dry-run
```

### Files Changed
- `devops/skypilot/launch.py` - Added validation workflow and job summary
- `devops/skypilot/utils.py` - New validation and display utilities

The changes maintain backward compatibility while adding safety guardrails for the most common launch mistakes.

### DIFF
--- a/devops/skypilot/launch.py
+++ b/devops/skypilot/launch.py
@@ -3,10 +3,17 @@ import argparse
 import copy
 import shlex
 import subprocess
+import sys
 
 import sky
 
-from devops.skypilot.utils import launch_task
+from devops.skypilot.utils import (
+    check_config_files,
+    check_git_state,
+    display_job_summary,
+    get_user_confirmation,
+    launch_task,
+)
 
 
 def patch_task(
@@ -82,7 +89,16 @@ def main():
         default=None,
         help="Automatically terminate the job after this many hours (supports decimals, e.g., 1.5 for 90 minutes)",
     )
+    parser.add_argument("--skip-git-check", action="store_true", help="Skip git state validation")
+    parser.add_argument("--skip-validation", action="store_true", help="Skip confirmation prompt")
     (args, cmd_args) = parser.parse_known_args()
+
+    # Run validations
+    if not check_git_state(args.skip_git_check):
+        sys.exit(1)
+
+    if not check_config_files(cmd_args):
+        sys.exit(1)
 
     git_ref = args.git_ref
     if not git_ref:
@@ -104,15 +120,36 @@ def main():
         task, cpus=args.cpus, gpus=args.gpus, nodes=args.nodes, no_spot=args.no_spot, timeout_hours=args.timeout_hours
     )
 
+    # Show job summary
+    extra_details = {}
+    if args.copies > 1:
+        extra_details["copies"] = args.copies
+
+    display_job_summary(
+        job_name=args.run,
+        cmd=args.cmd,
+        task_args=cmd_args,
+        git_ref=git_ref,
+        timeout_hours=args.timeout_hours,
+        task=task,
+        **extra_details,
+    )
+
+    # Get confirmation
+    if not get_user_confirmation(args.dry_run, args.skip_validation, "Should we launch this task?"):
+        sys.exit(0)
+
+    # Launch the task(s)
     if args.copies == 1:
         launch_task(task, dry_run=args.dry_run)
     else:
-        for _ in range(1, args.copies + 1):
+        for i in range(1, args.copies + 1):
             copy_task = copy.deepcopy(task)
-            run_id = args.run
+            run_id = f"{args.run}_{i}"
             copy_task = copy_task.update_envs({"METTA_RUN_ID": run_id})
             copy_task.name = run_id
             copy_task.validate_name()
+            print(f"\nLaunching copy {i}/{args.copies}: {run_id}")
             launch_task(copy_task, dry_run=args.dry_run)
 
 
--- a/devops/skypilot/utils.py
+++ b/devops/skypilot/utils.py
@@ -1,9 +1,19 @@
+import os
 import re
 import sys
+from pathlib import Path
+from typing import List, Optional
 
 import sky
 
-from metta.util.colorama import blue, bold, green
+from metta.util.colorama import blue, bold, cyan, green, red, use_colors, yellow
+from metta.util.fs import cd_repo_root
+from metta.util.git import (
+    get_commit_message,
+    get_current_commit,
+    has_unstaged_changes,
+    is_commit_pushed,
+)
 
 
 def print_tip(text: str):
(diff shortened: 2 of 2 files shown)


devops/skypilot/launch.py
```diff
@@ -82,37 +94,79 @@ def main():
...
         help="Automatically terminate the job after this many hours (supports decimals, e.g., 1.5 for 90 minutes)",
     )
+    parser.add_argument("--skip-git-check", action="store_true", help="Skip git state validation")
+    parser.add_argument("--skip-validation", action="store_true", help="Skip confirmation prompt")
+    parser.add_argument(
+        "--id_suffix",
+        dest="id_suffix",
+        type=str,
+        default=None,
+        help="Suffix to append to run ID (e.g., --id_suffix=0 makes 'base_name.0')",
```
daveey, line 104: when do we want this?

devops/skypilot/launch.py
```diff
@@ -82,37 +94,79 @@ def main():
...
     if args.copies == 1:
         launch_task(task, dry_run=args.dry_run)
     else:
-        for _ in range(1, args.copies + 1):
+        for i in range(1, args.copies + 1):
             copy_task = copy.deepcopy(task)
-            run_id = args.run
-            copy_task = copy_task.update_envs({"METTA_RUN_ID": run_id})
-            copy_task.name = run_id
+            copy_run_id = f"{run_id}_{i}"
```
daveey, line 165: let's not do this, it'll break sweeps workers

devops/skypilot/launch.py
```diff
@@ -82,37 +94,79 @@ def main():
         default=None,
         help="Automatically terminate the job after this many hours (supports decimals, e.g., 1.5 for 90 minutes)",
     )
+    parser.add_argument("--skip-git-check", action="store_true", help="Skip git state validation")
```
berekuk, line 97: Why do we need this option?

devops/skypilot/launch.py
```diff
@@ -66,7 +73,12 @@ def patch_task(
 
 
 def main():
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(
+        description="Launch SkyPilot training jobs",
+        epilog="Usage: %(prog)s <cmd> <run_id> [task_args...] [options]\n"
```
berekuk, line 78: `launch.py --help` already prints `usage:` line on top

devops/skypilot/utils.py
```diff
@@ -34,3 +39,236 @@ def launch_task(task: sky.Task, dry_run=False):
...
+    if not skip_validation:
+        response = input(f"{prompt} (Y/n): ").strip().lower()
+        if response not in ["", "y", "yes"]:
+            print(yellow("Action cancelled by user."))
+            return False
+
+    return True
+
+
+def validate_and_launch_task(
```
berekuk, line 225: This function is unused.

devops/skypilot/launch.py
```diff
@@ -82,37 +94,79 @@ def main():
         default=None,
         help="Automatically terminate the job after this many hours (supports decimals, e.g., 1.5 for 90 minutes)",
     )
+    parser.add_argument("--skip-git-check", action="store_true", help="Skip git state validation")
+    parser.add_argument("--skip-validation", action="store_true", help="Skip confirmation prompt")
```
berekuk, line 98: Are you sure we need this?
I remember David pushing back even on sync launch by default for the sake of performance.

devops/skypilot/utils.py
```diff
@@ -34,3 +39,236 @@ def launch_task(task: sky.Task, dry_run=False):
     print(f"- Or, visit: {bold(f'{dashboard_url()}/api/stream?request_id={short_request_id}')}")
     print("  - To sign in, use credentials from your ~/.skypilot/config.yaml file.")
     print(f"- To cancel the request, run: {bold(f'sky api cancel {short_request_id}')}")
+
+
+def check_git_state(skip_check: bool = False) -> bool:
+    """Check for uncommitted changes that won't be reflected in the cloud."""
```
berekuk, line 45: This function doesn't check that the changes are pushed to github. You should use `metta.util.git.is_commit_pushed` for the correct version.

devops/skypilot/utils.py
```diff
@@ -34,3 +39,236 @@ def launch_task(task: sky.Task, dry_run=False):
...
+        print(yellow("⚠️  Could not check git status (not in a git repo?)"))
+
+    return True
+
+
+def check_config_files(cmd_args: List[str]) -> bool:
+    """Check that config files referenced in arguments actually exist."""
+    config_files_to_check = []
+
+    for task_arg in cmd_args:
```
berekuk, line 71: I have two issues with these checks:
1) incomplete - we have more config groups than `agent` and `trainer`
2) copy-pasted (and the implementation doesn't distinguish between the first two things, which are hydra's config groups, and second two, which are our `config_from_path` things)

Suggestions:
- validate all of config options by running the fake `hydra.compose` (this might depend on the specific `cmd` though, because different commands load different configs, so it's tricky, ideally you'd have to find a way to validate without copy-pasting all pairs "train -> train_job", "sweep -> sweep_job")
- do two loops, one over Hydra's native "config groups" (`agent`, `trainer`, `user`), another over `key.name=/path/to/config` that are loaded with `config_from_path`

devops/skypilot/utils.py
```diff
@@ -34,3 +39,236 @@ def launch_task(task: sky.Task, dry_run=False):
...
+            trainer_value = task_arg.split("=", 1)[1]
+            config_files_to_check.append((task_arg, f"./configs/trainer/{trainer_value}.yaml"))
+
+        # Check for environment configuration
+        elif task_arg.startswith("trainer.curriculum="):
+            env_value = task_arg.split("=", 1)[1]
+            config_files_to_check.append((task_arg, f"./configs/{env_value}.yaml"))
+
+        # Check for evaluation configuration
+        elif task_arg.startswith("trainer.sim="):
```
berekuk, line 88: Are you sure we have `trainer.sim`? I can't find it in any configs or code

devops/skypilot/launch.py
rjwalters: the `--run` parameter is now positional. I used to have it be last so that it was faster to change my run id in the terminal but I can't do this anymore. The use case is: 

- launch a job with name "runid.0"
- make and push a small change
- launch the same job with name "runid.1"

open to alternatives for names or for making --run a named parameter again so that I can have it at the end of the command...

devops/skypilot/launch.py
rjwalters: epilog adds those final two lines at the bottom:
```
Usage: your_script.py <cmd> <run_id> [task_args...] [options]
Example: your_script.py train my_exp trainer=basic --timeout-hours=2 --id_suffix=001
```
happy to remove if you don't like that

devops/skypilot/utils.py
rjwalters: good catch -- I'll remove

devops/skypilot/utils.py
rjwalters: must have lost the reference to metta.util.git at some point -- thanks

devops/skypilot/launch.py
daveey: how about we support run=foo, this will make launch.py consistent with train.py, etc...

devops/skypilot/utils.py
daveey: can we reduce the copy-pasta:

required_configs = ["agent", "trainer", "trainer.curriculum", ...]
....
if value.starts_with("/") it's a path, otherwise it's relative

metta/util/git.py
berekuk, line 1: These changes are incompatible with `aws/batch/launch_task.py`

berekuk (changes requested):

metta/util/git.py
rjwalters: good catch - I accidentally dropped some functions that we are using in the deprecated scripts

PR #915, opened 2025-06-12
Create remote stats server and client

Barraketh:
Create remote stats server and client

Cleanup


Barraketh:
* **#915** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/915?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/915?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

metta/app/stats_repo.py
```diff
@@ -0,0 +1,227 @@
...
+from psycopg.rows import TupleRow
+from psycopg.types.json import Jsonb
+
+from metta.util.schema_manager import SqlMigration, run_migrations
+
+# This is a list of migrations that will be applied to the eval database.
+# Do not change existing migrations, only add new ones.
+EVAL_MIGRATIONS = [
+    SqlMigration(
+        version=0,
```
berekuk, line 14: Since all migration versions must be unique, I expect this module to be the only one that lists all migrations.

So I want to question the naming:
1) `EVAL_MIGRATIONS` suggests that these are for evals only, but in practice I expect that the DB will contain all metta-related data, not just evals ("Eval infrastructure" name in asana also didn't make sense to me, I think either I'm missing some language conventions that we have here, or it's just a legacy that no one bothered to question)
2) `StatsRepo` is also confusing to me, if this is the entrypoint to our entire database; how about `MettaDb` or something like that?

metta/app/stats_repo.py
```diff
@@ -0,0 +1,227 @@
...
+        description="Initial eval schema",
+        sql_statements=[
+            """CREATE TABLE training_runs (
+                id SERIAL PRIMARY KEY,
```
berekuk, line 18: A decision that's better to make early: do we want serial PKs, or uuids?

I'd vote for uuids, or [cuid2](https://github.com/gordon-code/cuid2): they're easier to debug, e.g. if you see the id=123 in logs but it's missing in the database then it's easier to tell if that's a code bug (someone used the wrong id from another table) or something else. This is even more important if we're going to have many databases - one shared and possibly local dbs for dev.

metta/app/stats_repo.py
```diff
@@ -0,0 +1,227 @@
...
+        ],
+    ),
+]
+
+
+class StatsRepo:
+    def __init__(self, db_uri: str) -> None:
+        self.db_uri = db_uri
+        with Connection.connect(self.db_uri) as con:
+            run_migrations(con, EVAL_MIGRATIONS)
```
berekuk, line 80: I'm not sure I'm 100% comfortable with running migrations optimistically on start, but we can try and see how often it breaks.

metta/app/stats_repo.py
```diff
@@ -0,0 +1,227 @@
...
+    def connect(self) -> Connection:
+        return Connection.connect(self.db_uri)
+
+    def query(self, query: Query, params: Tuple[Any, ...] = ()) -> Sequence[TupleRow]:
+        with self.connect() as con:
+            return con.execute(query, params).fetchall()
+
+    def execute(self, query: LiteralString, params: Tuple[Any, ...] = ()) -> None:
+        with self.connect() as con:
+            con.execute(query, params)
```
berekuk, line 91: You don't use `.query(...)` and `.execute(...)` anywhere, afaict

metta/app/stats_repo.py
```diff
@@ -0,0 +1,227 @@
...
+            return {}
+
+        with self.connect() as con:
+            res = con.execute(
+                """
+                SELECT id, name FROM policies WHERE name = ANY(%s)
+                """,
+                (policy_names,),
+            ).fetchall()
+            return {row[1]: row[0] for row in res}
```
berekuk, line 104: On all the following methods in this class: I don't understand the separation of concerns between this file and `dashboard_routes.py`, some SQL queries are here and some are there, can we consolidate somehow?

metta/app/dashboard_routes.py
```diff
@@ -0,0 +1,188 @@
...
+    params = (suite, metric, group) if group != "" else (suite, metric)
+
+    with con.cursor(row_factory=class_row(GroupDataRow)) as cursor:
+        cursor.execute(query, params)
+        return cursor.fetchall()
+
+
+def create_dashboard_router(stats_repo: StatsRepo) -> APIRouter:
+    """Create a dashboard router with the given StatsRepo instance."""
+    router = APIRouter(prefix="/observatory", tags=["observatory"])
```
berekuk, line 84: Can we either use `prefix="/dashboard"`, or rename to `observatory_routes.py`, etc.?

(I think I prefer the former; I expect observatory to be bigger than the dashboard, so far it's the codename for all our potential web frontend)

berekuk (changes requested):
I consider most of my comments to be minor, feel free to address/ignore however you want.

I'm still doing "request changes" because you'll want to address at least some of the comments, and broken tests is pretty critical. (I also think you should at least consider the uuid comment, the costs of changing it later will be high, but if you disagree and strongly prefer sequential keys it's also fine and I won't fight you too much about this)

metta/app/stats_client.py
```diff
@@ -0,0 +1,176 @@
...
+    PolicyEpochResponse,
+    PolicyIdResponse,
+    PolicyResponse,
+    TrainingRunCreate,
+    TrainingRunResponse,
+)
+
+
+class StatsClient:
+    """Client for interacting with the stats API."""
```
berekuk, line 18: One more nitpick on naming: are you sure we want "stats client" and not the generic "metta app client"?

metta/app/config.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+
+def env_var(name: str, default: str) -> str:
+    value = os.getenv(name)
+    if value is None:
+        return default
+    return value
```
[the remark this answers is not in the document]
Barraketh: Oh, I didn't notice this - will fix

metta/app/config.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+stats_db_uri = env_var("STATS_DB_URI", "postgres://postgres:password@127.0.0.1/postgres")
```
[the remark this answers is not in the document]
Barraketh: I'm gonna add a readme on how to run this locally, and it'll include this URI

metta/app/config.py
```diff
@@ -0,0 +1,14 @@
...
+
+host = env_var("HOST", "0.0.0.0")
```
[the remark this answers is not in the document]
Barraketh: Sure, no problem

metta/app/dashboard_routes.py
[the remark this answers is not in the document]
Barraketh: Ok

metta/app/dashboard_routes.py
Barraketh: Sure

metta/app/stats_repo.py
Barraketh: This was probably premature optimization on my part - I've read about UUIDs being less efficient in indexes (because they are so spread out), and was concerned about joins with the episode_agent_metrics table, which is gonna get pretty large. I'm open to switching to UUIDs, and tackling the performance issue if/when we ever hit it. @daveey any preference?

metta/app/stats_repo.py
Barraketh: I mean, generally it should be a noop, but i'm happy to make this somehow more explicit

metta/app/stats_repo.py
Barraketh: Fair - that was holdover from older code

metta/app/stats_repo.py
berekuk: They are less efficient, but not by that much. If performance is a concern, we can also make a fake `pk SERIAL PRIMARY KEY` but never refer to it in code, use a separate `id TEXT`.

tests/app/test_stats_server.py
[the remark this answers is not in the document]
Barraketh: I'm genuinely not sure how to debug this, since it works both on my machine and on CI. Any chance you could try 'uv remove psycopg' and 'uv add psycopg[binary]' and see if that works for you? I have no issues with using psycopg[binary] instead - just got no idea if that's the actual problem on your machine

metta/app/stats_repo.py
Barraketh: Yeah, I guess i haven't been too consistent with the naming. I do think that it's possible that we'll have other services / migrations, but happy to rename to generic MIGRATIONS / MettaDb for now, and if in the future we need to change it we can

tests/app/test_stats_server.py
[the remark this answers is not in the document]
berekuk: `uv add psycopg[binary]` works.
Or, at least, makes it fail in a new way...
Screenshot 2025-06-12 at 9 11 08 PM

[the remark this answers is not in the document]
Barraketh: Do you have docker running?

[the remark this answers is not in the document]
berekuk: Yes, `docker ps` and `docker image ls` and so on work
(I don't use the docker app though, I usually run [`colima`](https://github.com/abiosoft/colima)` start --foreground` in a terminal tab when I need docker)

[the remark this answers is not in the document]
berekuk: Ok, I've looked into this... colima is the real docker server, but its socket is in `~/.colima/default/docker.sock` instead of `/var/run/docker.sock`. Docker CLI tool works because it knows about docker contexts:

```
$ docker context ls
NAME       DESCRIPTION                               DOCKER ENDPOINT                                     ERROR
colima *   colima                                    unix:///Users/berekuk/.colima/default/docker.sock
default    Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
```

But Docker python lib doesn't.

This snippet fixes it:

```python
@pytest.fixture(scope="session", autouse=True)
def setup_context_endpoint():
    try:
        result = subprocess.run(
            ["docker", "context", "ls", "--format", "{{ if .Current }}{{ .DockerEndpoint }}{{ end }}"],
            capture_output=True,
            text=True,
            check=True,
        )

        endpoint = result.stdout.strip()
        if not endpoint:
            return

        os.environ["DOCKER_HOST"] = endpoint
    except subprocess.CalledProcessError:
        pass
```

With this, testcontainers starts the postgres container, but it fails with another error now:

```
E           psycopg.OperationalError: connection failed: connection to server at "127.0.0.1", port 32775 failed: could not receive data from server: Connection refused
E           Multiple connection attempts failed. All failures were:
E           - host: 'localhost', port: '32775', hostaddr: '::1': connection failed: connection to server at "::1", port 32775 failed: could not receive data from server: Connection refused
E           - host: 'localhost', port: '32775', hostaddr: '127.0.0.1': connection failed: connection to server at "127.0.0.1", port 32775 failed: could not receive data from server: Connection refused
```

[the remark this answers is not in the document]
berekuk: Ugh, alright, after spending another half an hour on this, it's a colima issue too. They bind the port with some delay or something like that ([link 1](https://github.com/testcontainers/testcontainers-python/issues/553#issuecomment-2070874682), [link 2](https://github.com/testcontainers/testcontainers-python/issues/802)). I'll switch to docker desktop I guess (though I don't like it because it's semi-proprietary with a sketchy license).

Btw, what are we going to do for everyone else who's going to run tests? I like to be able to run `pytest tests/` locally if I did big changes, it can be more convenient than waiting for the CI.

Update `devops/macos/Brewfile`, I guess? And maybe do a detection whether docker is present in test and warn/skip if it isn't?

Barraketh:
Addressed most of the requests

berekuk (approved):

PR #920, opened 2025-06-12
be consistent about using self._task

sasmith:
Removes a redundant parameter. Since `self._task` is already available as a class member, there's no need to pass it as an argument to the `_initialize_c_env` method. Moreover, passing it adds (or suggests) that the parameters might be different.

There's more cleanup we could do, like the fact that `labels` are read only off the first task. And it's not great that you need to remember to set a new `_task` before calling init, and that it's done in 2 places. But baby steps.

### DIFF
--- a/mettagrid/mettagrid/mettagrid_env.py
+++ b/mettagrid/mettagrid/mettagrid_env.py
@@ -81,7 +81,7 @@ class MettaGridEnv(pufferlib.PufferEnv, gym.Env):
         self.labels = self._task.env_cfg().get("labels", None)
         self._should_reset = False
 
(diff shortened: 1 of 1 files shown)


sasmith:
* **#920** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/920?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/920?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #922, opened 2025-06-12
Write remote stats during simulation

Barraketh:
PR 2 of 3 - use a stats client to write stats to our stats webapp

PR #923, opened 2025-06-12
Use server for dashboard

Barraketh:
Update dashboard FE to use the webapp apis to load data.

Will not actually merge this until we figure out deployment


Barraketh:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/923?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#923** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/923?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/923?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#922** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/922?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#915** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/915?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

.gitignore
```diff
@@ -39,5 +39,6 @@ cpplint_output*
 
 observatory/node_modules/
 observatory/.vite/
+observatory/dist/
 
 wandb
```
berekuk, line 44: btw, can you remove this line while you're at it? (we have `/wandb` at line 26)

.gitignore
```diff
@@ -39,5 +39,6 @@ cpplint_output*
 
 observatory/node_modules/
 observatory/.vite/
+observatory/dist/
```
berekuk, line 42: Mild bikeshedding preference for per-package ignores when they get this specific.

Rationale: top-level dir shouldn't know that observatory is powered by vite. I think `node_modules` ignore is safe on the top level, as is any other name that's so specific that we're sure we won't have it anywhere, e.g. `.vite`. But when things get specific, `observatory/dist`, then it's observatory package's responsibility to say "we're ignoring X, Y and Z here". Ignoring any `dist` dir is _probably_ safe, though on another project I had `src/dist` that was for statistical distributions. Other ignores in this file without leading `/` also make me a bit nervous.

observatory/src/App.tsx
```diff
@@ -11,58 +10,42 @@ function App() {
...
-      setState({ type: "repo", repo: new DataRepo(data) });
-    } catch (err: any) {
-      setState({
-        type: "default",
-        error: "Failed to load data from file: " + err.message,
-      });
-    }
-  };
-
   if (state.type === "default") {
```
berekuk, line 49: "default" state after the refactoring above now always has `error` set, so it should be changed to `{ type: "error", error: string }`.

berekuk (approved):

observatory/src/Heatmap.tsx
```diff
@@ -1,13 +1,13 @@
-import { useState } from 'react'
-import Plot from 'react-plotly.js'
-import { HeatmapData } from './repo'
+import { useState } from "react";
+import Plot from "react-plotly.js";
+import { HeatmapData } from "./repo";
```
berekuk, line 3: Many of these changes revert Robb's biome linting from #995. I assume this is because you use Prettier vscode extension.

I don't really care if we use prettier or biome, but it would be good to avoid back-and-forths on this.

At the same time, if we're switching to biome, I'd like us to configure it so that we have training semicolons, at least (automatic semicolon insertions are bad). I care less about double-quotes vs single-quotes but lean toward double.

Barraketh: You're right - i'll reformat using biome

PR #924, opened 2025-06-12
Randomize first episode length

daveey:
## Summary
- tweak MettaGridEnv to randomize the first episode's length when training

## Testing
- `uv run ruff check mettagrid/mettagrid/mettagrid_env.py metta/sim/vecenv.py metta/rl/pufferlib/trainer.py`
- `PYENV_VERSION=3.11.12 uv run mypy mettagrid/mettagrid/mettagrid_env.py metta/sim/vecenv.py metta/rl/pufferlib/trainer.py` *(fails: missing stubs)*
- `uv run pytest -q`

------
https://chatgpt.com/codex/tasks/task_b_684adc352370832bbe455b0d45540b49


metta/sim/vecenv.py
```diff
@@ -21,6 +21,7 @@ def make_env_func(
     render_mode="rgb_array",
     stats_writer: Optional[StatsWriter] = None,
     replay_writer: Optional[ReplayWriter] = None,
+    training: bool = False,
```
sasmith, line 24: I don't think this should have a default value. If we're going to have a `training` param, it should be required.

sasmith (changes requested):
I feel okay-not-great about the intended implementation. I'd rather get there by (e.g.) puffer holding back some of the envs that it makes, to have them not start until later.

I.e., I think we have about 2x the environments we need in order to fill a mini batch. We could start by initializing 1x of these, and using them for training, and then (as CPU permits?) initializing others and having them join the mix / queue. We could also do something like bin them based on % of steps complete, and bias our picking of which envs to step forward in a way to try to get uniform bins.

A couple of reasons this feels off to me:
* The environments are, in a sense, incomplete. I may be over-weighting this, given that we definitely stop environments abruptly when we feel like it, and this is roughly the same.
* We pay the full cost for these environments -- may as well get use out of them.

metta/sim/vecenv.py
daveey: i got rid of this completely

daveey:
This is only for the first episode. We throw away a little data to de-sync the envs.

We could look into doing this in vecenv but that's more work

daveey:
Fwiw the standard approach is to step the first envs some number of times

PR #935, opened 2025-06-13
Cleanup training environments

daphnedemekas:


relh (approved):
the navigation curriculum changing ratios will change what it means to do navigation but a move towards a more curated better set of tasks is great.

I'm guessing the max_steps 2000 change in terrain_from_numpy might also change what's possible in those envs.

I like the fixed altar amount as a default, I think soon we're going to be pulling sampling out of yaml anyways and putting it into curriculum which I like.

I'm personally a fan of the all the all.yamls.. I think we should try and:
- train our models on an all curriculum
- sim on an all sequence (like the configs/sim/all.yaml)
- do everything for benchmarking on an all sequence

Basically, soon we can have a github CI smoke test which trains on ALL then sims on ALL and then flags any specific evals that have tanked (maybe we already have this?)

PR #943, opened 2025-06-13
Move global tokens into compute_observation

daveey:
### TL;DR

Refactored observation computation in MettaGrid to pass action parameters directly and moved ActionArg type definition to types.hpp.


mettagrid/mettagrid/mettagrid_c.cpp
```diff
@@ -236,8 +237,27 @@ void MettaGrid::_compute_observation(unsigned int observer_row,
   // we don't need to do that here.
   if (_use_observation_tokens) {
     size_t attempted_tokens_written = 0;
```
sasmith, line 239: This should start at `tokens_written`.

sasmith (approved):
I realized this is going to mess up normalization, since that's handled by the observation encoder, and this works outside of that. I'll follow up.

PR #947, opened 2025-06-13
Show Recipe in the hover panels.

treeform:
Add game config to the replay file format.
Add recipes to the hover panels.
Rename info panels to hover panels.

image


treeform:
The recipe shows the number of times required and number of times produced? The inventory is actual numbers? Maybe we can double them up some how...

PR #956, opened 2025-06-13
cp

daveey:


sasmith (approved):
I assume this is because `num_workers` is poorly named, and we actually run `num_workers` for each GPU?

PR #661, opened 2025-05-25
Add heartbeat monitor for SkyPilot jobs

daveey:
## Summary
- add `metta.util.heartbeat` to update and monitor a heartbeat file
- start the heartbeat monitor in `devops/train.sh`
- write heartbeats from `tools/train.py`
- source `.venv/bin/activate` in `devops/train.sh`

## Testing
- `ruff format metta/util/heartbeat.py tools/train.py`
- `ruff check metta/util/heartbeat.py tools/train.py`
- `pytest -q`

PR #957, opened 2025-06-13
Self destroy jobs that got stuck and inform user

dominik3141:
**Overview**
We modify the heartbeat system introduced in #661 to require manual calls to the heartbeat recording function.
If no heartbeat has been detected for a certain amount of time, `wandb.alert()` is called which notifies the user of this timeout before killing the training process.

**Technical internals**
Allowing the `wandb.alert()` to be send was quite challenging as the heartbeat monitor lives in another process and can't just access the same wandb run object. To solve this problem we use IPC over file to share the wandb run_id and other internal with the monitor process.


metta/rl/pufferlib/trainer.py
```diff
@@ -225,6 +227,25 @@ def __init__(
 
         logger.info(f"PufferTrainer initialization complete on device: {self.device}")
 
+        # W&B Stuck Run Detection - Initialize attributes
+        self.last_wandb_log_time = time.time()  # Start with current time
```
daveey, line 231: _last

metta/rl/pufferlib/trainer.py
```diff
@@ -225,6 +227,25 @@ def __init__(
...
+        self._original_wandb_log_fn = None
+        self.wandb_stuck_monitor_thread = None
```
daveey, line 233: _wandb

metta/rl/pufferlib/trainer.py
```diff
@@ -225,6 +227,25 @@ def __init__(
 
         logger.info(f"PufferTrainer initialization complete on device: {self.device}")
 
+        # W&B Stuck Run Detection - Initialize attributes
```
daveey, line 230: can we make this into stuck_run_detector.py:StuckRunDetector()

metta/rl/pufferlib/trainer.py
```diff
@@ -225,6 +227,25 @@ def __init__(
...
+        self.last_wandb_log_time = time.time()  # Start with current time
+        self._original_wandb_log_fn = None
+        self.wandb_stuck_monitor_thread = None
+
+        if self.wandb_run and self._master:
```
daveey, line 235: why use wandb instead of a tmpfile?

dominik3141: wandb.log calls are a good indication of whether a job is actually running or just in a silent error state

metta/rl/pufferlib/trainer.py
```diff
@@ -927,6 +948,67 @@ def _make_vecenv(self):
...
+    def _monitor_wandb_activity(self):
+        # Fixed check interval: 60 seconds
+        check_interval = 60
+        # Read from config with a default of 1800 seconds (30 minutes)
+        wandb_inactivity_timeout_seconds = self.trainer_cfg.get("wandb_inactivity_timeout_seconds", 1800)
+
+        logger.info(
+            f"W&B monitor thread started. Rank: {rank}, Local Rank: {local_rank}. "
+            f"Check interval: {check_interval}s, Max log inactivity timeout: {wandb_inactivity_timeout_seconds}s."
+        )
```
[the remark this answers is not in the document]
dominik3141: Good idea!

metta/rl/pufferlib/trainer.py
```diff
@@ -927,6 +948,71 @@ def _make_vecenv(self):
         rank = int(os.environ.get("RANK", 0))
         self.vecenv.async_reset(self.cfg.seed + rank)
 
+    # Method to wrap wandb.log and update timestamp
+    def _wandb_log_with_timestamp_update(self, data: dict, **kwargs):
```
daveey, line 952: let's just call  self._heartbeat_helper.heartbeat() in trainer.py:train(), wrapping wandb works, but also ties heartbeats to wandb unnesseraily

metta/rl/pufferlib/trainer.py
```diff
@@ -927,6 +948,71 @@ def _make_vecenv(self):
...
+                            level=AlertLevel.ERROR,
+                        )
+                    except Exception as e:
+                        # This will catch errors if .alert() is missing, not callable, or fails internally
+                        logger.error(f"Failed to send wandb.alert for stuck run: {e}", exc_info=True)
+
+                logger.info("Attempting to close vecenv before forced exit due to W&B inactivity...")
+                try:
+                    if hasattr(self, "vecenv") and self.vecenv is not None:
+                        self.vecenv.close()
```
berekuk, line 1008: `self.vecenv` is always set by this point, and you could call `self.close()` instead.

metta/rl/pufferlib/trainer.py
```diff
@@ -927,6 +948,71 @@ def _make_vecenv(self):
...
+                )
+                logger.critical(alert_message)
+
+                if self.wandb_run:
+                    try:
+                        self.wandb_run.alert(
+                            title="Critical: Training Run Stuck",
+                            text=alert_message,
+                            level=AlertLevel.ERROR,
+                        )
```
berekuk, line 1000: I suspect that our existing jobs got stuck because of wandb outage; we have "failed sending: POST https://api.wandb.ai/graphql giving up after 21 attempt(s)" at the end of their logs.

So I don't fully trust `wandb_run.alert()` to finish in this situation either, if it's not wrapped in a careful timeout.

berekuk (changes requested):
I'm not sure that relying on `wandb.run()` calls is the right way to do monitoring, and I don't love the monkey-patching of a method.

I'd suggest looking for some other way to monitor, like updating `self._last_meaningful_activity` in a trainer code in some hot function, e.g. on every epoch, and relying on that in the monitoring thread.

This is not a blocker for me though; the reason I'm "requesting changes" because `wandb_run.alert()` would probably make this safety net malfunction in the situation that caused you to do this PR, and you'll probably want to handle that before the merge.

---

Another observation is that we have `metta.util.heartbeat` that David did in #661.

I wonder if we could modify it so that trainer would have to actively call `heartbeat()` from time to time, instead of `start_heartbeat` blindly updating the timestamp. This might be complicated though, I'm not sure if it's worth it.

(I've checked the stuck jobs and they do update the heartbeat.)

metta/rl/pufferlib/trainer.py
dominik3141: Good point, I will wrap it in a timeout

dominik3141:
> > We could also build something like a manual heartbeat function, but when I think where I would call that function, it would most likely be in just the same places in which we already call `wandb.log()`.
> 
> Yes, the manual heartbeat function is what I was thinking of. I would expect that there's just one place where you could call the heartbeat function next to `wandb.log()`, it would achieve the same goal and be less tightly coupled.

Ok, I gave up on the `wandb.log()` approach and now explicitly call a heartbeat function. Do you like this approach better? @berekuk

berekuk (approved):
Sorry for the delay!

I was surprised about IPC stuff, though it makes sense, and wanted to figure out an easier way to achieve the same thing (I think it should be doable, maybe if you look for the dir that contains the heartbeat file and deriving wandb metadata from it).

I don't like that this PR got stuck in review review limbo though, so I'm approving.

PR #824, opened 2025-06-05
hotfix: make docker logs wider

rjwalters:
### Problem
Currently, our training logs are being split across multiple lines when viewed in AWS environments (CloudWatch, AWS Console), making them difficult to read:

```
6685 2025-06-05 17:45:00  [10:45:00.167979] INFO    Epoch 770 - rollout: 1.110s, train:    trainer.py:249
6686 2025-06-05 17:45:00                            0.598s, stats: 0.001s, [118588610
6687 2025-06-05 17:45:00                            steps/sec]
```

This happens because the default terminal width in AWS Batch containers is too narrow (typically 80 columns), causing the logging output to wrap.

### Solution

```python
    # set env COLUMNS if we are on AWS
    if os.environ.get("AWS_BATCH_JOB_ID"):
        os.environ["COLUMNS"] = "200"
```

This function is called during logger setup with a width of 200 characters, which provides enough space for our log lines plus ~45 characters of metadata (timestamps, log levels, file info).

### Result
Logs now display properly on a single line in AWS environments:
```
[10:45:00.167979] INFO    Epoch 770 - rollout: 1.110s, train: 0.598s, stats: 0.001s, [118588610 steps/sec]
```

### Technical Details
- Only activates when `AWS_BATCH_JOB_ID` environment variable is present
- Sets `COLUMNS` environment variable to influence terminal-aware applications
- No impact on local development or non-AWS environments

### Additional Changes
- Updated Docker entrypoint and AWS Batch job definitions to use login shells (`/bin/bash -l -c`)
- Enhanced `build.sh` script with better error handling and Docker installation checks
- Added platform specification for AWS-compatible linux/amd64 builds

### Impact
- ✅ Fixes log readability specifically in AWS CloudWatch and AWS Batch Console
- ✅ No changes to local development experience
- ✅ Backwards compatible - only affects AWS Batch environments
- ✅ Zero performance impact - minimal overhead only during logger initialization

### Testing
- Verified logs display correctly in AWS Batch environments
- Confirmed no impact on local Docker containers
- Tested that non-AWS environments remain unaffected

PR #964, opened 2025-06-14
fix: make log lines wider on skypilot

rjwalters:
The changes in #824 seem ineffective on skypilot.

Debug prints show:

```
[12:08:04.496446] INFO     Terminal Information:                
[12:08:04.497338] INFO     stdin isatty: False                   
[12:08:04.498064] INFO     stdout isatty: False            
[12:08:04.498781] INFO     stderr isatty: False          
[12:08:04.499530] INFO     Terminal size: 80x24       
[12:08:04.500305] INFO     COLUMNS env: Not set    
[12:08:04.501004] INFO     LINES env: Not set        
[12:08:04.501711] INFO     AWS Environment Variables:    
[12:08:04.502556] INFO     No AWS environment variables found  
```

I think the easiest is to to set COLUMNS based on `SKYPILOT_TASK_ID` presence too.

testing:

image

### DIFF
--- a/metta/util/logging.py
+++ b/metta/util/logging.py
@@ -27,7 +27,7 @@ def restore_io():
 
 # Create a custom formatter that supports milliseconds
 class MillisecondFormatter(logging.Formatter):
-    def formatTime(self, record, datefmt=None):
+    def formatTime(self, record: logging.LogRecord, datefmt: str | None = None) -> str:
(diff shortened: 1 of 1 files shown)


berekuk (approved):
I continue to think that a different logger config for the cloud would be better, but this is clearly an improvement.

PR #972, opened 2025-06-16
get rid of feature_names()

sasmith:
Generate offsets directly using feature_ids, instead of implicitly based on feature_name. This lets us remove feature_names.

An important change here is that this also switches normalizations to be by feature_id as well. Previously they were generated based on the order in which feature_names were encountered. This was related to, but not that same as, the order given by feature_id.

I think the normalization _happened_ to be right, prior to this change (and global features); but it was mild accidental, and definitely fragile.

Note that I'm trying to fix this for existing box observations rather than ripping out box observations, since I want us to have something correct to fall back to.

---
  - https://app.asana.com/0/0/1210550931090512


mettagrid/mettagrid/objects/agent.hpp
```diff
@@ -127,32 +127,31 @@ class Agent : public MettaObject {
...
-    obs[offsets[3]] = hp;
-    obs[offsets[4]] = frozen;
-    obs[offsets[5]] = orientation;
-    obs[offsets[6]] = color;
-
-    for (int i = 0; i < InventoryItemCount; i++) {
-      obs[offsets[7 + i]] = inventory[i];
+  virtual void obs(ObsType* obs) const override {
+    const auto offsets = Agent::offsets();
+    obs[offsets[0]] = _type_id;
```
daveey, line 132: now that we're in cpp, let's do ++offset instead of 1,2,3..

mettagrid/mettagrid/objects/constants.hpp
```diff
@@ -136,37 +136,35 @@ const std::vector<std::string> ObservationFeatureNames = []() {
...
-    {"inv:laser", 100.0},
-    {"inv:armor", 100.0},
-    {"inv:blueprint", 100.0},
-    {"agent:kinship", 10.0},
-    {"hp", 30.0},
-    {"converting", 1.0},
-    {"color", 10.0},
-    {"swappable", 1.0},
-    {"type_id", 10.0},
+const std::map<uint8_t, float> FeatureNormalizations = {
```
daveey, line 139: fine for now, but the plan is to move these into configs and out of constants. so changing strings to constants may be a move in the wrong direction

daveey (approved):
approved, but consider the comments

mettagrid/mettagrid/objects/constants.hpp
sasmith: Understood. Right now, global observations broke normalization in a way that's interfering with the token -> grid model. I'm trying to fix that so I can get us to just tokens.

PR #839, opened 2025-06-06
v0 new trainer

relh:

PR #973, opened 2025-06-16
MettaAgent revisioning to wrap PolicyRecord and Policy v0 

relh:
These were the goals:

Implement a process that allows us to update agent parts without breaking old agents as much as possible. Document the process in a readme.md under metta/agent. Change PAW to NonMettaAgentWrapper. Document goals for this also in the readme.md.

> Either:
> MettaAgent becomes wrapper for BrainPolicy or PufferPolicy
> Delete PolicyRecord, we only ever save or load MettaAgents
> MettaAgents are inextricably linked from their policies, when we initialize a new MettaAgent we can load a policy URI but that's the last time it can be hotswapped
> 
> OR:
> Add PufferAgent as a type of Component
> Only create MettaAgents
> 
> OR:
> Make MettaAgent support things other than components
> MettaAgent becomes complicated
> 
> OR:
> Keep PolicyRecord
> Policies are fungible and can go into any type of agent
> Once built policies are just different types of neural networks
> 
> OR: 
> environment/action space versioning matters
> everything stores the space that matters
> if loaded with new type of environment/etc, uses thunking layer to adapt

This PR makes MettaAgent a wrapper, deletes policy record, and we only ever save or load MettaAgents. 

We also switch to jit tracing/script to ensure that we can save/load without the class staying the same.

We also make the normal construction of agents a new class called BrainPolicy. We're setting up to change PufferPolicy to something more akin to ExternalPolicy.


relh:
Going to wait for this to merge before finalizing our MettaAgent structural changes to support revisioning:

https://github.com/Metta-AI/metta/pull/839

PR #975, opened 2025-06-16
fix: drop "puffer" from our trainer namespace

rjwalters:
Our trainer has forked pretty far away from PufferLib and should get it's own name. This will also help us to differentiate when we want to train policies using the vanilla PufferLib code.
    

---

## Summary by cubic
Renamed our trainer from PufferTrainer to MettaTrainer and removed the pufferlib namespace to reflect code changes and reduce confusion with PufferLib.

- **Refactors**
  - Updated all imports and config references from pufferlib to the new structure.
  - Removed the old pufferlib module from our codebase.


metta/rl/policy.py
```diff
@@ -28,11 +28,18 @@ def load_policy(path: str, device: str = "cpu", puffer: DictConfig = None):
...
     policy = instantiate(puffer, env=env, policy=None)
     policy.load_state_dict(weights)
-    policy = PufferAgent(policy).to(device)
+    policy = PytorchAgent(policy).to(device)
     return policy
 
 
-class PufferAgent(nn.Module):
+class PytorchAgent(nn.Module):
+    """Adapter to make loaded Puffer policies compatible with MettaAgent interface.
```
relh, line 36: we could say "torch.nn.Module-based policies" if we want but not needed

relh (approved):

relh (approved):

PR #980, opened 2025-06-16
fix: accurate loss reporting

rjwalters:
### Problem

We have been reporting incremental averages during loss accumulation:

```python
self.losses.policy_loss += pg_loss.item() / total_minibatches
```

This approach will lead to **Incorrect averages on early exit**: 

When training exits early due to KL divergence threshold, we've divided by total_minibatches but only accumulated a subset

### Solution

This PR changes the loss tracking to:
1. Accumulate raw loss values without division
2. Track the actual number of minibatches processed
4. Compute averages only when needed in `to_dict()`

### Changes

- Modified `Losses` class to store cumulative sums (`*_sum` attributes)
- Added `minibatches_processed` counter
- Updated `to_dict()` to compute true averages: `sum / actual_minibatches_processed`
- Fixed KL divergence check to use the running average instead of the last minibatch value

    

---

## Summary by cubic
Fixed loss averaging to ensure accurate training stats and correct KL-based early stopping, even when training exits early.

- **Bug Fixes**
  - Loss values now accumulate raw sums and track the number of minibatches processed.
  - Averages are computed only when needed, using actual minibatch counts.
  - KL divergence check now uses the running average, not just the last minibatch.

### DIFF
--- a/metta/rl/losses.py
+++ b/metta/rl/losses.py
@@ -3,19 +3,39 @@ class Losses:
         self.zero()
 
     def zero(self):
-        """Reset all loss values to 0.0"""
-        self.policy_loss = 0.0
-        self.value_loss = 0.0
-        self.entropy = 0.0
-        self.old_approx_kl = 0.0
-        self.approx_kl = 0.0
-        self.clipfrac = 0.0
+        """Reset all loss values and counters"""
+        # Cumulative values
+        self.policy_loss_sum = 0.0
+        self.value_loss_sum = 0.0
+        self.entropy_sum = 0.0
+        self.old_approx_kl_sum = 0.0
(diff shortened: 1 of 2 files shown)


dominik3141:
Have you done some empirical testing? How much did the stats change?

rjwalters:
> Have you done some empirical testing? How much did the stats change?

image

No - but I don't think we should expect to see a difference unless we are exiting early via

```

            if self.trainer_cfg.target_kl is not None:
                average_approx_kl = self.losses.approx_kl_sum / self.losses.minibatches_processed
                if average_approx_kl > self.trainer_cfg.target_kl:
                    break
                    
```

PR #901, opened 2025-06-10
feat: environment timing detail

rjwalters:
moves stopwatch utility to mettagrid and implements additional logging to wandb

PR #982, opened 2025-06-16
fix: mv stopwatch to mettagrid

rjwalters:
(#901 is getting too big so I am splitting off code in incremental PRs)

This PR moves the Stopwatch from metta utils to mettagrid utils so that we can use it in mettagrid without a dependency on the parent package.

As part of the move, the Stopwatch class has been significantly refactored with the following improvements:

### 1. **Improved Architecture**
- Converted from dictionary-based storage to dataclass-based `Timer` objects
- Added proper TypedDict definitions for `TimerReference` and `Checkpoint`
- Cleaner separation of concerns with a dedicated Timer dataclass

### 2. **New Decorators for Easy Integration**
- Added `@with_timer` decorator for timing functions with external Stopwatch instances
- Added `@with_instance_timer` decorator for timing class methods using instance-level timers
- Both decorators support automatic logging and proper exception handling

### 3. **Enhanced Functionality**
- **Auto-incrementing lap counter**: Checkpoints/laps now support automatic step counting when steps aren't provided
- **File reference tracking**: Automatically captures and stores where each timer is used (filename + line number)
- **Bulk operations**: New `checkpoint_all()` and `lap_all()` methods for recording checkpoints across multiple timers
- **Better lap tracking**: New `get_lap_steps()` method to retrieve step counts for specific laps

### 4. **Unit Test Suite**
- Tests cover basic functionality, edge cases, decorators, and real-world scenarios
- Ensures reliability and catches regressions

## Example Usage

```python
# Using decorators
@with_timer(my_stopwatch, "data_processing")
def process_data(data):
    # automatically timed
    return transform(data)

class MyClass:
    def __init__(self):
        self.timer = Stopwatch()
    
    @with_instance_timer("method_timer", log_level=logging.INFO)
    def my_method(self):
        # automatically timed and logged
        pass

# Auto-incrementing laps
stopwatch.start("training")
for epoch in range(10):
    train_epoch()
    stopwatch.lap()  # automatically increments internal counter
```

    

---

## Summary by cubic
Moved the Stopwatch utility from metta to mettagrid, refactored its architecture, and added new features for easier timing and integration.

- **Refactors**
  - Rebuilt Stopwatch using dataclasses and TypedDicts for cleaner code and better type safety.
  - Improved separation of concerns and internal structure.

- **New Features**
  - Added @with_timer and @with_instance_timer decorators for simple function and method timing.
  - Checkpoints now support auto-incrementing lap counters and file reference tracking.
  - Added bulk checkpoint/lap methods and a get_lap_steps helper.
  - Included a full unit test suite for Stopwatch.


sasmith (approved):
Approving since this is a move.

I'm trying to see where the stopwatch fits, since it's doing things like capturing line-numbers, which feels like "I'm a profiler"; but this is higher level and aiming to be always-on. So, I guess I'll be happy to see the output.

PR #984, opened 2025-06-17
Add a shell.nix to make it easier to setup on Linux

monofuel:
- Wanted to get the project running locally on CPU on Linux, made a `shell.nix` to automate setting up the specific version of python and everything.
- I've tested this working on cpu on 2 systems. one is a laptop running NixOS, and also on a Fedora server with Nix installed.

- I don't currently have an Nvidia system to test on, but it should probably work if nvcc is in the environment and working.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,6 +1,6 @@
 let
   pkgs = import (fetchTarball {
-    url = "https://nixos.org/channels/nixos-25.05/nixexprs.tar.xz";
+    url = "https://nixos.org/channels/nixos-24.11/nixexprs.tar.xz";
   }) {
     config.allowUnfree = true;
   };
@@ -22,8 +22,10 @@ pkgs.mkShell {
     cmake
     stdenv.cc.cc.lib
 
+    # nix 24.11 comes with gcc13, which works with nvcc 12
+
     # for mettascope
-    nodejs_24
+    nodejs_22
     typescript
   ];
 
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,59 @@
+let
+  pkgs = import (fetchTarball {
+    url = "https://nixos.org/channels/nixos-25.05/nixexprs.tar.xz";
+  }) {
+    config.allowUnfree = true;
+  };
+
+  flake-compat = import (fetchTarball {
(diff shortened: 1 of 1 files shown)


cab404:
haven't seen anyone doing locking manually on shell.nix in a long time...
you might want to either use niv (or something like that) for locking deps, or flakes if you are not a strong opponent of latter

maybe you have some rationale behind this choice?

monofuel:
I only started using Nix at the start of the year, haven't ran into `niv` yet. The only mention of niv in the official nix docs is that they recommend using npins instead https://nix.dev/guides/recipes/dependency-management#dependency-management-npins . My first thought is usually to go with a simple shell.nix and avoid extra tools.

I tried out both flake and niv locally, I think I like the flake approach more since cachix/nixpkgs-python is also a flake. I've gone ahead and updated the PR

monofuel:
Thanks, that's much nicer!! I saw in cachix/nixpkgs-python it suggested using `cachix use nixpkgs-python` to use a pre-built version, much nicer to see it can be done automatically.

Another option is to just use the dynamically linked python provided by `uv` (I tested that working fine on my Fedora system), but it doesn't [work on nixos by default](https://nix.dev/guides/faq#how-to-run-non-nix-executables). It would be trivial to just enable nix-ld on nixos but I feel like that defeats the spirit of things. I don't have strong opinions.

cab404:
> Another option is to just use the dynamically linked python provided by `uv` (I tested that working fine on my Fedora system), but it doesn't [work on nixos by default](https://nix.dev/guides/faq#how-to-run-non-nix-executables). It would be trivial to just enable nix-ld on nixos but I feel like that defeats the spirit of things. I don't have strong opinions.

technically you can patchelf it, but probably don't. I am not even sure having such granularity of choice is even needed, and maybe just using py from nixpkgs is enough

if you really want, you can use buildFHS from nixpkgs, and create FHS-compatible environment where base uv will be happy to install and run python

sasmith (approved):
My ability to read and parse nix is limited; but based on that, this looks good.

relh:
@zfogg is a nix guy, I've assigned him this to review. Let me know if we need to make more updates.

relh:
@nishu-builder

PR #987, opened 2025-06-17
More consistent TypeScript formatting.

treeform:
* Take a stance on semicolons -> choose the less visual noise non-semicolon style.


rjwalters (changes requested):
I was expecting to see a parameter change in prettier or biome config -- but maybe you are not using either? we should pick one and set it up!

treeform:
Ai tried to setup autoformatter in another branch but it just mangled the code to the point it no longer compiled.
Here i just wanted to get to consistent states as some lines had ; and others did not. I am just using the plain VS code formatter.

PR #989, opened 2025-06-17
fix: replace puffer profiler

rjwalters:
This PR removes the legacy puffer profiler system and replaces it with our internal `Stopwatch` class for performance monitoring. This simplifies the codebase and provides more consistent timing metrics across the trainer.

#### Removed
- **Deleted `metta/rl/profile.py`** - Removed the entire Profile class and associated profiling infrastructure
- **Deleted `tests/rl/pufferlib/` directory** - Removed all puffer-related tests including `test_profile.py`
- Removed all `profile_section` decorators and `with profile.*` context managers throughout the code

#### Modified
- **`metta/rl/trainer.py`**:
  - Replaced all `Profile` usage with `Stopwatch` timing
  - Added `@with_instance_timer` decorators for high-level operations like `_rollout`, `_train`, `_evaluate_policy`, etc.
  - Improved timing metrics reporting with more granular breakdown of operations
  - Enhanced wandb logging to include:
    - Per-epoch timing fractions
    - Cumulative timing fractions
    - Steps per second (both instantaneous and lap-based)
    - Training efficiency metrics


relh (changes requested):
This is great, but the new trainer.py is very re-written. Can you try redoing the changes onto it and I'll review again?

rjwalters:
> This is great, but the new trainer.py is very re-written. Can you try redoing the changes onto it and I'll review again?

was definitely painful to merge. I am running some test training runs to confirm the stats look okay but it should be ready for your review in the interim

metta/rl/trainer.py
```diff
@@ -404,104 +407,112 @@ def _rollout(self):
...
+            vtrace_cfg.get("vtrace_c_clip", 1.0),
+        )
 
         # Optimizing the policy and value network
-        for mb in range(self._total_minibatches):
-            with profile.train_misc:
+        _total_minibatches = experience.num_minibatches * trainer_cfg.update_epochs
+        minibatch_idx = 0
+
+        for _epoch in range(trainer_cfg.update_epochs):
```
relh, line 506: Getting this double loop correct with indexing can be annoying and an easy bug but I like this because it stays closer to what we did before

relh (approved):
thank god you found a way to get rid of all the with statements

PR #1008, opened 2025-06-18
Adding back average reward if we want that feature

relh:
## Summary by cubic
Added support for average reward adjustment in training, allowing the trainer to use an average reward baseline when enabled.

- **New Features**
  - Tracks and updates average reward using an exponential moving average.
  - Loads and saves average reward in checkpoints.
  - Adjusts reward and advantage calculations when the feature is enabled.


rjwalters (commented):
I think we should hold off on adding this until somebody wants to use it.

PR #1011, opened 2025-06-18
Performant token obs encoder

Al-does:
Revised to be performant both in sample efficiency and SPS.


configs/agent/dep_cnn_simple.yaml
```diff
@@ -1,9 +1,11 @@
-# see reference_design.yaml for explanation of components
+# ========= DEPRECATED: use simple.yaml instead =========
```
sasmith, line 1: We're here!

sasmith (approved):
Can we have a `agent/README.md` that gives bullet point descriptions of each agent, and what their differences mean?

relh (approved):
LGTM glad the line count when down

PR #1038, opened 2025-06-18
App docker file

Barraketh:


app_backend/pyproject.toml
```diff
@@ -0,0 +1,25 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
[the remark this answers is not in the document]
berekuk: Or maybe we should migrate wandb_carbs to hatchling, in that case, nevermind.
(I don't know anything about hatchling, looks like it's an officially blessed more modern thing.)

app_backend/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+  "uvicorn[standard]>=0.34.2",
+  "psycopg[binary]>=3.2.9",
+  "pydantic>=2.11.5",
+]
+
+[project.scripts]
+app-backend = "src.server:main"
+
+[tool.hatch.build.targets.wheel]
+packages = ["src"]
```
berekuk, line 22: This looks weird and I think it's wrong.

Suggestion:
- move code to `./app_backend` relative to this file
- or, if you want to keep src, then `./src/app_backend`
- this is because backend serves as a package (we import it in metta), and should be installed normally, in editable mode through workspaces (see my comment on `uv.lock`)

metta/rl/trainer.py
```diff
@@ -15,11 +15,11 @@
 from omegaconf import DictConfig, ListConfig
 from pufferlib import unroll_nested_dict
 
+from app_backend.src.stats_client import StatsClient
```
berekuk, line 18: When you'll switch to workspaces, this will be `from app_backend.stats_client` (will import the installed package from `.venv`, same as we do with mettagrid)

app_backend/build.sh
```diff
@@ -0,0 +1,15 @@
...
+# Generate uv.lock file
+echo "Generating uv.lock file..."
+uv lock
+
+# Build Docker image
+echo "Building Docker image..."
+docker build -t metta-app-backend:latest .
+
+echo "Build complete! You can run the container with:"
+echo "docker run -p 8000:8000 -e STATS_DB_URI=your_db_uri metta-app-backend:latest"
```
berekuk, line 15: Unless your db is remote, this won't work, containers can't access localhost without extra configuration.
(When is this file useful btw?)

berekuk (changes requested):

metta.code-workspace
```diff
@@ -66,48 +66,54 @@
...
+      "istream": "cpp",
+      "sstream": "cpp",
+      "algorithm": "cpp",
+      "list": "cpp",
+      "queue": "cpp",
+      "stack": "cpp",
+      "bitset": "cpp",
+      "deque": "cpp",
+      "regex": "cpp"
+    }
```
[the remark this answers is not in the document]
Barraketh: Yeah, i didn't update this on purpose - I'll take it out of the PR

app_backend/Dockerfile
```diff
@@ -0,0 +1,43 @@
+# Use Python 3.11 slim image as base
+FROM python:3.11-slim
+
+# Install system dependencies
+RUN apt-get update && apt-get install -y \
+    curl \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install uv
+RUN pip install uv
```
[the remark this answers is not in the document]
Barraketh: Ok

tests/app/test_stats_server.py
berekuk, line 1: Should this file be in `app_backend`?
This will require updating all places where we call `pytest` to make sure we run it in all sub-packages, though, so it's fine if we do this later (in a more principled way than copy-pasting it in CI, etc.)

pyproject.toml
```diff
@@ -10,54 +10,54 @@ readme = "README.md"
...
+  "tabulate>=0.9.0",
+  "tensordict>=0.8.3",
+  "termcolor>=2.4.0",
+  "testcontainers>=4.10.0",
+  "torchrl>=0.8.1",
+  "tqdm>=4.67.1",
+  "uvicorn[standard]>=0.34.2",
+  "wandb>=0.19.11",
+  "wandb-carbs",
+  "wandb-core>=0.17.0b11",
```
berekuk, line 60: Formatting is off here.
I don't really care but @rjwalters might want to investigate (was it because `uv add` reformats? or vscode?)

app_backend/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+WORKDIR /app
+
+# Copy uv configuration files from top-level directory
+COPY uv.lock ./
+COPY app_backend/pyproject.toml ./
+
+COPY app_backend ./app_backend
+
+# Install dependencies using uv
+RUN uv sync --frozen
```
berekuk, line 22: pretty cool that this does the right thing when you're in `/app` and not in `/app/app_backend`, I didn't expect that

berekuk (approved):

PR #1049, opened 2025-06-19
Change terrain from numpy so that we don't save to root

daphnedemekas:


newfadel (approved):
The PR description "Change terrain from numpy so that we don't save to root" suggests a positive change. For `mettagrid/mettagrid/room/terrain_from_numpy.py`, I'd focus on ensuring the new save location is robust and properly handled.

Here's a professional code review comment:

"Great to see the change to prevent saving directly to the root! This improves security and adherence to best practices. Could you confirm that the new save location is configurable or derived in a way that's flexible for different deployment environments (e.g., using a dedicated application data directory)? Also, please ensure proper error handling is in place if the new target directory doesn't exist or isn't writable. 😄"

georgejwdeane (approved):

PR #1061, opened 2025-06-20
Revisioning via Torch.package v0

relh:
This PR adds `torch.package` as a primary method for saving/loading policies.

It defaults to using save/load using `state_dict` if loading a legacy checkpoint. 

If you try to load an old policy and then save it again it will copy it into a new agent and save that. 

---
 
> I had a longer PR that tried to do a big overhaul of how MettaAgent and PolicyRecord works. What we want to do instead is, starting from where we are now, create a MettaAgentBuilder. 
> 
> Things we want:
> 1) we want to be able to build in one of two ways, either BrainPolicies or PytorchPolicies (see the PR linked)
> 2) we want to avoid having 3 layers of wrapping, i.e., PolicyRecord wraps MettaAgent wraps BrainPolicy. Ideally, we could build MettaAgents one of two ways, either through building a BrainPolicy or through building a PytorchPolicy. Then, after this, we shouldn't have to think anymore about how the policy was built. 
> 3) The final feature we need is to fix agent revisioning so we can save/load old models using torch.jit/script/trace
> 
> How we're approaching this is add a MettaAgentBuilder class to metta_agent.py. The builder should do one of 3 things:
> - create a BrainPolicy 
> - create a PytorchPolicy (formerly PytorchAgent)
> - return an agent loaded from disk
> 
> Specifically, when PolicyRecord calls .policy() it should basically be going through the MettaAgentBuilder.
> 
> We want to keep PolicyRecords for now. Our main goal here is to pull the BrainPolicy (here in an example format) out from the MettaAgent so that BrainPolicy and the new PytorchPolicy can be equal status. 
> 
> This is a 4th take on:
> https://github.com/Metta-AI/metta/pull/973
> https://github.com/Metta-AI/metta/pull/1043
> https://github.com/Metta-AI/metta/pull/1046
> 
> It is a simplified version of the following that keeps the PolicyRecord class:
> 
> In part 1, we change to conceptually only dealing with MettaAgents and not PolicyRecords. A MettaAgent may be currently saved on disk or loaded in memory, but we shouldn't have to worry or think about this.
> 
> This part 1 ONLY refactors all mentions of policy_record to metta_agent (and other abbreviations like pr -> ma). It also squashes the MettaAgent <-> PolicyRecord class into one class (for now).
> 
> In part 2, we will change functionality so that the MettaAgent class becomes a wrapper for either a BrainPolicy (built by us) or a PytorchPolicy (anything external).
> 
> In part 3, we will change how saving/loading works so that we can use this new wrapper class to load any model, even if it's old.
> 
> [Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210596242460379)


metta/agent/policy_store.py
```diff
@@ -172,12 +176,116 @@ def __repr__(self):
 
         return "\n".join(lines)
 
+    def _clean_metadata_for_packaging(self, metadata: dict) -> dict:
+        import copy
```
daveey, line 180: prefer all imports at top

PR #1068, opened 2025-06-20
Residual MLPs

Al-does:
Added very deep residual MLPs to the nn_layer_library.
https://arxiv.org/pdf/2503.14858

Other work: 
- Added a prototypes folder under config/agent to keep things clean
- Added a config that generates an agent using a 64 layer MLP for the critic network.
- Removed the post-LSTM ReLU as it seems to slightly degrade performance.


metta/agent/lib/nn_layer_library.py
```diff
@@ -25,6 +25,74 @@
...
+            nn.Linear(hidden_size, hidden_size),
+            nn.LayerNorm(hidden_size),
+            Swift(),
+        )
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return x + self.block(x)
+
+
+class Swift(nn.Module):
```
relh, line 54: Is this a naming convention from the paper?

relh (changes requested):

relh (approved):

PR #1083, opened 2025-06-20
kickstarting updates

Al-does:
Kickstarting now runs again without error and student learns from teacher. However, student policy still collapses. Also added linear annealing to the kickstarting coefficients.

Future work will explore keeping the student stable as it anneals off the teacher.  

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210605001578935)

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210604862777481)

### DIFF
--- a/configs/trainer/puffer.yaml
+++ b/configs/trainer/puffer.yaml
@@ -78,9 +78,9 @@ kickstart:
   anneal_ratio: 0.65 # ratio of kickstart_steps to anneal to 0 coef
   kickstart_steps: 1_000_000_000
   additional_teachers:
-    - teacher_uri: wandb://run/m_alexv_ks_dr_lam_001:v22
-      action_loss_coef: 0.5
-      value_loss_coef: 0.6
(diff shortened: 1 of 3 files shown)


configs/trainer/puffer.yaml
```diff
@@ -75,11 +75,12 @@ kickstart:
   teacher_uri: null
   action_loss_coef: 1
   value_loss_coef: 1
-  kickstart_steps: 50_000_000
+  anneal_ratio: 0.65 # ratio of kickstart_steps to anneal to 0 coef
```
daveey, line 78: let's move this into "kickstarter" section?

daveey (approved):
looks good, but let's move these into trainer.kickstarter

configs/trainer/puffer.yaml
Al-does: I think the diff window cuts off but this is already indented under `kickstart:`

PR #1089, opened 2025-06-20
hotfix: fix test_process_metrics

rjwalters:


metta/agent/metta_agent.py
```diff
@@ -303,7 +303,6 @@ def forward(
                 # Training: x should have shape (B, T, obs_w, obs_h, features)
                 B, T, A = action.shape
                 assert A == 2, f"Action dimensionality should be 2, got {A}"
-                # assert_shape(x, (B, T, obs_w, obs_h, features), "training_input_x")
```
sasmith, line 306: sort of a weird pick up

rjwalters: good catch -- an alexv in office opportunistic request

PR #1090, opened 2025-06-20
Generate and serve replays without uploading them to S3.

treeform:
For people running our stuff from the outside there should be no need to configured S3. All they should do is compile the mettascope viewer locally and run generate reply tool.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210605337872774)


tools/replay.py
```diff
@@ -61,7 +61,12 @@ def main(cfg):
 
         # Only on macos open a browser to the replay
         if platform.system() == "Darwin":
-            webbrowser.open(f"https://metta-ai.github.io/metta/?replayUrl={http_url(replay_url)}")
+            if replay_url.startswith("./train_dir"):
+                local_url = replay_url.replace("./train_dir/", "http://localhost:8000/train_dir/")
```
daveey, line 65: this seems brittle. how about localhost:8000/local/ as the root, and adding the replay_url to it:

http://localhost:8000/local/train_dir/

that way as long as the url doesn't start with http, we do that

daveey (changes requested):

daveey (approved):

PR #1093, opened 2025-06-20
Bucketed configs

daphnedemekas:


run_memory_evals.sh
```diff
@@ -0,0 +1,150 @@
+#!/bin/bash
```
daveey, line 1: we don't want this in /, can we put this into ./scripts/

daveey (changes requested):
failing tests and merge conflicts

PR #1100, opened 2025-06-23
fix: move towards namespace packages

rjwalters:
Migrates the monorepo to use namespace packages, allowing better modularity and independent distribution of components while maintaining a unified `metta` namespace.

### Package Structure
- **Namespace Package Setup**: Converted `metta` to a namespace package using `pkgutil.extend_path()`
- **Module Reorganization**: 
  - Moved `mettagrid` → `metta.mettagrid` 
  - Created `metta.common` for shared utilities (extracted from mettagrid)
  - All imports updated throughout the codebase

### New Components
- **`metta-common` package**: 
  - Houses shared utilities like `Stopwatch` 
  - Independently installable with proper packaging configuration
  - Located at `common/` with standard Python package structure

### Build Configuration
- **scikit-build-core**: Added `wheel.packages = ["src/metta"]` to ensure Python files are included in wheels
- **Import paths**: Updated all imports from `mettagrid.*` to `metta.mettagrid.*`
- **C++ headers**: Updated include guards to remove redundant namespace prefixes

### Key Technical Details
- Each `metta/__init__.py` contains `__path__ = __import__("pkgutil").extend_path(__path__, __name__)` 
- Maintains compatibility with existing `metta/metta` directory structure
- All packages (`metta`, `metta-common`, `metta-mettagrid`) are independently installable
- C++ build paths updated to reflect new source structure


common/src/metta/common/stopwatch.py
```diff
@@ -87,15 +87,19 @@ def decorator(func: F) -> F:
 
         @functools.wraps(func)
         def wrapper(*args, **kwargs):
-            with timer.time(name, log_level=log_level, filename=filename, lineno=lineno):
+            with timer.time(
+                name, log_level=log_level, filename=filename, lineno=lineno
+            ):
```
berekuk, line 92: This and other code below looks like it was reformatted with 80 line length config by accident

common/pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "metta-common"
+version = "0.1.0"
+description = "Common utilities for metta packages"
+readme = "README.md"
+requires-python = "==3.11.7"
+license = { text = "MIT" }
+authors = [{ name = "Your Name", email = "your.email@example.com" }]
```
berekuk: placeholder value

common/src/metta/__init__.py
```diff
@@ -0,0 +1,2 @@
+# common/src/metta/__init__.py
+__path__ = __import__("pkgutil").extend_path(__path__, __name__)
```
berekuk, line 2: I see this is something called "[pkgutil-style namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages)", ok. That guide also says it's a legacy thing and native namespace packages should be preferred.

After doing some research I think we might be blocked by [this issue](https://github.com/scikit-build/scikit-build-core/issues/1040#issuecomment-2775908493) in scikit-build-core.

It's a pity that we can't use native namespace packages, and this is why my vote was for plain `mettacommon`, `mettagrid`, etc. I expect we'll have some friction in the future where imports won't work because someone will forget to add this file in a new package.

I think this might be worth documenting somewhere? Both the instructions on when this file is needed, and the explanation how we're blocked by scikit-build-core (if my diagnosis is correct).

berekuk (approved):

PR #1101, opened 2025-06-23
Msb multistage cric

mstormbull:
Feature: Add WandB logging for curriculum task sampling probabilities

[asana task](https://app.asana.com/1/1209016784099267/project/1210534060817260/task/1210534076114725)

Description:
This PR adds minimal WandB logging to track the probability distribution over curriculum tasks in the RandomCurriculum class. Each time a task is sampled, the current sampling probabilities for all tasks are logged to WandB under the key curriculum/task_probs. This enables monitoring and analysis of how the curriculum distribution evolves over time in the WandB dashboard.

Key changes:
*Added WandB logging in RandomCurriculum.get_task() to report the probability of sampling each task. Since this change is implemented in the RandomClass get_task() method, it is inherited by: Learning Progress, Low Reward. We also add the changes Progressive for similar tracking as it does not inheriet this method directly
*Only the normalized probabilities are logged (not raw weights or task IDs).
*Logging occurs automatically if a WandB run is active.

Impact:
This change allows for better visibility into curriculum task selection dynamics during training, supporting debugging and curriculum analysis.

Sample run avaliable at: https://wandb.ai/metta-research/metta/runs/msb_testLogging_02/workspace

### DIFF
--- a/mettagrid/mettagrid/curriculum/progressive.py
+++ b/mettagrid/mettagrid/curriculum/progressive.py
@@ -5,6 +5,7 @@ import logging
 from typing import Optional
 
 from omegaconf import DictConfig, OmegaConf
+import wandb
 
 from .curriculum import Task
 from .sampling import SamplingCurriculum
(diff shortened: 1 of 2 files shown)


daveey (commented):
This won't work, we make many copies of the env and having them all log to wandb is going to break things.

The env should get probs from the curriculum, and stick it into infos. then they will get logged to wandb automatically

daveey (changes requested):
move to env

PR #1105, opened 2025-06-23
Refactor 'Launch SkyPilot Training Job' to use reusable GitHub action

dominik3141:
The GitHub workflow 'Launch SkyPilot Training Job' handles all the necessary credential stuff to simply create a new SkyPilot training run. So far, this workflow is being used for a smoke test on new merges to main.

**Changes**
Refactor the 'Launch SkyPilot Training Job' workflow to use a newly created GitHub action to deploy the SkyPilot job.

**Context**
I am working on a smoke test for distributed training that also needs to deploy SkyPilot jobs, and copying this workflow would be suboptimal.

### DIFF
--- a/.github/actions/launch-skypilot-job/action.yml
+++ b/.github/actions/launch-skypilot-job/action.yml
@@ -20,14 +20,6 @@ inputs:
     required: false
     default: ""
 
-outputs:
-  run_name:
-    description: "The W&B run name used for the launched job."
-    value: ${{ inputs.run_name }}
-  sky_job_name:
-    description: "The name given to the SkyPilot job."
-    value: ${{ inputs.run_name }}
-
 runs:
   using: "composite"
   steps:
--- a/.github/actions/launch-skypilot-job/action.yml
+++ b/.github/actions/launch-skypilot-job/action.yml
@@ -32,9 +32,9 @@ runs:
   using: "composite"
   steps:
     - name: Setup Python and uv
-      uses: ./.github/actions/setup-uv # Assuming this action is in the repo root
+      uses: ./.github/actions/setup-uv
       with:
-        include-dev: true # Assuming dev dependencies might be needed by launch.py or related scripts
+        include-dev: true
 
     - name: Determine Git Ref
       id: determine_git_ref
--- a/.github/actions/launch-skypilot-job/action.yml
+++ b/.github/actions/launch-skypilot-job/action.yml
@@ -1,0 +1,88 @@
+name: "Launch SkyPilot Job"
+description: "Sets up and launches a SkyPilot job with necessary configurations."
+
+inputs:
+  trainer_env:
+    description: "Training environment configuration (e.g., env/mettagrid/simple)"
+    required: true
+  timeout_hours:
+    description: "Job timeout in hours"
+    required: true
+  run_name:
+    description: "The unique name for the W&B run and SkyPilot job."
+    required: true
+  num_gpus:
+    description: "Number of GPUs to request. If empty, uses SkyPilot task default."
+    required: false
+    default: ""
+  git_ref_override: # Allow overriding the git ref for specific cases, normally determined automatically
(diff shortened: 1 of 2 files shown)


.github/actions/launch-skypilot-job/action.yml
```diff
@@ -0,0 +1,88 @@
...
+    required: false
+    default: ""
+
+outputs:
+  run_name:
+    description: "The W&B run name used for the launched job."
+    value: ${{ inputs.run_name }}
+  sky_job_name:
+    description: "The name given to the SkyPilot job."
+    value: ${{ inputs.run_name }}
```
rjwalters, line 29: are these always going to be the same?

.github/actions/launch-skypilot-job/action.yml
```diff
@@ -0,0 +1,88 @@
...
+        fi
+        echo "GIT_REF_TO_USE=$GIT_REF_TO_USE" >> $GITHUB_OUTPUT
+        echo "Using Git Ref: $GIT_REF_TO_USE"
+
+    - name: Set up Weights & Biases credentials
+      shell: bash
+      run: |
+        echo "machine api.wandb.ai" > $HOME/.netrc
+        echo "login user" >> $HOME/.netrc
+        echo "password ${{ secrets.WANDB_API_KEY }}" >> $HOME/.netrc
```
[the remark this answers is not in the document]
rjwalters: @claude open-pr can you add `wandb_api_key` to the inputs for this action and update the use here

.github/workflows/sky-train.yml
```diff
@@ -34,64 +39,56 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
-          fetch-depth: 2 # Need at least 2 commits to detect merge commit messages
+          fetch-depth: 0 # Full history is useful as the action determines the exact git ref.
```
rjwalters, line 42: My suggestion would be keep the pr naming scheme for the main push CI runs but skip it for workflow_dispatch runs

we can have:
"github.sky.pr1092.a63fc8" for pushes to main 
"github.sky.a63fc8" for dispatched actions


then you can avoid fetching the entire history and just skip the pr name retrieval step conditionally

rjwalters (changes requested):
consider switching the manual launch to require a commit hash instead of a PR number

.github/workflows/sky-train.yml
dominik3141: Done

dominik3141:
> consider switching the manual launch to require a commit hash instead of a PR number

This is now possible.

.github/actions/launch-skypilot-job/action.yml
rjwalters: let's just have one parameter then? "run name"

rjwalters:
@claude open-pr
  add `wandb_api_key` to the inputs for this action and update the use here
  combine the outputs from the action to a single parameter "run_name"

rjwalters:
@claude open-pr I think I have fixed the permission issues... let's try once more to open a pr to fix the problems on this branch! 

⚠️ ISSUE: Switch to correct branch: claude/auto-1105-2025-06-24T03-57-54 (cannot do - no Bash permissions)
Read repository standards from CLAUDE.md
Examine current GitHub Actions files
Add wandb_api_key input to action.yml
Update workflow to pass wandb_api_key parameter
Combine action outputs to single "run_name" parameter
⚠️ ISSUE: Run quality checks if available (cannot do - no Bash permissions)
⚠️ ISSUE: Commit changes (MCP failed with 404, no Bash permissions for fallback)

.github/actions/launch-skypilot-job/action.yml
rjwalters: or maybe no outputs if we aren't using them

dominik3141:
> I think you can drop the outputs from the action but otherwise looks good

Good idea, I only needed the outputs before we specified the name as an argument. Now they don't make sense anymore.

PR #1116, opened 2025-06-24
fix: torch.package re-packaging error

rjwalters:
### Problem

When attempting to save policies that were previously loaded from torch packages, the system would fail with:
```
RuntimeError: Failed to save policy using torch.package: Can't pickle <class
'<torch_package_2>.metta.agent.metta_agent.MettaAgent'>: it's not found as
<torch_package_2>.metta.agent.metta_agent.MettaAgent
```

This occurs because models loaded from torch packages have module paths like `<torch_package_N>.*` which cannot be re-packaged.

### Solution

This PR implements a more robust save/load system that:

1. **Detects torch-packaged models** by checking for `<torch_package` in the module path
2. **Falls back to checkpoint format** when re-packaging would fail
3. **Provides unified loading** that seamlessly handles both package and checkpoint formats

#### Save Logic (`_save_policy`)
- Split into two internal methods: `_save_as_package()` and `_save_as_checkpoint()`
- Automatically detects if a model was loaded from a torch package
- Falls back to checkpoint format if package save fails
- Preserves all metadata in both formats for compatibility

#### Load Logic (`_load_from_file`)
- Attempts package format first, falls back to checkpoint format
- Simplified error handling and format detection
- Maintains backward compatibility with legacy checkpoints

#### Checkpoint Format Improvements
- Stores complete metadata including `obs_shape`, `action_space_nvec`, `feature_normalizations`
- Preserves optional model attributes like `config` and `hparams`
- Uses consistent metadata structure between formats

This change maintains backward compatibility with existing saved policies.

### DIFF
--- a/metta/agent/policy_store.py
+++ b/metta/agent/policy_store.py
@@ -179,36 +179,75 @@ class PolicyStore:
                 "train_time": 0,
             },
         )
-        self.save_policy(path, policy, pr)
+        self._save_policy(path, policy, pr)
         pr._policy = policy
         return pr
 
     def save(self, name: str, path: str, policy: nn.Module, metadata: dict) -> PolicyRecord:
         """Convenience method to create and save a policy in one step."""
         pr = PolicyRecord(self, name, "file://" + path, metadata)
-        return self.save_policy(path, policy, pr)
+        return self._save_policy(path, policy, pr)
 
-    def save_policy(self, path: str, policy: nn.Module, pr: PolicyRecord) -> PolicyRecord:
-        """Save a policy and its metadata using torch.package."""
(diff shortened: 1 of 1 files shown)


relh:
This is a bit verbose, I have code in trainer that should catch this use case, how did you generate this error? just curious what command got us here.

I have a small match that should fix this in just a jiffy.

relh (changes requested):
Okay well that hypothesis is gone because of no pulling on the machine, let's try to recreate the error on main later

rjwalters:
closing -- the reconstruction of the policy was done in trainer and I had accidentally deleted the fix in a merge commit to 1104

PR #1117, opened 2025-06-24
map editor in mettamap

berekuk:
map editor in mettamap

new tiles

refactor

refactor

refactor

new transforms

wip

fix pan and zoom

new encoding, various fixes

remove legacy assets

refactor

refactor

bugfixes, team agents

tabs component

fix

ascii mode editing

mettamap/start script

link to ascii maps


rjwalters:
do you want to add biome?

berekuk:
> do you want to add biome?

Open to it, but I don't want to increase the diff on this even more (biome will probably slightly reformat all files). I don't have biome vscode extension configured, and we don't have it in `.vscode/extensions.json`, which means I was auto-formatting all these files with prettier.

The plan is to merge mettamap code into observatory soon, and at that point I'll reformat everything in a single style.

PR #1118, opened 2025-06-24
log memory use

rjwalters:


metta/rl/trainer.py
```diff
@@ -479,6 +480,21 @@ def _rollout(self):
                     except TypeError:
                         self.stats[k] = [self.stats[k], v]  # fallback: bundle as list
 
+        memory_info = {
```
daveey, line 483: can we do

self._memory_monitor=MemoryMonitor()
self._memory_monitor.add({....})

...

**memory_monitor.stats()

mettagrid/src/metta/mettagrid/mettagrid_env.py
```diff
@@ -267,6 +269,18 @@ def process_episode_stats(self, infos: Dict[str, Any]):
         for n, v in infos["agent"].items():
             infos["agent"][n] = v / self._c_env.num_agents
 
+        attributes: dict[str, Any] = {
```
daveey, line 272: this is confusing, why not just infos["attribures"].update({....})

daveey (changes requested):

PR #1136, opened 2025-06-24
improve README installation and usage sections

nishu-builder:
Installation section changes:
- Include mention of setup_machine scripts
- Clarify what each of the setup steps does
- Add note to reach out to us about permission failures for non-team members (let me know if we should advise something else)

Updated usage section
- Add hydra explanation with link to docs. Explain parameter override (`param=value`) vs config composition (`+group=option`) 
- Clearer formatting/param explanations
- And add example user config file and reference it

Updated "visualizing a model" section

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210633558401137)

---
  - https://app.asana.com/0/0/1210632432027187

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,41 +183,29 @@ Now you can run training with your personal WandB config:
 
 ## Visualizing a Model
 
-### Run the interactive simulation
+### Mettascope: in-browser viewer
 
-```
-./tools/play.py run=my_experiment +hardware=macbook wandb=off
-```
+Mettascope allows you to run and view episodes in the environment you specify. It goes beyond just spectator mode, and allows taking over an agent and controlling it manually.
 
-This launches a human-controlled session using the same configuration flags as
-training. It is useful for quickly testing maps or policies on your local
-hardware.
+For more information, see [./mettascope/README.md](./mettascope/README.md).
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -100,37 +100,76 @@ This README provides only a brief overview of research explorations. Visit the [
...
+Choose the appropriate setup method for your platform:
 
+**macOS:**
 ```bash
-./devops/setup_dev.sh
+./devops/macos/setup_machine.py
 ```
 
-After git updates, you might need to run `uv sync` to reinstall all necessary dependencies.
+**Linux:**
```
rjwalters, line 122: I think the script at /devops/mettabox has stuff that is specific to the mettabox hardware -- my suggestion would be to either add a new script at /devops/linux/setup_machine.py that would correctly configure any linux machine (and reference this from the /mettabox hardward-specific setup script) or instead to just drop the linux section for now

rjwalters (changes requested):

README.md
nishu-builder: Makes sense, thank you!

rjwalters (approved):

PR #1151, opened 2025-06-25
Allow inventory items to be configured from outside cpp

sasmith:
Adds explicit inventory item configuration to the MettaGrid environment, replacing the previous hardcoded enum approach with a more flexible system. The inventory items are now defined in the game configuration and passed through to the C++ code.

* Gets rid of InventoryItem
* Does some weird passing of "what's a laser, what's armor".
* Still has hard-coded names, which we use for stats. We'll need to pass the inventory names in separately.

This also requires "inventory_item_names" to be defined up at the top, which causes a little redundancy, but gives easier understanding / safety.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210642391358505)

### DIFF
--- a/configs/env/mettagrid/mettagrid.yaml
+++ b/configs/env/mettagrid/mettagrid.yaml
@@ -8,6 +8,20 @@ game:
   num_observation_tokens: 200
   max_steps: 1000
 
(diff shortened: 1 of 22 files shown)


sasmith:
@OpenHands please fix the failing actions on PR #1151

PR #1167, opened 2025-06-26
Remove unused YAML and consolidate puffer.yaml and trainer.yaml together

relh:
Changes to `trainer.yaml` surprised people who were used to looking at `puffer.yaml`. We're merging the two of them so there's a single place to look at trainer details called `common.yaml`. As needed people can make more trainer-related YAMLs but we're avoiding people looking in multiple places for trainer details.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210644726463790)


daveey (approved):
i think trainer.yaml seems better than common

PR #1174, opened 2025-06-26
feat: launch skypilot training from anywhere

rjwalters:
This PR enhances the `lt` alias to work from any directory within the repository by automatically finding and changing to the repository root before executing the launch script.

## Changes
  - Replaces the simple alias with a bash function that accepts arguments
  - Automatically navigates to repo root before launching
  - Restores original working directory after execution
  - Preserves exit codes from the launch script

## Usage
Before this change, `lt` only worked from the repository root:
```bash
# Only worked from repo root
./devops/skypilot/launch.py train
```

After this change, `lt` works from anywhere in the repo:
```bash
# Now works from any subdirectory
cd src/models/
lt train trainer.curriculum=env/mettagrid/simple --timeout-hours=3 -c --no-spot run=my.job.1
```

## Technical Details
- The function searches upward through the directory tree until it finds a `.git` directory
- Uses `unalias lt 2>/dev/null` to handle cases where an alias might already exist
- Preserves all command-line arguments with `"$@"`
- Maintains the original working directory after script execution

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210644958829352)

### DIFF
--- a/devops/skypilot/setup_shell.sh
+++ b/devops/skypilot/setup_shell.sh
@@ -1,3 +1,28 @@
+#!/bin/bash
+
+cd_repo_root() {
+    local current_dir="$(pwd)"
+    local search_dir="$current_dir"
+
+    # Search current directory and all parent directories
+    while [[ "$search_dir" != "/" ]]; do
+        if [[ -d "$search_dir/.git" ]]; then
+            cd "$search_dir"
+            return 0
+        fi
+        search_dir="$(dirname "$search_dir")"
+    done
+
+    # Check root directory as well
+    if [[ -d "/.git" ]]; then
+        cd "/"
+        return 0
+    fi
+
+    echo "Repository root not found - no .git directory in current path or parent directories" >&2
+    return 1
+}
+
 export AWS_PROFILE=softmax
 
 # list jobs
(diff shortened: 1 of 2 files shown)


devops/skypilot/setup_shell.sh
```diff
@@ -15,4 +28,13 @@ alias jll='sky jobs logs $(jj | grep -A1 TASK | grep -v TASK | awk "{print \$1}"
 alias jllc='sky jobs logs --controller $(jj | grep -A1 TASK | grep -v TASK | awk "{print \$1}")'
 
 # launch training
-alias lt="./devops/skypilot/launch.py train"
+unalias lt 2>/dev/null
+lt() {
+    local original_dir="$(pwd)"
+    if cd_repo_root; then
+        ./devops/skypilot/launch.py "$@"
```
berekuk, line 35: No `train` anymore?

Note that the version without `train` ls already aliased to `skypilot` (if venv is activated), see root pyproject file:
```
[project.scripts]
skypilot = "devops.skypilot.launch:main"
```

If people still do activate venv (I know I'm doing it less often now, not sure about others), then you might as well alias the script there.

berekuk (approved):

devops/skypilot/setup_shell.sh
berekuk: It tries to run `./devops/trainer.curriculum=env/mettagrid/simple.sh` :)

berekuk: But that happens after the machine is spinned up, maybe we need some sanity check in `launch.py` for this.

rjwalters: I see -- the cmd_args are just parsed as the cmd
`    (args, cmd_args) = parser.parse_known_args(filtered_args)`

maybe we should consider an allow list

PR #1181, opened 2025-06-26
Supporting initialize_to_environment for if/when symbols change.

relh:
We support a map that holds feature_names -> ids.

We store this when agents are created including an unknown token.

When agents are initialized they check with the environment to see if things have changed and if so they do remapping.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210664578045454)


metta/rl/policy.py
```diff
@@ -80,6 +80,15 @@ def activate_actions(self, action_names: list[str], action_max_params: list[int]
             self.policy.activate_actions(action_names, action_max_params, device)
         self.device = device
 
+    def initialize_to_environment(
+        self, features: dict[str, dict], action_names: list[str], action_max_params: list[int], device
+    ):
+        """Initialize to environment - forward to wrapped policy if it has this method."""
+        if hasattr(self.policy, "initialize_to_environment"):
```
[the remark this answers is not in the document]
relh: I agree! I think we can just enforce this and at the same time maybe refactor `activate_actions` to `_initialize_actions` for consistency since they're both breaking changes

relh:
This is updated, but because we have other errors with saving/loading checkpoints, I need to revisit it.

metta/agent/metta_agent.py
```diff
@@ -149,9 +155,134 @@ def _setup_components(self, component):
...
+        """
+        Initialize the policy to the current environment's features and actions.
+        This should be called exactly once per time the policy is "brought out of storage".
+
+        Args:
+            features: Dictionary mapping feature names to their properties:
+                {
+                    feature_name: {
+                        "id": byte,  # The feature_id to use during this run
+                        "type": "scalar" | "categorical",
```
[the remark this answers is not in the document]
sasmith: https://app.asana.com/1/1209016784099267/project/1209041170403474/task/1210668408833122 tracking this

tests/agent/test_feature_remapping.py
```diff
@@ -0,0 +1,348 @@
...
+
+    # Create input tensor dict
+    td = TensorDict({"x": observations})
+
+    # Process through pad strip
+    output_td = pad_strip._forward(td)
+    remapped_obs = output_td["test_pad_strip"]
+
+    # Verify remapping
+    assert remapped_obs[0, 0, 1] == 1  # 3->1
```
sasmith, line 53: I assume we're not writing our own tests, but think this would be easier to read if we did all the assertions as a big block. Like
```
assert remapped_obs = [
    [0x12, 1, 10],
    [0x34, 2, 20], 
  ...
```

I'm not worried about this being too coarse, since in case or errors, I think it would be better to report the full `remapped_obs` and `expected_obs` anyways.

tests/agent/test_feature_remapping.py
```diff
@@ -0,0 +1,348 @@
...
+    agent._update_normalization_factors = lambda features: MettaAgent._update_normalization_factors(agent, features)
+    agent._apply_feature_remapping = lambda features, unknown_id: MettaAgent._apply_feature_remapping(
+        agent, features, unknown_id
+    )
+    agent.get_original_feature_mapping = lambda: MettaAgent.get_original_feature_mapping(agent)
+
+    MettaAgent._initialize_observations(agent, new_features, "cpu", True)
+
+    # Verify remapping was created
+    assert hasattr(agent, "feature_id_remap")
```
sasmith, line 133: Nit: I kind of feel like assertions of this level are unneeded. If it doesn't have the property, we're going to fail anyways, and the messaging will be clear.

I'm on the fence about whether we should be telling our LLMs to stop being this verbose, and how much we should bother correcting stuff like this, but I've been having to update a lot of tests recently, so I'm probably at peak make-tests-maintainable at the moment.

tests/agent/test_feature_remapping.py
```diff
@@ -0,0 +1,348 @@
...
+
+    # Create a mock _obs_ component
+    class MockObsComponent:
+        def update_feature_remapping(self, remap_table):
+            self.remap_table = remap_table
+
+    mock_obs = MockObsComponent()
+    agent.components["_obs_"] = mock_obs
+
+    # Bind MettaAgent methods to mock agent for testing
```
sasmith, line 117: This setup seems pretty ugly. Why not just make the `MockAgent` be an actual subclass of `MettaAgent`?

If we want to do it this way, please bundle the setup at the top of the test, instead of switching between setup and actual testing and setup again.

sasmith (approved):
I'm being picky about tests. Decide on how leveraged it is to follow up. Happy to chat more about testing philosophy if it's helpful / interesting.

tests/agent/test_feature_remapping.py
relh: Yeah, I agree. I think attention is a premium resource and we've been not being concise with it when it comes to testing. And now we have a lot of tests its becoming really noticeable. Basically, I think we should probably have a review of all tests and make many of them more concise.

PR #1184, opened 2025-06-26
feat: adjust timeouts based on trainer checkpoint

rjwalters:
This PR adds timeout adjustment for training jobs that accounts for elapsed training time when resuming jobs. When launching a training job with a timeout, the system now checks how long the job has already been running and adjusts the timeout accordingly to prevent jobs from running longer than intended.

## Problem

Previously, when resuming a training job with a timeout (e.g., `--timeout-hours=2.0`), the full timeout would be applied regardless of how long the job had already been running. This could cause jobs to run significantly longer than intended:

- Start job with 2-hour timeout → job runs for 1.5 hours, gets interrupted
- Resume same job with 2-hour timeout → job could run for another 2 hours (3.5 total)
- This defeats the purpose of having timeouts for cost control and resource management

### Key components

1. **New utility module** (`metta/rl/util/trainer_state.py`):
   - `get_elapsed_time()`: Extracts elapsed time from trainer checkpoint files
   - Loads stopwatch state from `trainer_state.pt` files
   - Handles missing files and corrupted state gracefully

2. **Enhanced launch script** (`devops/skypilot/launch.py`):
   - Automatically calculates adjusted timeout before job launch
   - Shows clear messages about timeout adjustments
   - Displays both original and adjusted timeouts in job summary

3. **Timeout calculation logic** (`devops/skypilot/utils.py`):
   - `calculate_adjusted_timeout()`: Core logic for timeout adjustment
   - Handles three scenarios:
     - No existing training → use original timeout
     - Partial training → use remaining time
     - Exceeded timeout → apply minimal timeout with warning

### DIFF
--- a/common/src/metta/common/fs.py
+++ b/common/src/metta/common/fs.py
@@ -2,9 +2,12 @@ import os
 from pathlib import Path
 
 
-def cd_repo_root():
+def get_repo_root() -> Path:
     """
-    Ensure we're running in the repository root.
+    Get the repository root directory.
+
+    Returns:
+        Path to the repository root
 
     Raises:
         SystemExit: If repository root cannot be found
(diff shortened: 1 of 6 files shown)


berekuk (dismissed):
assuming that you've tested it, lgtm

rjwalters:
needs to know where the train_dir lives... maybe we can add ENV for this

PR #1193, opened 2025-06-26
Organize some trainer configs into subconfigs

nishu-builder:
Organize some trainer configs into subconfigs at Richard's request

Also updates existing configs and commands in READMEs to conform to this

Also adds a test to run through all of the configs we have stored in configs/user and configs/hardware to check if their trainer configs conform

Does not yet check sweeps

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210654261293548)

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210654369914572)


configs/trainer/trainer.yaml
```diff
@@ -2,103 +2,3 @@ _target_: metta.rl.trainer.MettaTrainer
 
 num_workers: ???
 
-curriculum: /env/mettagrid/curriculum/simple
```
relh, line 5: can we keep trainer.yaml around for now but restructure it to match the new configs as per trainer_config.py?

configs/user/bullm_lp.yaml
```diff
@@ -0,0 +1,37 @@
+# @package __global__
+
+# Learning Progress Training Configuration
+# Uses the learning progress navigation curriculum defined in env/mettagrid/curriculum/learningprogress_navigation.yaml
+
+defaults:
```
relh, line 6: delete this file because matt is using bullm.yaml now instead

relh (changes requested):
Basically, I'm hoping that we can re-structure the YAML files to match our new structured configs.

So if we restore trainer.yaml and adjust it to match the hierarchy of the new trainer_config, we'll be able to both be supporting hydra and supporting structured configs for validation. This way we smooth over the transition to structured configs.

PR #1196, opened 2025-06-26
Remove converter type_ids from cpp

sasmith:
Adds explicit type_id fields to all object configurations in the MettagGrid environment. This change decouples the object type identification from hardcoded enums.

Right now the type_ids are explicit in config files, which seems probably not where we want to end; but we need them for the moment because we want the type_ids to be stable. Happily, the iteration on getting rid of this should now let most of this iteration be moved out of cpp.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210654310008279)

### DIFF
--- a/mettagrid/src/metta/mettagrid/mettagrid_c.cpp
+++ b/mettagrid/src/metta/mettagrid/mettagrid_c.cpp
@@ -102,13 +102,25 @@ MettaGrid::MettaGrid(py::dict cfg, py::list map, int seed) {
   }
   init_action_handlers();
 
-  auto agent_groups = cfg["agent_groups"].cast<py::dict>();
+  auto object_configs = cfg["objects"].cast<py::dict>();
 
-  for (const auto& [key, value] : agent_groups) {
-    auto agent_group = value.cast<py::dict>();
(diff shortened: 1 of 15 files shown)


mettagrid/src/metta/mettagrid/mettagrid_c.cpp
```diff
@@ -102,13 +102,31 @@ MettaGrid::MettaGrid(py::dict cfg, py::list map, int seed) {
...
-    _group_sizes[id] = 0;
-    _group_reward_pct[id] = agent_group["group_reward_pct"].cast<float>();
+  object_type_names.resize(object_configs.size());
+
+  for (const auto& [key, value] : object_configs) {
+    auto object_cfg = value.cast<py::dict>();
+    TypeId type_id = object_cfg["type_id"].cast<TypeId>();
+
+    if (type_id >= object_type_names.size()) {
+      // Sometimes the type_ids are not contiguous, so we need to resize the vector.
```
sasmith, line 114: We weren't doing this before (in earlier versions of my PR), and it resulted in heap corruption.

PR #1198, opened 2025-06-26
Fix train script exit code

dominik3141:
## Summary
- propagate exit codes in `devops/train.sh`

## Testing
- `uv run ruff format`
- `uv run ruff check`
- `uv run pytest -q`
- `uv run mypy .` *(fails: pyenv: mypy: command not found)*

------
https://chatgpt.com/codex/tasks/task_e_685dc90c3d108325b08f2f89e8586ea4

## Context from a real human
We have the problem that skypilot jobs sometimes fail but show status 'SUCCESS', for example in the case of hydra config problems. This is a problem for our monitoring system that only looks at jobs that have failed. Skypilot determines success or failure based on the return code of the job, so propagating the exit code should be an easy fix.


relh (approved):
I am highly skeptical of exit codes as they often never arrive and/or distributed multi-gpu training often just hangs at the end for some unknown reason, but LGTM

PR #1202, opened 2025-06-27
fix: repair memory leak in stopwatch timer

rjwalters:
The `Stopwatch` class was experiencing a significant memory leak due to accumulating timer references without deduplication. Each time a timer was started (e.g., during training iterations on every parallel env), the system stored filename/line number pairs in a list without checking for duplicates. This resulted in:

- Continuously growing memory usage during long-running training sessions
- Out-of-memory issues in extended runs

## Root Cause

The `Timer.references` field was implemented as a `list[TimerReference]` that accumulated entries on every timer start without any deduplication logic. In training loops where the same timer is started repeatedly from the same location, this created thousands of duplicate entries.

## Solution

- **Changed data structure**: Converted `Timer.references` from `list[TimerReference]` to `set[tuple[str, int]]`
- **Automatic deduplication**: Sets inherently prevent duplicate entries
- **Maintained functionality**: All existing timer features work unchanged

## Key Changes
- `common/src/metta/common/stopwatch.py`: Modified Timer class to use set-based references
- Updated serialization/deserialization logic to handle set ↔ list conversion
- Fixed test cases to work with new tuple-based reference format
- Added safeguards for memory monitoring (only track on master processes)

## Performance Impact
- **Memory usage**: Significantly reduced memory footprint in long-running processes
- **Compatibility**: No breaking changes to public API

## Before/After
image
image

Not sure why we have two asana tasks, but:

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210655615413465)

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210655613276138)

### DIFF
--- a/common/src/metta/common/fs.py
+++ b/common/src/metta/common/fs.py
@@ -1,5 +1,4 @@
 import os
-import subprocess
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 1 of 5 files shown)


relh:
I see this as a mark in favor of the make-classes-for-things-sparingly camp. As primitive types in a set resolve this.

relh (approved):

PR #1203, opened 2025-06-27
Move type_id to config

sasmith:
Moves type_id to configuration files.

This is a portion of #1196, which I'm having trouble getting to pass tests.

Medium term, we should expect type_ids to be auto-generated rather than specified. At the moment we're keeping it stable since I expect that having them shift around is more likely to break things like visualizers that depend on type_ids being consistent.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210657050874854)

### DIFF
--- a/configs/env/mettagrid/mettagrid.yaml
+++ b/configs/env/mettagrid/mettagrid.yaml
@@ -93,6 +93,7 @@ game:
 
   objects:
     altar:
+      type_id: 8
       input_battery.red: 3
       output_heart: 1
       max_output: 5
@@ -101,6 +102,7 @@ game:
       initial_items: 1
 
     mine_red:
+      type_id: 2
       output_ore.red: 1
       color: 0
       max_output: 5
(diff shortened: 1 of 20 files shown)


sasmith:
* **#1203** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

configs/env/mettagrid/mettagrid.yaml
```diff
@@ -198,9 +210,11 @@ game:
       initial_items: 1
 
     wall:
+      type_id: 1
       swappable: false
 
     block:
+      type_id: 1
```
sasmith, line 217: It's bad form for these to have the same type_id. At least, it probably is. But for the moment we're preserving existing behavior.

sasmith: This PR also doesn't actually start _using_ type_ids from walls.

PR #1215, opened 2025-06-27
Add PufferLib Sweep Parameters

relh:
image

If we want to use these updated parameters they get this result

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210663499192858)


configs/env/mettagrid/mettagrid.yaml
```diff
@@ -36,15 +36,15 @@ game:
       # action_failure_penalty: 0.00001
       action_failure_penalty: 0
 
-      ore.red: 0.005
-      ore.blue: 0.005
-      ore.green: 0.005
+      ore.red: 0.17088483842567775
```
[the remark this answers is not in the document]
relh: To provide an understanding that getting more Ore is less important than using Ore to get batteries

relh:
https://github.com/Metta-AI/metta/pull/1201

David has this PR to eliminate the reward, we could just tweak the default settings and not mess with the rewards.

configs/env/mettagrid/mettagrid.yaml
daveey, line 39: can we truncate to X.XX ?

daveey, line 39: let's put these into puffer.rewards.yaml and import them when we want them rather than changing the entire mettagrid base. and set everything but hearts here to 0

configs/trainer/trainer.yaml
```diff
@@ -24,10 +24,10 @@ replay_dir: s3://softmax-public/replays/${run}
 
 total_timesteps: 50_000_000_000
 
-clip_coef: 0.1
-ent_coef: 0.0021
-gae_lambda: 0.916
-gamma: 0.977
+clip_coef: 0.14919147162017737
```
daveey, line 27: round to x.xx

configs/trainer/trainer.yaml
```diff
@@ -51,32 +51,32 @@ lr_scheduler:
...
+  vtrace_rho_clip: 2.296343917695581  # Default: on-policy (no off-policy correction)
+  vtrace_c_clip: 2.134490283650365    # Default: on-policy bootstrapping
 
 zero_copy: true
 require_contiguous_env_ids: false
 verbose: true
 
 batch_size: 262144
-minibatch_size: 16384
+minibatch_size: 32768
```
daveey, line 79: make sure you test this, mb size is hardware dependent and joseph has different gpus

daveey (changes requested):
x

configs/env/mettagrid/mettagrid.yaml
relh: good idea

tests/rl/test_trainer_config.py
```diff
@@ -152,9 +152,9 @@ def test_trainer_config_defaults(self):
         trainer_config = parse_trainer_config(make_cfg(incomplete_config))
 
         # Check that defaults were applied
-        assert trainer_config.optimizer.beta1 == 0.9
-        assert trainer_config.optimizer.beta2 == 0.999
-        assert trainer_config.optimizer.eps == 1e-12
+        assert trainer_config.optimizer.beta1 == 0.89
```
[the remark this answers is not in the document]
nishu-builder: cool sounds good

relh:
I don't know if we want to merge this PR, but it's doing a good job of showing how updates to defaults in the future might look. Here we're updating both the `trainer.yaml` and the structured configs as well as the comments on the configs, which can help track how values came to be.

daveey (changes requested):
i think we don't merge

relh:
It's interesting, these parameters worked well for simple but don't work as well for navigation.

tests/rl/test_trainer_config.py
```diff
@@ -150,11 +149,26 @@ def make_cfg(trainer_cfg: dict) -> DictConfig:
 
 
 class TestTypedConfigs:
+    """Tests for typed configuration parsing and validation.
```
nishu-builder, line 152: i like this change

tests/rl/test_trainer_config.py
```diff
@@ -343,6 +351,11 @@ def test_all_trainer_configs_comprehensive(self):
                 assert 0 < validated_config.ppo.gamma <= 1
```
nishu-builder, line 351: we can probably remove all these asserts; `create_trainer_config` does the job

relh:
Although better for simple, and initially equal for navigation training, the training collapses after sometime:

image

I'm assigning this to myself as a todo task to figure out why training collapses after some time.

PR #1216, opened 2025-06-27
Do not crash training when simulation fails

Barraketh:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210663049702353)


nishu-builder (approved):
seems good! since these are non-critical to the training loop, what do you think about making them fire-and-forget async? maybe asyncio.create_task() without awaiting or greenlet.spawn?

PR #1219, opened 2025-06-27
Add category and env_name to episodes

Barraketh:


app_backend/metta_repo.py
```diff
@@ -234,24 +245,32 @@ def record_episode(
         attributes: Dict[str, Any],
     ) -> uuid.UUID:
         with self.connect() as con:
+            # Parse eval_category and env_name from eval_name
+            eval_category = eval_name.split('/', 1)[0] if eval_name else None
```
nishu-builder: if super easy: I wonder if there's somewhere centralized in the codebase where we define the scheme for these eval names that we can sync here, so that if we ever change convention on how we concat category and name, this'll get updated too

but my instinct is there is no such place becuase there are just defined separately as a bunch of strings in the configs :P

PR #1229, opened 2025-06-28
configure_jobs_controller.py script

berekuk:
https://app.asana.com/1/1209016784099267/project/1210595772808575/task/1210662420883865?focus=true

Mostly hacked together with LLMs, but it works.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210664844744988)


daveey (approved):
I think diamond had a good suggestion

berekuk:
> I think diamond had a good suggestion

Doesn't really matter, `rm -rf` without args doesn't do anything.

PR #1237, opened 2025-06-30
Add indexes and performance logging

Barraketh:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210669497069677)

### DIFF
--- a/app_backend/query_logger.py
+++ b/app_backend/query_logger.py
@@ -1,0 +1,103 @@
+"""Database query logging utilities for performance monitoring."""
+
+import logging
+import time
+from contextlib import contextmanager
+from typing import Any, Generator, Tuple
+
+from psycopg import Connection, Cursor
+
+# Logger for database query performance
+query_logger = logging.getLogger("db_performance")
+query_logger.setLevel(logging.INFO)
+
+# Threshold for slow query warnings (1 second)
+SLOW_QUERY_THRESHOLD_SECONDS = 1.0
+
+
+@contextmanager
+def timed_query(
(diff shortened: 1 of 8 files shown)


app_backend/query_logger.py
```diff
@@ -0,0 +1,98 @@
...
+
+            yield cursor
+
+    except Exception as e:
+        execution_time = time.time() - start_time
+        query_logger.error(f"{query_id}query failed after {execution_time:.3f}s: {e}\nQuery: {query}\nParams: {params}")
+        raise
+
+
+def log_query_execution(con: Connection, query: Query, params: Tuple[Any, ...] = (), description: str = "") -> Any:
```
nishu-builder, line 65: minor: does this always return a list? list[Any] maybe

nishu-builder: nit: log_query_execution -> logged_query_execution or execute_query_logged

app_backend/routes/dashboard_routes.py
```diff
@@ -355,14 +414,17 @@ def create_dashboard_router(metta_repo: MettaRepo) -> APIRouter:
     user_or_token = Depends(create_user_or_token_dependency(metta_repo))
 
     @router.get("/suites")
+    @timed_route("get_suites")
```
nishu-builder, line 417: no big deal but we could turn @router.x(...) and @timed_route(...) into a single decorator @router_timed.x(...) where the timed_route func name is inferred

nishu-builder (approved):
looks good! indexes are wonderful. we could consider adding prometheus or datadog or something at a later point to auto-decorate our endpoint and give us histograms of timings/etc; i've loved those in the past

PR #1239, opened 2025-06-30
Made device check a decorator that fails

relh:
Instead of silently fixing the device this adds a decorator to main that throws an error if the device is wrong

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210674427895182)


relh:
added and checked on `tools/play.py`

PR #1247, opened 2025-06-30
auto-format python files with ruff on save

nishu-builder:


rjwalters (changes requested):
we set this in `.vscode/settings.json`

PR #1257, opened 2025-07-01
make setup profiles dynamically resolve components, update studio installation to not stall

nishu-builder:
Users now automatically get new components when added to their profile without needing to reconfigure. Fixed an issue where `./metta.sh status` didn't show slava's studio addition.

Prompts users to update with `./metta.sh configure` if their configuration version is old

Also allows users to provide a "y" input to the studio installation (otherwise blocked)


berekuk (approved):
works in local testing, lgtm

PR #1258, opened 2025-07-01
Convert some maps to mapgen; new mapgen features

berekuk:
Notes:
- the boundaries of this PR are arbitrary, I don't want to accumulate so many changes that it'll get hard to review, so I'll make a checkpoint here and stack the next iteration on top of it
- I tried to keep maps as close to the previous version as possible; some might have off-by-1 differences in border sizes, but afaik almost everything is exactly as it was before
- some configs, such as `puffer.yaml` and `full.yaml`, are intentionally more verbose, at the cost of small amount of copy-paste

Non-trivial features:
1. `instances` and `instance_border_width` in top-level mapgen class - without this new configs were too verbose (always required `RoomGrid` + `children`)
2. `intrinsic_size` in `InlineAscii` (I couldn't find a way to take into account optional scene sizes deeper in the scene tree, but at least we can do this for the top-level scene)
3. `Layout` scene - previous `bases.yaml` relied on `RoomGrid` behavior that placed smaller scenes in the center of areas; mapgen version is more verbose and explicit
4. Labels - without this, new configs would have a regression, won't report `env_map_reward/*` metrics.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210681490159435)

PR #1259, opened 2025-07-01
deterministic hierarchical randomness in mapgen

berekuk:
This PR refactors the random number generation system used in map generation to provide strict determinism and more structured distribution types.

When the seed is set at the root scene, if all scene classes are well-behaved, then the entire map will be deterministic, by spawning new rngs for all child scenes based on the parent's rng.

Other changes:
- Added proper distribution classes for both float and integer values with strong typing
- Added `Scene.factory()` helper method to simplify scene creation
- Removed the `order_by_seed` parameter (wasn't used)
- Random objects in Auto scene are now placed using percentage-based distributions rather than absolute counts.

The main reason for this PR is that I want to minimize the dependency on OmegaConf resolvers in map configs, because:
1. they're not flexible enough (get resolved too early, can't control seeds)
2. I want to load map scenes from `./scenes/**.yaml` configs, and ideally I want to avoid omegaconf in these configs entirely, or at least avoid hydra-dependent parts
3. `${sampling:...}` relies on top-level `sampling` parameter, so it depends on hydra, and I can't use that resolver in `./scenes/`

This PR makes some progress on (1) and (2), but it doesn't handle `sampling` yet. My current plan for handling `sampling` is to make it available to python code through [contextvars](https://docs.python.org/3/library/contextvars.html) and then make use of it in `metta.map.random.*` classes.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210681519163094)


berekuk:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1259?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1259** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1259?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1259?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1258** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

tests/map/random/test_float.py
```diff
@@ -0,0 +1,255 @@
...
+    def test_sample_returns_constant_value(self):
+        dist = FloatConstantDistribution(value=3.14)
+        rng = np.random.default_rng(seed=123)
+
+        # Should always return the same value regardless of rng
+        assert dist.sample(rng) == 3.14
+        assert dist.sample(rng) == 3.14
+        assert dist.sample(rng) == 3.14
+
+    def test_sample_with_negative_value(self):
```
sasmith, line 22: The negative and zero feel like unnecessary tests.

tests/map/random/test_float.py
```diff
@@ -0,0 +1,255 @@
...
+        assert all(5.0 <= sample <= 5.1 for sample in samples)
+
+    def test_sample_negative_range(self):
+        dist = FloatUniformDistribution(low=-10.5, high=-1.2)
+        rng = np.random.default_rng(seed=123)
+
+        samples = [dist.sample(rng) for _ in range(100)]
+        assert all(-10.5 <= sample <= -1.2 for sample in samples)
+
+    def test_sample_zero_crossing_range(self):
```
sasmith, line 66: this also seems superfluous. Do you have an example of what this might be guarding against?

sasmith (approved):

tests/map/random/test_float.py
berekuk: Thanks for spotting, I'm mostly LLMing tests these days, and when LLMs one-shot tests for the already working code it's hard to care about the specifics. Removed.

PR #1264, opened 2025-07-01
`./metta.sh {cmd}` -> `metta {cmd}`

nishu-builder:
New setup process is:
```
./install.sh
uv run metta configure
uv run metta install
```

Decisions I'm not sure about
- Keeps metta.sh around but just as a thin backwards-compatibility wrapper
- Updates all documentation to say `uv run metta {cmd}`. You could imagine instead saying `meta {cmd}`, and either assuming users have activated your uv-managed virtualenv, or instructing them to do so. I picked the option that Claude told me was canonical, but I'm sympathetic to the others too

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210688659378306)


daveey (changes requested):
Thanks for doing this. I really hate making users do "uv run" for everything. Can we make install.sh (optionally) add "source $METTA_DIR/.venv/bin/activate" to their shell profile? Or .... we could put "metta" as an executable with a 

"#!/usr/bin/env -S uv run" at the top

metta/setup/installer/install.sh
```diff
@@ -0,0 +1,238 @@
...
+PROJECT_DIR="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+
+# Source the install utilities from the same directory as the actual script
+. "$SCRIPT_DIR/install_utils.sh"
+
+# Parse command line arguments
+for arg in "$@"; do
+    case "$arg" in
+        --add-to-path)
+            say "--add-to-path has been deprecated; please set METTA_NO_MODIFY_PATH=0 in the environment"
```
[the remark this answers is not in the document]
nishu-builder: hah indeed, result of me re-squishing something I had set up across two PRs. good catch thank you!

devops/setup_dev.sh
```diff
@@ -2,10 +2,13 @@
 
 set -euo pipefail
 
-echo "This script has been replaced by metta.sh. Will run it with some default settings..."
```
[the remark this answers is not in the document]
nishu-builder: ah i see, you want clean to work without having to go through the metta cli at all, else we'll rely on dependencies being available to even run clean

metta/setup/installer/install_utils.sh
```diff
@@ -0,0 +1,303 @@
+#!/bin/sh
```
rjwalters, line 1: are we using anything other than "say"?

.gitignore
```diff
@@ -47,3 +47,7 @@ observatory/dist/
 
 # can be created by skypilot
 /Miniconda3-Linux.sh
+
+# Generated environment scripts
+metta/setup/installer/bin/env
```
daveey, line 52: i dont love putting this into the repo paths like this. how about putting it into .metta

CLAUDE.md
```diff
@@ -15,11 +15,12 @@ The codebase consists of:
 ## Development Environment Setup
 
 ```bash
-# Install uv if not already installed
-curl -LsSf https://astral.sh/uv/install.sh | sh
+# Initial setup (installs uv if needed and sets up environment)
+./install.sh
```
daveey, line 19: seems weird to call install twice. what if ./install.sh also called metta configure and metta install?

metta/setup/installer/install_utils.sh
nishu-builder: add_install_dir_to_path, add_to_ci_path, check_for_shadowed_bins, and replace_home

nishu-builder: the big one I want to use is add_install_dir_to_path, really

metta/setup/installer/install_utils.sh
```diff
@@ -0,0 +1,303 @@
...
+# shellcheck shell=dash
+# shellcheck disable=SC2039  # local is non-POSIX
+#
+# PATH modification utilities adapted from uv installer (MIT licensed)
```
daveey, line 5: I really hate having long shell scripts. Can we just use install.sh to install uv, and then run a python script to do the rest of the configure / install?

daveey (changes requested):

CLAUDE.md
nishu-builder: oh I like that

relh:
"#!/usr/bin/env -S uv run"

I'm a big fan of using this command at the top of scripts it's been helpful in my other API branch too

daveey (approved):

PR #1274, opened 2025-07-01
Pasha/add training run description 2

Barraketh:


app_backend/metta_repo.py
```diff
@@ -149,20 +149,34 @@
...
+        description="Add description field to training_runs table",
+        sql_statements=[
+            """ALTER TABLE training_runs ADD COLUMN description TEXT""",
+        ],
+    ),
+    SqlMigration(
+        version=8,
+        description="Add tags field to training_runs table",
+        sql_statements=[
+            """ALTER TABLE training_runs ADD COLUMN tags TEXT[]""",
```
nishu-builder, line 177: can do this later, but if we're going to support filtering/grouing/etc by tag, we may want tags to be its own table and `training_run_tag_associations` to be a many-to-many table; would make it more efficient

app_backend/metta_repo.py
```diff
@@ -624,3 +650,39 @@ def update_saved_dashboard(
                 (name, description, dashboard_type, Jsonb(dashboard_state), dashboard_uuid, user_id),
             )
             return result.rowcount > 0
+
+    def update_training_run_description(self, user_id: str, run_id: str, description: str) -> bool:
```
nishu-builder, line 654: i am slowly slowly (not there yet!) moving towards wanting an ORM :P optional usage, maybe!

metta/rl/trainer.py
```diff
@@ -326,10 +326,21 @@ def train(self) -> None:
         trainer_cfg = self.trainer_cfg
 
         if self._stats_client is not None:
-            name = self.wandb_run.name if self.wandb_run is not None and self.wandb_run.name is not None else "unknown"
-            url = self.wandb_run.url if self.wandb_run is not None else None
+            if self.wandb_run is not None:
+                name = self.wandb_run.name if self.wandb_run.name is not None else "unknown"
+                url = self.wandb_run.url
+                tags: list[str] | None = list(self.wandb_run.tags) if self.wandb_run.tags is not None else None
```
nishu-builder, line 332: we could make `run.tags` have a server default of `[]`; then we guarantee it's non-null. would lock up the prod db for a sec as we added it because we already have millions of rows, but coding going forward would be easier

app_backend/generate_test_data.py
```diff
@@ -0,0 +1,447 @@
+#!/usr/bin/env python3
+"""
```
nishu-builder, line 2: nice! idk if we want to have something to check that this is running in a non-production environment or not

nishu-builder (approved):
looks awesome!

PR #1275, opened 2025-07-01
(draft) fix: mv agent to isolated subpackage

rjwalters:


agent/Makefile
```diff
@@ -0,0 +1,20 @@
+
+.PHONY: help install test clean
```
relh, line 2: is the makefile needed for having agent be its own submodule? I thought we didn't want makefiles but I think I see why

agent/README.md
```diff
@@ -0,0 +1,26 @@
+# metta-agent
+
+Agent/Policy utilities for metta packages.
```
relh, line 3: We currently seem to be congregating around the idea of putting all .md and/or READMEs in the docs/ folder. I think since they're not code, it's nice to group them all together for real human-eyes browsing. Maybe we can make that move?

relh (approved):
If you could address the two comments that would be great, I thought this was going to be a large PR but then I got to the bottom half and it was just the reshuffling you mentioned.

agent/Makefile
rjwalters: just a convenience -- we can delete later

sasmith:
```
Exception has occurred: ModuleNotFoundError       (note: full exception trace is shown but execution is paused at: <module>)
No module named 'metta.agent.util.debug'
  File "/Users/sasmith/code/metta/agent/src/metta/agent/metta_agent.py", line 13, in <module> (Current frame)
    from metta.agent.util.debug import assert_shape
```

PR #1282, opened 2025-07-02
Add profile for softmax docker installation

nishu-builder:
This PR:
- adds a config profile for softmax docker boxes
- makes a few things more maintainable for when we choose to add more profiles

Motivation:
Richard noticed that our docker images were failing to run devops/setup_dev.sh properly after the introduction of the new metta.sh setup wizard. This is because setup_dev.sh now just wraps the new setup process as if it were a softmax user, which results in it trying to build some of our node-dependent projects (mettascope), but the docker images aren't built with node. our usage of the docker containers doesn't rely on those node-dependent projects being built

A future enhancement would be to have a single source of truth for list of system packages across macos/docker/linux/etc, have support for installing that list across those platforms' preferred package management system (brew vs apt etc), and and to have the required system packages be inferred from the set of requested components. That's coming....soon....

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210692503821506)

### DIFF
--- a/metta/setup/components/aws.py
+++ b/metta/setup/components/aws.py
@@ -2,7 +2,6 @@ import json
 from pathlib import Path
 
 from metta.setup.components.base import SetupModule
-from metta.setup.config import UserType
 from metta.setup.registry import register_module
 from metta.setup.utils import info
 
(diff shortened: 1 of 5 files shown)


berekuk (approved):
lgtm

PR #1284, opened 2025-07-02
specify node version in brewfile

nishu-builder:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210692583383215)


berekuk (approved):
On my end, I plan to continue using nvm; note that it's an [official way to install node.js](https://nodejs.org/en/download) these days. But that's more work.

PR #1285, opened 2025-07-02
feat: deterministically random omegaconf resolvers

rjwalters:
This PR introduces deterministic randomness to the OmegaConf custom resolvers, allowing for reproducible results in configurations that use random sampling. When a seed is provided in the configuration, the resolvers will produce consistent, reproducible random values across runs.

### Changes
- **Added deterministic seeding to `oc_scaled_range` (sampling resolver)**: The function now uses a seeded `np.random.RandomState` when a seed is provided through the `_root_` parameter
- **Extended `oc_uniform` with optional deterministic behavior**: Added `_root_` parameter to access seed value; falls back to non-deterministic behavior when no seed is provided
- **Extended `oc_choose` with optional deterministic behavior**: Similar pattern as `oc_uniform`, using seeded randomness when available
- **Added global counters**: Each resolver maintains its own call counter (`NUM_CALLS_TO_SAMPLING`, `NUM_CALLS_TO_UNIFORM`, `NUM_CALLS_TO_CHOOSE`) to ensure unique seeds for each invocation

### Motivation
Previously, configurations using the `sampling`, `uniform`, and `choose` resolvers would produce different results on each run, making it difficult to:
- Reproduce experimental results
- Debug configuration-dependent issues
- Create consistent test scenarios
- Share reproducible configurations with collaborators

### Implementation Details
The deterministic behavior is achieved by:
1. Using the base seed from the configuration (accessed via `_root_.get("seed", 0)`)
2. Adding a unique counter value for each resolver call to create distinct seeds
3. Creating isolated `RandomState` instances to avoid affecting global random state
4. Maintaining backward compatibility with a fallback to non-deterministic behavior when no seed is provided

### Usage Example
```yaml
# With seed - deterministic behavior
seed: 42
sampling: 0.5
some_param: ${sampling:10,20,15}  # Will always return the same value

# Without seed - non-deterministic behavior (backward compatible)
sampling: 0.5
some_param: ${sampling:10,20,15}  # Will return different values each run
```

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210692656101174)

### DIFF
--- a/common/src/metta/common/util/resolvers.py
+++ b/common/src/metta/common/util/resolvers.py
@@ -11,17 +11,44 @@ from metta.common.util.logging import setup_mettagrid_logger
 T = TypeVar("T")  # For generic conditional function
 Numeric = Union[int, float]  # Type alias for numeric types
 
+# Global counters for deterministic randomness
+NUM_CALLS_TO_SAMPLING = 0
+NUM_CALLS_TO_UNIFORM = 0
+NUM_CALLS_TO_CHOOSE = 0
+
 
 def oc_if(condition: bool, true_value: T, false_value: T) -> T:
     return true_value if condition else false_value
 
 
-def oc_uniform(min_val: Numeric, max_val: Numeric) -> float:
-    return float(np.random.uniform(min_val, max_val))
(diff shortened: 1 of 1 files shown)


rjwalters:
let's do this in curriculum instead!

PR #1286, opened 2025-07-02
[Draft] Improve external collaborator ergonomics with s3, wandb touchpoints

nishu-builder:
Make it easier to disable and to specify non-softmax locations for cloud services (wandb, s3). None of this should change behavior for softmax researchers (where in-code
defaults have changed, trainer.yaml specifies old values)

Torch profiler:
- Consumes a refactored TorchProfilerConfig config, which now defaults to being disabled
- Uses the write_file utility that replays and stats do. This removes its local reliance on boto3 or aws being configured, and also adds support for putting torch profiles in local files (tested) or wandb artifacts
- Changed s3_bucket to upload_dir to support both S3 paths (s3://) and local directories

Simulation replays:
- Was previously defaulting to a softmax s3 bucket in code. Now defaults to ${run_dir}/replays in trainer_config.py, but trainer.yaml overrides this to s3://softmax-public/replays/${run}, so behavior for softmax researchers should be unchanged

Wandb behavior:
- Now requires explicit command-line overrides (wandb=off) for external users due to Hydra config composition rules
- Documents that WANDB_ENTITY environment variable is required for file uploads (separate from wandb.entity config used for metrics)

New external user config:
- configs/user/external_user.yaml provides instructions on how to give custom s3 and wandb connections
- Explains that wandb=off and stats_server_uri=null must be specified at command line (not in user config) due to Hydra's handling of + prefixed configs

Updated README:
- Consolidates existing and new advice on how to configure a custom connection to s3 and wandb into one section
- Clarifies the required command-line arguments for external users to run fully locally
- Documents the WANDB_ENTITY environment variable requirement for file uploads


README.md
```diff
@@ -125,7 +125,7 @@ The repository contains command-line tools in the `tools/` directory. Most of th
 ### Training a Model
 
 ```bash
-./tools/train.py run=my_experiment +hardware=macbook wandb=off +user=<name>
+./tools/train.py run=my_experiment +hardware=macbook wandb=off +user=<name> stats_server_uri=null
```
daveey, line 128: how about we make a +org=softmax profile, just like we have +hardware. we can enable things like the stats_server in there, and put all softmax related settings. that way train.py could just run with s3/wandb/stats_server off by default

we could also auto-generate the run if not provided, then people could really just call "metta train"

metta/rl/torch_profiler.py
```diff
@@ -119,14 +119,25 @@ def _save_profile(self, prof):
 
         output_filename_json = f"{self._profile_filename_base}.json"
         output_filename_gz = f"{output_filename_json}.gz"
-        temp_json_path = os.path.join(self._profile_dir, output_filename_json)  # for uncompressed, to be deleted
-        final_gz_path = os.path.join(self._profile_dir, output_filename_gz)  # compressed path
+
+        # Use temp directory for intermediate files
+        import tempfile
```
daveey, line 124: we now have common.util.fs that has atomic_write(), maybe we can add write_gz_file() in there?

daveey (changes requested):
Looks like this is a draft with merge conflicts, so not sure if you're just looking for feedback. But gave you some ideas

PR #1291, opened 2025-07-02
nit: print newsletter date range in PST

rjwalters:


nishu-builder (approved):
Huh cool didnt know about zoneinfo; have only used pytz before. looks good!

PR #1293, opened 2025-07-02
change freeze_duration to a signed short

bluewin4:
Changes freeze_duration to a signed short and adds logic for negative freeze_duration's to result in a permanent freezing effect. 

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210701959974668)


sasmith:
Ideally we'd have a test about this behavior. I don't know if we do.

PR #1313, opened 2025-07-03
chore: reorganize backend subprojects

rjwalters:
Renames "app backend" to "observatory backend" and creates "studio backend"


common/src/metta/common/util/stats_client_cfg.py
```diff
@@ -4,7 +4,7 @@
 from httpx import Client
 from omegaconf import DictConfig, ListConfig
 
-from metta.app_backend.stats_client import StatsClient
+from metta.observatory_backend.stats_client import StatsClient
```
berekuk, line 7: Can we name it `metta.observatory` instead of `metta.observatory_backend`, for python? it's not like we're going to have `metta.observatory_frontend` python package.

studio_backend/README.md
```diff
@@ -0,0 +1,3 @@
+# Studio Backend
+
+This is the backend for the "Metta Studio" app
```
berekuk, line 3: As we discussed in the DMs, I don't think a separate `studio_backend` is useful.

The reason why `obsevatory_backend` is a separate package is because we need to dockerize it, so we need to separate observatory backend-only python dependencies from all other metta dependencies. This is not the case for studio backend.

Also, the dependency arrow goes in the opposite direction here: studio backend depends on metta package (on hydra, but potentially on anything else in `metta.*`). This means that top-level metta package can't depend on studio, it would create a recursive dependency. The code in this PR probably works, but it violates the semantics of pyproject files, because studio-backend pyproject doesn't declare the metta dependency.

So I think you should move the studio code back to the root package, there's no problem with having it there, afaict.

berekuk (changes requested):
see the comments

common/src/metta/common/util/stats_client_cfg.py
rjwalters: we will have a filesystem conflict on the folder names... we could have `observatory_client` and `observatory` if you prefer

studio_backend/README.md
rjwalters: there will be no more "root package" when we are done -- i can put your server code into metta.common if you prefer?

PR #1317, opened 2025-07-04
Auto-detect parameters related to hardware to remove most configs

nishu-builder:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210713292337220)


relh:
https://github.com/Metta-AI/metta/pull/1239

this PR shows how to put something in a decorator instead of overriding the config. 

I am a big fan of auto-detecting the device, it's always been robust for me in the past.

common/src/metta/common/util/script_decorators.py
```diff
@@ -72,3 +77,65 @@ def wrapper(cfg: DictConfig, *args, **kwargs) -> T:
...
+            cfg.device = "cuda"
+        else:
+            cfg.device = "cpu"
+        logger.debug(f"Auto-detected device: {cfg.device}")
+    elif cfg.device.startswith("cuda") and not cuda_available:
+        logger.warning(f"Device '{cfg.device}' was requested but CUDA is not available. Overriding to 'cpu'.")
+        cfg.device = "cpu"
+
+    # Handle vectorization
+    multiprocessing_available = is_multiprocessing_available()
```
daveey, line 103: Nitpick:

I find it cleaner to first figure out what the user wants, defaulting to multiprocessing:

vec = cfg.get(..., "multiprocessing")

if not multiprocessing_available():
  vec = "serial"

configs/trainer/trainer.yaml
```diff
@@ -1,6 +1,5 @@
 _target_: metta.rl.trainer.MettaTrainer
-
-num_workers: ???
+num_workers: ???  # auto-detected by @metta_script decorator if not specified
```
daveey, line 2: This should only be detected in train.py not in all scripts

configs/trainer/trainer.yaml
```diff
@@ -71,7 +70,7 @@ vtrace:
   vtrace_rho_clip: 1.0  # Default: on-policy (no off-policy correction)
   vtrace_c_clip: 1.0    # Default: on-policy bootstrapping
 
-zero_copy: true
+zero_copy: ???  # auto-detected by @metta_script decorator if not specified
```
daveey, line 73: Let's try to always default to true, and only change it in train.py if we need to

configs/trainer/trainer.yaml
```diff
@@ -88,7 +87,7 @@ profiler:
   profile_dir: s3://softmax-public/torch_traces/${run}
 
 forward_pass_minibatch_target_size: 4096
-async_factor: 2
+async_factor: ???  # auto-detected by @metta_script decorator if not specified
```
daveey, line 90: Let's set to 2, I don't think we need a fallback for this

daveey (changes requested):

configs/trainer/trainer.yaml
nishu-builder: just moved it to train.py; you're right that @metta_script wasn't the right place

daveey:
assigning back,. assign to me when ready to review

tests/rl/test_trainer_config.py
```diff
@@ -94,19 +95,38 @@
 }
 
 
+@patch("metta.common.util.script_decorators.setup_mettagrid_environment", return_value=None)
+@patch("metta.common.util.script_decorators.torch.cuda.is_available", return_value=True)
+@patch("metta.common.util.script_decorators.is_multiprocessing_available", return_value=True)
+def process_cfg_like_metta_script_main(cfg: DictConfig, *_mocks) -> DictConfig:
```
nishu-builder, line 101: we have some tests that aim to check if checked-in configs can be parsed into valid trainer configs. Those checked-in configs now undergo our `set_hardware_configurations` as part of `metta_script`, which gives them e.g. `vectorization` values if they're otherwise unspecified.

So to test these realistically but without making our unit tests too heavy or environment-dependent, I use metta_script here with some parts stubbed out

common/src/metta/common/util/script_decorators.py
```diff
@@ -72,3 +76,41 @@ def wrapper(cfg: DictConfig, *args, **kwargs) -> T:
...
+
+def is_unspecified(cfg: DictConfig, key: str) -> bool:
+    """Check if a config value is unspecified (not set or ???)."""
+    return key not in cfg or OmegaConf.is_missing(cfg, key)
+
+
+def set_hardware_configurations(cfg: DictConfig, logger: logging.Logger) -> None:
+    OmegaConf.set_struct(cfg, False)
+
+    device = cfg.get("device", "cuda")
```
daveey, line 89: Cfg.device is always set

tools/train.py
```diff
@@ -27,13 +28,38 @@ class TrainJob(Config):
...
+
+    # Round down to nearest power of 2
+    num_workers = 1
+    while num_workers * 2 <= ideal_workers:
+        num_workers *= 2
+
+    return max(1, num_workers)
+
+
+def set_num_workers_if_unspecified(cfg: DictConfig | ListConfig) -> None:
```
daveey, line 47: Cfg is always DictConfig (not list config)

tools/train.py
```diff
@@ -27,13 +28,38 @@ class TrainJob(Config):
...
+    if isinstance(cfg, DictConfig) and "trainer" in cfg:
```
daveey, line 48: This is always true

tools/train.py
```diff
@@ -27,13 +28,38 @@ class TrainJob(Config):
...
+        OmegaConf.set_struct(cfg, False)
+        if OmegaConf.is_missing(cfg.trainer, "num_workers"):
+            cfg.trainer.num_workers = _calculate_default_num_workers(cfg.get("vectorization") == "serial")
```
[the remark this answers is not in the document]
daveey: cfg.vectorization

Don't use get unless you expect it to be not set

daveey (changes requested):

daveey (approved):

PR #1318, opened 2025-07-04
Show metadata and map labels in studio

berekuk:
Follow-up to labels feature from #1258.

Also, this PR adds auto-reload to the studio backend, and improves error handling a little bit.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210713581106032)

### DIFF
--- a/metta/map/utils/storable_map.py
+++ b/metta/map/utils/storable_map.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import time
 from dataclasses import dataclass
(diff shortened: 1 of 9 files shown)


sasmith (approved):
For this diff, I'd love a sentence or two about how you tested this. That's probably true for many diffs, but on this one I was more aware of the question as I looked across the typescript.

berekuk:
> For this diff, I'd love a sentence or two about how you tested this. That's probably true for many diffs, but on this one I was more aware of the question as I looked across the typescript.

Mostly just by looking at the UI. Next.js has hot reload so the feedback loop is very fast, and I'm not sure there's much value in unit tests for features like this (and changes in python in this PR mostly just move the code around).

Screenshot 2025-07-08 at 12 23 00 AM

PR #1308, opened 2025-07-03
feat: turn on more c++ compile time warnings

rjwalters:
This PR enables comprehensive compile-time warnings for the C++ codebase to improve code quality and catch potential issues early. 

**The build currently generates ~50 warnings that could be addressed in follow-up PRs.**

### 1. Enhanced Compiler Warnings
Added extensive warning flags for GCC, Clang, and MSVC compilers including:
- **Type safety**: `-Wconversion`, `-Wsign-conversion`, `-Wcast-align`, `-Wcast-qual`
- **Code quality**: `-Wshadow`, `-Wold-style-cast`, `-Wpedantic`
- **Safety checks**: `-Wnull-dereference`, `-Wfloat-equal`, `-Wdouble-promotion`
- **Object-oriented best practices**: `-Woverloaded-virtual`, `-Wnon-virtual-dtor`
- **Compiler-specific warnings**: Thread safety analysis for Clang, logical operation checks for GCC

### 2. Build Configuration Fixes
- Fixed CMake to properly detect and use Python from virtual environments
- Added `Python3_FIND_VIRTUALENV FIRST` to prioritize venv Python over system Python
- Fixed build directory configuration for the production build target

### 3. Debug Support
- Added `-fno-omit-frame-pointer` for better debugging and profiling capabilities

## Notes
- **Future work**: These warnings should be addressed incrementally to improve code quality
- **No runtime impact**: These are compile-time checks only

The warnings are intentionally set as non-fatal to allow gradual improvement of the codebase. Future PRs should address these warnings by category (e.g., sign conversions, shadowing, etc.).

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210712959986625)

PR #1324, opened 2025-07-04
feat: cpp compilation warnings

rjwalters:
(this pr incrementally applies patches from #1308)

This PR enables comprehensive compile-time warnings for the C++ codebase to improve code quality and catch potential issues early. 

**The build currently generates ~600 warnings that could be addressed in follow-up PRs.**

### 1. Enhanced Compiler Warnings
Added extensive warning flags for GCC, Clang, and MSVC compilers including:
- **Type safety**: `-Wconversion`, `-Wsign-conversion`, `-Wcast-align`, `-Wcast-qual`
- **Code quality**: `-Wshadow`, `-Wold-style-cast`, `-Wpedantic`
- **Safety checks**: `-Wnull-dereference`, `-Wfloat-equal`, `-Wdouble-promotion`
- **Object-oriented best practices**: `-Woverloaded-virtual`, `-Wnon-virtual-dtor`
- **Compiler-specific warnings**: Thread safety analysis for Clang, logical operation checks for GCC

### 2. Build Configuration Fixes
- Fixed CMake to properly detect and use Python from virtual environments
- Added `Python3_FIND_VIRTUALENV FIRST` to prioritize venv Python over system Python
- Fixed build directory configuration for the production build target

### 3. Debug Support
- Added `-fno-omit-frame-pointer` for better debugging and profiling capabilities

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210715087116790)


mettagrid/tests.cmake
```diff
@@ -64,6 +68,54 @@ function(mettagrid_add_tests GLOB_PATTERN # e.g. "${CMAKE_CURRENT_SOURCE_DIR}/te
 
     target_link_libraries(${output_name} PRIVATE ${LINK_LIBS})
 
+    # Debug-only runtime sanitizers
+    target_compile_options(mettagrid_obj PRIVATE
```
sasmith, line 72: This is duplicative with CMakeLists.txt.

My cmake knowledge isn't good, so I can't tell if the right answer is to accept this, or if we should be consolidating the flags.

My default, I would have expected for tests to be build and run using the settings in CMakeLists, but that may represent a misunderstanding.

sasmith: I guess: can you confirm that what we're doing makes sense? Or maybe @berekuk has opinions here?

I don't think we need perfect, but I want to steer away from what may be over-verbosity.

berekuk: I don't possess the deep knowledge of cmake either, and when I did my initial PR I couldn't afford allocating an extra day to read its documentation thoroughly.

So I talked a lot with chatgpt, asking it to self-criticize when it proposed something, and tried to figure out how I could condense the configuration into the minimal version, then I made many iterations on that.

I recommend doing the same here. My hunch is that many `-W` could be collapsed under `-Wall -Wextra`, etc., and the repetitive code could be moved to a shared reusable function.

rjwalters:
> General though: Right now this appear not to show warnings when things are compiled using `uv`. Because of this, I think it's more okay to merge in before fixing all the warnings, but I want a way for folks (me) to set things up so that the warnings _will_ show up when we run with `uv`. Maybe my thinking about this is wrong, since caching means we won't hit these warning again?
> 
> Requesting changes: `make test` in `mettagrid` fails for me locally. My assumption is that something slipped through with flags not being right on os x, but they're right on ubuntu. This should be fixed before merge.


make test fails on main for me so I think there is a deeper problem to be fixed? I think the benchmark file needs to be updated to the new config system

I will explore options for DRY between cmake files

sasmith (approved):

PR #1331, opened 2025-07-05
Validate all env configs; fix game/objects in all env configs

berekuk:
Initial test output before `configs/` updates: 76 failed, 61 passed.

Most fixes are about mounting various `/env/mettagrid/game/objects/*`, but there are also some fixes for `heart_max: 1000` which was out of allowed bounds.

The patch is somewhat verbose, and maybe the better solution would be to:
- bring back all objects in `mettagrid.yaml`?
- or add `mettagrid_full.yaml` with all objects but without a map?

We can do these later though, my main goal here is to fix things that are obviously broken.

---

Note that this test doesn't guarantee that the env will run. I validate by calling `cpp_config_dict`, and maybe I should've instantiated the full `MettagridEnv` for each config instead. (Would this validate the maps too? I'm pretty sure we still have some envs that don't configure objects but use maps that include those unconfigured objects)

But I ran out of time for today and wanted to get this in quickly.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210719699897175)


configs/env/mettagrid/ants.yaml
```diff
@@ -6,8 +6,9 @@ defaults:
       - mines
   - _self_
 
+group_reward_pct: 0
```
berekuk, line 9: unknown fields under `game` are forbidden now

configs/env/mettagrid/laser_tag.yaml
```diff
@@ -74,3 +75,9 @@ game:
       input_ore_red: 0
       initial_items: 5
       cooldown: 5
+
+  groups:
+    team_1:
+      group_reward_pct: ${....group_reward_pct}
+    team_2:
+      group_reward_pct: ${....group_reward_pct}
```
berekuk, line 83: I'm assuming these were removed by accident when David extracted `game.groups`, they were there before

tests/test_validate_all_envs.py
```diff
@@ -0,0 +1,67 @@
...
+@pytest.mark.parametrize("cfg_metadata", map_or_env_configs(), ids=[cfg.path for cfg in map_or_env_configs()])
+class TestValidateAllEnvs:
+    def test_map(self, cfg_metadata):
+        try:
+            map_cfg = cfg_metadata.get_cfg().get_map_cfg()
+            map_builder_cfg_to_storable_map(map_cfg)
+        except (NoCredentialsError, ProfileNotFound) as e:
+            pytest.skip(f"Skipping {cfg_metadata.path} because it requires AWS credentials: {e}")
+        except Exception as e:
+            pytest.fail(f"Failed to validate map config {cfg_metadata.path}: {e}")
```
berekuk, line 52: This test is from `tests/map/test_validate_all_env_maps.py` that I renamed. Maybe it could be a separate test, but there's some overlap in the shared fixture.

Map tests are slower than config tests, but I managed to iterate quickly by running `pytest -k test_mettagrid_config` (and, sometimes, `pytest -k test_mettagrid_config[specific_config_name.yaml]`)

configs/env/mettagrid/ants.yaml
daveey: you can remove this field

daveey (approved):

configs/env/mettagrid/ants.yaml
berekuk: this config refers to this field later though, same as `laser_tag.yaml`

PR #1340, opened 2025-07-06
Hyperparameter scheduler

harshbhatt7585:
img

One Hyperparameter Scheduler class schedules hyperparameter values based on agent steps.

The following hyperparameters are scheduled in this class:
`learning_rate`

Starting with a higher learning rate allows for faster initial convergence and exploration of the parameter space, while gradually reducing it helps fine-tune the model and achieve better final performance.

`ppo_clip_coef`

The PPO clip coefficient controls how much the policy can change in a single update. Early in training, larger policy updates can help with exploration and faster learning. As the policy stabilizes, smaller clips prevent destructive updates and maintain training stability.

`ppo_vf_clip_coef`

Similar to the policy clip, the value function clip prevents large value function updates that could destabilize training. The value function typically needs more aggressive updates early on to catch up with policy changes, then smaller updates for refinement.

`ppo_ent_coef`

Entropy regularization encourages exploration by penalizing overly deterministic policies. High entropy is crucial early in training for exploration, but should be reduced as the agent learns to avoid suboptimal random actions.

`ppo_l2_reg_loss_coef`

L2 regularization prevents overfitting by penalizing large weights. Early in training, the focus should be on learning useful representations, so strong regularization can be counterproductive. As training progresses and overfitting becomes a concern, increasing regularization helps generalization.

`ppo_l2_init_loss_coef`

This coefficient penalizes deviation from initial parameter values, which can help with transfer learning or prevent catastrophic forgetting. The importance of staying close to initialization typically decreases as the model learns task-specific features.


configs/wandb/external_user.yaml
```diff
@@ -1,7 +1,7 @@
 enabled: true
 
 project: metta
-entity: ???  # Replace with your WandB username
+entity: harshbhatt7585  # Replace with your WandB username
```
daveey, line 4: don't commit this change

metta/rl/hyperparameter_scheduler.py
```diff
@@ -0,0 +1,153 @@
...
+            "decay_rate": schedule.get("decay_rate", 0.1),
+        }
+
+    def _compute_scheduled_value(self, param_name: str, current_step: int) -> float:
+        """Compute the scheduled value for a hyperparameter at the current step."""
+        initial_value = self.initial_values[param_name]
+        schedule = self.schedules[param_name]
+        schedule_type = schedule["type"]
+
+        if schedule_type == "constant":
```
daveey, line 115: why doesn't this get handled by the ConstantSchedule

metta/rl/hyperparameter_scheduler.py
```diff
@@ -0,0 +1,153 @@
...
+            return self.initial_value
+        log_progress = math.log1p(self.decay_rate * progress) / math.log1p(self.decay_rate)
+        return self.initial_value + (self.min_value - self.initial_value) * log_progress
+
+
+
+
+class HyperparameterScheduler:
+
+    scheduler_registry: Dict[str, Callable[..., BaseSchedule]] = {
```
daveey, line 66: consider using hydra.instanciate() instead of your own registry. but i don't feel strongly about this pattern

metta/rl/hyperparameter_scheduler.py
```diff
@@ -0,0 +1,153 @@
...
+        "constant": ConstantSchedule,
+        "linear": LinearSchedule,
+        "cosine": CosineSchedule,
+        "exponential": ExponentialSchedule,
+        "logarithmic": LogarithmicSchedule,
+    }
+
+    def __init__(self, trainer_cfg: DictConfig, total_timesteps: int, logging):
```
daveey, line 74: this could take trainer, that way you don't need to pass it in during step() which i think is a bit cleaner

metta/rl/trainer.py
```diff
@@ -274,11 +278,11 @@ def __init__(
                     f"Environment observation shape: {environment_shape}"
                 )
 
-        self.lr_scheduler = None
-        if trainer_cfg.lr_scheduler.enabled:
-            self.lr_scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(
-                self.optimizer, T_max=trainer_cfg.total_timesteps // trainer_cfg.batch_size
-            )
+        # self.lr_scheduler = None
```
daveey, line 281: just remove these, no need to leave in the code

daveey (changes requested):
This looks like a nice clean PR. Left some minor change requests, but should be good to go after you make those.

metta/rl/hyperparameter_scheduler.py
harshbhatt7585: Nice, loading the regitry through hydra now

metta/rl/hyperparameter_scheduler.py
harshbhatt7585: Fixed it.

metta/rl/trainer.py
harshbhatt7585: Done!

relh:
This looks great, feel free to make more changes if you need. I am going to run a compare+contrast on a few of our experimental setups to see if the decay added here helps or hurts performance. 

We can also add the feature and have everything use a Constant schedule by default.

relh (approved):

PR #1342, opened 2025-07-07
extract policy evaluation into standalone module

nishu-builder:


metta/eval/policy_evaluator.py
```diff
@@ -0,0 +1,97 @@
...
+        self.evals: dict[str, float] = {}
+        categories: Set[str] = set()
+
+        for sim_name in self.sim_suite_config.simulations.keys():
+            categories.add(sim_name.split("/")[0])
+
+        for category in categories:
+            score = stats_db.get_average_metric_by_filter("reward", policy_record, f"sim_name LIKE '%{category}%'")
+            logger.info(f"{category} score: {score}")
+
```
nishu-builder, line 79: I removed heartbeating here

I'm not sure it's necessary to do this here for each category (it would be done very frequently (in between each sql query; shouldn't take that long)

But moreover, I dont think it's necessary to do it in policy evaluation at all. trainer does this every 10 epochs by default. This heartbeating would run more infrequently: every `evaluate_interval` (300 by default)

metta/eval/policy_evaluator.py
```diff
@@ -0,0 +1,97 @@
...
+        sim_suite_config: SimulationSuiteConfig,
+        policy_store: PolicyStore,
+        stats_client: Optional[StatsClient] = None,
+    ):
+        self.sim_suite_config = sim_suite_config
+        self.policy_store = policy_store
+        self.device = torch.device(device)
+        self.vectorization = vectorization
+        self.evals: dict[str, float] = {}  # used for wandb
+        self.stats_dir = stats_dir
```
daveey, line 32: Let's rename members to self._*

daveey (approved):

PR #1344, opened 2025-07-07
Gitignore local-only vscode settings

nishu-builder:
we have .vscode/settings.json for anything we want to check in 

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210724568429312)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .env
 .task
 .vscode/PythonImportHelper-v2-Completion.json
+.vscode/settings.local.json
 /checkpoints/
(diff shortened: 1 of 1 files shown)


berekuk:
Are you sure this feature exists?
https://github.com/microsoft/vscode/issues/40233 is open, and I can't find any mentions in vscode documentation.

If I'm right and it doesn't exist, then what we could do instead is move all `settings.json` content to `metta.code-workspace`, and leave `.vscode/settings.json` for local customizations.

PR #1351, opened 2025-07-07
Add MPS-Compatible Fallback for compute_advantage

harshbhatt7585:
This PR adds support for Apple Silicon (MPS backend) by replacing the custom operator torch.ops.pufferlib.compute_puff_advantage with a native PyTorch implementation only when running on MPS.

### DIFF
--- a/metta/rl/functions.py
+++ b/metta/rl/functions.py
@@ -239,34 +239,49 @@ def compute_advantage(
     vtrace_c_clip: float,
     device: torch.device,
 ) -> Tensor:
-    """CUDA kernel for puffer advantage with automatic CPU fallback.
-
-    This matches the trainer.py implementation exactly.
-    """
-    # Get correct device
     device = torch.device(device) if isinstance(device, str) else device
 
-    # Move tensors to device and compute advantage
-    tensors = [values, rewards, dones, importance_sampling_ratio, advantages]
-    tensors = [t.to(device) for t in tensors]
-    values, rewards, dones, importance_sampling_ratio, advantages = tensors
+    if str(device) == "mps":
+        # Native PyTorch implementation (MPS-compatible)
+        values = values.to(device)
+        rewards = rewards.to(device)
+        dones = dones.to(device)
+        importance_sampling_ratio = importance_sampling_ratio.to(device)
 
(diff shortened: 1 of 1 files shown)


metta/rl/functions.py
```diff
@@ -239,34 +239,61 @@ def compute_advantage(
...
-            dones,
-            importance_sampling_ratio,
-            advantages,
-            gamma,
-            gae_lambda,
-            vtrace_rho_clip,
-            vtrace_c_clip,
-        )
+    if str(device) == "mps":
+        # Native PyTorch implementation (MPS-compatible)
```
daveey, line 249: lets move this into metta/rl/mps.py:advantages()

metta/rl/functions.py
```diff
@@ -239,34 +239,61 @@ def compute_advantage(
...
+        lastpufferlam = torch.zeros(B, device=device)
+
+        for t in range(T - 2, -1, -1):
+            lastpufferlam = delta[t] + gamma_lambda * c[t] * lastpufferlam * nextnonterminal[t]
+            advantages[t] = lastpufferlam
+
+        return advantages
+
+    else:
+        # CUDA implementation using custom kernel
```
daveey, line 277: let's not have this in the else: clause

if mps.is_mps(device):
  return mps.advantages(....)

previous cuda code

daveey (changes requested):
Thanks for this PR. A few small code org suggestions, but looks good otherwise.

metta/rl/functions.py
harshbhatt7585: make sense, thanks, done!

metta/rl/functions.py
harshbhatt7585: done in the latest commit

daveey (approved):

harshbhatt7585:
> Looks great! Sorry I missed this.
> 
> `compute_puff_advantage` already has a CPU version and a CUDA version. Happy to add this but not sure if we need it?
> 
> Maybe you can explain?

I was trying to run through GPU on Mac, it only have CPU and CUDA. CPU implementation of puff is slower compared to running over MPS.

PR #1353, opened 2025-07-07
Log extra observatory timing

Barraketh:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210734658206430)


app_backend/tests/generate_test_data.py
berekuk, line 1: update shebang to `#!/usr/bin/env -S uv run`

app_backend/src/metta/app_backend/metta_repo.py
```diff
@@ -516,15 +517,18 @@ def validate_machine_token(self, token: str) -> str | None:
...
                 RETURNING user_id
-                """,
+                """
+            result = execute_single_row_query_and_log(
+                con,
+                query,
                 (token_hash,),
-            ).fetchone()
+                "validate_machine_token",
+            )
```
berekuk, line 531: This this specific benchmark? I don't expect it to ever be slow,`machine_tokens` table is tiny.

If this is from observations (lol) that observatory backend becomes unresponsive, my first guess is that we're running out of python threads somewhere else.

berekuk (approved):
I don't see any problems but also don't understand what you're trying to achieve

PR #1354, opened 2025-07-07
Add MettaScope tests

monofuel:
- add some very simple smoke tests for testing the MettaScope frontend.
- test loading the page, test loading a replay, and test loading a replay and waiting for it to start playing.

- add a test workflow using Playwright to test MetaScope in CI
- enable all the magic flags for webGPU on chromium
- add an AGENTS.md file to help LLMs work with metascope

- this can be tested locally with `act pull_request -W .github/workflows/test-mettascope.yml` (assuming nektos/act is installed)


mettascope/tests/smoke.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  const consoleErrors: string[] = [];
+  await page.goto(`http://localhost:2000/?replayUrl=${encodeReplayUrl(replayUrl)}`);
+  
+  // Wait for the page to fully load the replay and render the first frame
+  await page.waitForFunction(() => {
+    const state = (window as any).state;
+    return state && state.replay !== null;
+  }, { timeout: 10000 });
+  expect(consoleErrors).toHaveLength(0);
+});
```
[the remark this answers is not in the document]
monofuel: Oops! good catch, bad copy paste

treeform (approved):

.github/workflows/test-mettascope.yml
```diff
@@ -0,0 +1,62 @@
+name: "Mettascope UI Test (Headless Chrome)"
+
+on:
+  pull_request:
```
daveey, line 4: This will test if mettascope is broken or can't handle the s3 replay file. But it won't handle changes to the replay format.

How about we run replay.py, generate the replay, and then pass that to mettascope? That way we don't need a human to update this test or generate new replays.

.github/workflows/test-mettascope.yml
```diff
@@ -0,0 +1,62 @@
...
+          npm install --no-save playwright
+          npx playwright install --with-deps chromium
+
+      - name: Build Mettascope
+        run: |
+          cd mettascope
+          npm ci
+          npx tsc
+
+      - name: Install uv
```
daveey, line 35: i think we have an action to setup uv

.github/workflows/test-mettascope.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: astral-sh/setup-uv@v6
+        with:
+          version: "0.7.3"
+          enable-cache: true
+
+      - name: Generate Atlas
```
daveey, line 41: should we make a mettascope-build action and move all the setup logic in there?

daveey (changes requested):
This is a great step, but I think with a bit more work it can be much better.

.github/workflows/test-mettascope.yml
monofuel: Yea that sounds like a good idea, rather than doing the setup commands piece by piece

.github/workflows/test-mettascope.yml
monofuel: I borrowed this from deploy-mettascope.yml, let me know if we had a better action for this

.github/workflows/test-mettascope.yml
monofuel: That's a very good point!

might be worth removing replay tests from this PR and instead make a new PR focused on better replay testing.

mettascope/src/main.ts
```diff
@@ -504,6 +504,10 @@ async function parseUrlParams() {
   // Load the replay.
   const replayUrl = urlParams.get('replayUrl')
   const wsUrl = urlParams.get('wsUrl')
+  if (urlParams.get('play') !== null) {
+    // if the play parameter is set, we should preserve it when loading a replay
+    state.isPlaying = urlParams.get('play') === 'true'
```
monofuel, line 509: This feels weird- I wanted to test that the replay can actually start, but loading the ws url was resetting all the query params because we don't load them unless we have a replay.

.github/workflows/test-mettascope.yml
monofuel: I've updated the test to generate a replay (using a very simple model only trained for 10 steps) instead of using one from S3.

mettascope/src/common.ts
```diff
@@ -109,7 +109,10 @@ export const state = {
   // Playing over a WebSocket
   ws: null as WebSocket | null,
   isOneToOneAction: false,
-}
+};
+
+// Expose state for easier testing
+(window as any).state = state
```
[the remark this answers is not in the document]
treeform: It's a little strange but I have done this exact thing many times.

.github/workflows/test-mettascope.yml
treeform: You could try using the mettascope/install.sh

mettascope/src/main.ts
treeform: Actually ignore this, you made it work. All good.

treeform (approved):

.github/workflows/test-mettascope.yml
monofuel: updated!

monofuel:
I think the asana issue is happening because I'm merging from monofuel:monofuel/mettascope ( a fork)

now that I'm a contributor on the project I'll push the branch up and try again, will have to make a new PR

monofuel:
new PR that is passing properly: https://github.com/Metta-AI/metta/pull/1390

PR #1357, opened 2025-07-07
Gatekeep Hydra to the ./tools/ layer for readability v0

relh:
https://app.asana.com/1/1209016784099267/project/1210730620223091/task/1210730955259252

### DIFF
--- a/agent/src/metta/agent/metta_agent.py
+++ b/agent/src/metta/agent/metta_agent.py
@@ -2,13 +2,13 @@ import logging
 from typing import TYPE_CHECKING, Optional, Union
 
 import gymnasium as gym
-import hydra
 import numpy as np
 import torch
 from omegaconf import DictConfig, ListConfig, OmegaConf
 from torch import nn
 from torch.nn.parallel import DistributedDataParallel
 
+from metta.agent.component_factory import ComponentFactory
 from metta.agent.policy_state import PolicyState
 from metta.agent.util.debug import assert_shape
 from metta.agent.util.distribution_utils import evaluate_actions, sample_actions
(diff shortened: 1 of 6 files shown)


agent/src/metta/agent/metta_agent.py
```diff
@@ -29,18 +29,23 @@ def make_policy(env: "MettaGridEnv", cfg: ListConfig | DictConfig) -> "MettaAgen
         }
     )
 
-    # Here's where we create MettaAgent. We're including the term MettaAgent here for better
-    # searchability. Otherwise you might only find yaml files.
-    return hydra.utils.instantiate(
-        cfg.agent,
+    # Extract agent config
+    agent_cfg = cfg.agent if hasattr(cfg, "agent") else cfg
```
daveey, line 33: When would cfg.agent not be set ?

agent/src/metta/agent/metta_agent.py
```diff
@@ -29,18 +29,23 @@ def make_policy(env: "MettaGridEnv", cfg: ListConfig | DictConfig) -> "MettaAgen
...
+    return MettaAgent(
         obs_space=obs_space,
         obs_width=env.obs_width,
         obs_height=env.obs_height,
         action_space=env.single_action_space,
         feature_normalizations=env.feature_normalizations,
         global_features=env.global_features,
-        device=cfg.device,
-        _recursive_=False,
+        device=cfg.device if hasattr(cfg, "device") else "cuda",
```
daveey, line 47: Cfg.device is required

metta/rl/policy.py
```diff
@@ -43,8 +42,35 @@ def load_pytorch_policy(path: str, device: str = "cpu", pytorch_cfg: DictConfig
         ),
     )
 
-    policy = instantiate(pytorch_cfg, env=env, policy=None)
+    # Import the policy class dynamically based on _target_ field
```
daveey, line 45: Should we do metta.common.util.instantiate() or something?

daveey (changes requested):
I like where this is heading. Do you think we could just use importlib to create class instances without registries? Basically a light-weight hydra.instanciate?

relh:
tested on local, mettabox, 4-gpu sandbox

common/src/metta/common/util/instantiate.py
```diff
@@ -0,0 +1,109 @@
...
+import importlib
+import logging
+from typing import Any, Dict
+
+from omegaconf import DictConfig, OmegaConf
+
+logger = logging.getLogger(__name__)
+
+
+def instantiate(config: Dict[str, Any] | DictConfig, *, _recursive_: bool = False, **kwargs) -> Any:
```
berekuk, line 17: Recursive mode is only used for maps, right? If so, soon (by the end of this week) it won't be necessary, mapgen doesn't need it.

PR #1360, opened 2025-07-08
Make all db queries async

Barraketh:


app_backend/src/metta/app_backend/metta_repo.py
```diff
@@ -184,26 +186,52 @@
...
+        self._pool: AsyncConnectionPool | None = None
+        # Run migrations synchronously during initialization
         with Connection.connect(self.db_uri) as con:
             run_migrations(con, MIGRATIONS)
 
-    def connect(self) -> Connection:
-        return Connection.connect(self.db_uri)
+    async def _ensure_pool(self) -> AsyncConnectionPool:
+        if self._pool is None:
+            self._pool = AsyncConnectionPool(self.db_uri, min_size=2, max_size=10, open=False)
```
nishu-builder, line 196: i bet we can handle more than 10 concurrent queries :)

nishu-builder: I could see us wanting these params to be configurable via `.env` so that they're different by default when running locally

Barraketh: Although I don't know when we would - in an ideal world, the queries are short enough that we don't need to keep open connections for them

nishu-builder: true. i experienced some that took a while but that was before your upgrade :)

PR #1372, opened 2025-07-08
Track environment configurations in WandB artifacts

bluewin4:
Adds tracking of configs in the artifacts logged by weights and biases. Updates config dynamically when new ones are used in curriculum learning and provides the first and last config used. 

Weights and biases needs to be updated so that it doesn't delete these within a month of being made.

### DIFF
--- a/docs/wandb_env_config_tracking.md
+++ b/docs/wandb_env_config_tracking.md
@@ -1,0 +1,136 @@
+# Tracking Environment Configurations in W&B
+
+## Problem Statement
+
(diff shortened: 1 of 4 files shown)


nishu-builder (commented):
Thinking aloud a bit here --

We already send over the full config as `wandb.config`, and the thing that changes over time is the selected curriculum. is there a cleaner way to key that than to manually form the env we upload as an artifact in a fully-expanded way?

relh:
Good point Nishad. Once you sort that out Jack I'll take a look

PR #1378, opened 2025-07-08
feat: expand mettagrid tests

rjwalters:
wip

### DIFF
--- a/app_backend/src/metta/app_backend/metta_repo.py
+++ b/app_backend/src/metta/app_backend/metta_repo.py
@@ -3,7 +3,7 @@ import secrets
 import uuid
 from contextlib import asynccontextmanager
 from datetime import datetime, timedelta
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Tuple
 
 from psycopg import Connection
 from psycopg.types.json import Jsonb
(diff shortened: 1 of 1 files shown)


mettagrid/src/metta/mettagrid/mettagrid_c_config.py
```diff
@@ -75,14 +75,14 @@ def from_mettagrid_config(mettagrid_config_dict: dict[str, Any]) -> GameConfig_c
...
-                color=object_config.color,
+                color=object_config.color or 0,
             )
             object_configs[object_type] = converter_config_cpp
         elif isinstance(object_config, WallConfig_py):
             wall_config = WallConfig_cpp(
                 type_id=object_config.type_id,
                 type_name=object_type,
-                swappable=object_config.swappable,
+                swappable=object_config.swappable or False,
```
sasmith, line 85: let's put this in the pydantic.

In general, prefer to put default values in the pydantic vs in code. I don't think that's always possible, but let's aim for it when it is.

mettagrid/tests/test_mettagrid.py
```diff
@@ -1,334 +1,603 @@
...
+
+
+@dataclass
+class EnvConfig:  # Renamed from TestConfig to avoid pytest confusion
+    NUM_AGENTS = 2
+    OBS_HEIGHT = 3
+    OBS_WIDTH = 3
+    NUM_OBS_TOKENS = 100
+    OBS_TOKEN_SIZE = 3
+    GLOBAL_TOKEN_LOCATION = 17  # 0x11
```
sasmith, line 30: Let's explicitly note that this comes from the OBS_HEIGHT and OBS_WIDTH. Or maybe even generate it with PackedCoordinate.

sasmith (approved):
Scanned, but didn't read the tests in depth. I have some concern about how fragile they'll be, but as long as we have a low bar for ripping and re-writing them when we figure that out, it seems fine.

PR #1382, opened 2025-07-09
Robb/0708 gridcoord type

rjwalters:
more type safety in mettagrid

### DIFF
--- a/mettagrid/src/metta/mettagrid/actions/attack.hpp
+++ b/mettagrid/src/metta/mettagrid/actions/attack.hpp
@@ -68,10 +68,14 @@ protected:
   }
 
   bool _handle_target(Agent* actor, GridLocation target_loc) {
-    target_loc.layer = GridLayer::Agent_Layer;
-    Agent* agent_target = static_cast<Agent*>(_grid->object_at(target_loc));
+    GridObject* obj = _grid->object_at(target_loc);
+    if (!obj) return false;
 
-    bool was_frozen = false;
+    Agent* agent_target = dynamic_cast<Agent*>(obj);
+    if (!agent_target) return false;
+
+    bool was_frozen = agent_target->frozen > 0;
+
     if (agent_target) {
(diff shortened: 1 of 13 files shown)


mettagrid/src/metta/mettagrid/actions/attack_nearest.hpp
sasmith, line 1: Dave is combining attack and attack nearest, so you should expect conflicts here.

mettagrid/src/metta/mettagrid/objects/metta_object.hpp
sasmith, line 1: yay

mettagrid/src/metta/mettagrid/actions/attack.hpp
```diff
@@ -39,32 +38,54 @@ class Attack : public ActionHandler {
...
         }
-        num_skipped++;
+        num_found++;
       }
     }
+
     // If we got here, it means we skipped over all the targets. Attack the last one.
-    if (target) {
-      return _handle_target(*actor, *target);
+    if (last_agent != nullptr && arg >= num_found) {
```
sasmith: the second half of the check feels more risky than good. If we're here, we didn't return early.

mettagrid/src/metta/mettagrid/actions/attack.hpp
```diff
@@ -39,32 +38,45 @@ class Attack : public ActionHandler {
...
+          num_skipped++;
         }
-        num_skipped++;
       }
     }
+
     // If we got here, it means we skipped over all the targets. Attack the last one.
-    if (target) {
-      return _handle_target(*actor, *target);
+    if (last_agent != nullptr && arg >= num_skipped) {
```
sasmith, line 76: the second half of the check feels more risky than good. If we're here, we didn't return early.

sasmith (approved):

PR #1384, opened 2025-07-09
chore: cpp unused variable warnings

rjwalters:
This pull request addresses several `-Wunused-parameter` and `-Wunused-variable` compiler warnings to improve code hygiene and reduce build log noise.

Resolving these warnings ensures that the build output is clean, making it easier to spot new, more critical issues in the future.

### Changes

* **Handle Unused Function Parameters:**
    * Added the `[[maybe_unused]]` attribute to parameters that are required by an overridden virtual function's signature but are not used in the specific implementation.
    * This was applied to the `_handle_action` method in `AttackNearest`, `GetOutput`, `Noop`, `PutRecipeItems`, and `Swap`.
    * It was also applied to the `handle_event` method in `ProductionHandler` and `CoolDownHandler`.

* **Handle Unused Local Variable:**
    * In `mettagrid_c.cpp`, the local variable `observation_view` inside `_compute_observations` was unused, causing a `-Wunused-variable` warning.
    * The line has been commented out to resolve the warning.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210749349080014)


mettagrid/src/metta/mettagrid/actions/attack_nearest.hpp
```diff
@@ -17,7 +17,7 @@ class AttackNearest : public Attack {
   }
 
 protected:
-  bool _handle_action(Agent* actor, ActionArg arg) override {
+  bool _handle_action(Agent* actor, [[maybe_unused]] ActionArg arg) override {
```
sasmith, line 20: Let's do

```
void _handle_action(Agent* actor, ActionArg /*arg*/) override {
...
```
instead,

sasmith (approved):

mettagrid/src/metta/mettagrid/mettagrid_env.py
```diff
@@ -86,7 +86,8 @@ def __init__(
         self._replay_writer = replay_writer
         self._episode_id: str | None = None
         self._reset_at = datetime.datetime.now()
-        self._current_seed = 0
+        self._current_seed: int = 0
+        assert self._current_seed >= 0
```
sasmith, line 90: We just set it to zero. I think we don't need this assert.

sasmith (approved):

PR #1385, opened 2025-07-09
Flat REPO_REORGANIZATION plan

relh:
A flat version of our repository that keeps semantics at the top-level for context/scoping clarity. 

Imports will still look like `import metta.cogworks` but instead of folder structure such as `metta/cogworks/src/metta/cogworks` to accomplish this it will be via symlink.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210757747553588)


REPO_ORGANIZATION.md
```diff
@@ -0,0 +1,336 @@
...
+metta-agent                   → (merged into metta-cogworks)
+metta-mettagrid              → metta-mettagrid
+metta-app-backend            → (split into respective apps)
+```
+
+## Installation Examples
+
+```bash
+# Just the environment
+pip install metta-mettagrid
```
rjwalters, line 260: it feels strange to be holding onto "metta" here.. why not just "softmax-*"?

relh: yeah good point, lets change all these project names to softmax-X, though we're punting on renaming the repo just yet.

REPO_ORGANIZATION.md
```diff
@@ -0,0 +1,429 @@
...
+- Clear company branding for published packages (using `softmax-` prefix)
+- Minimal code changes required (existing `metta.*` imports work as-is)
+
+### Goals of the Flattened Structure
+
+1. **Independent Packages**: Each component is its own Python package with distinct namespaces
+2. **Simple Imports**: Direct package names without deep nesting
+3. **Clear Boundaries**: Each package has a focused purpose and minimal dependencies
+4. **Consistent Branding**: All Python packages use `softmax-` prefix for PyPI distribution
+5. **Developer Ergonomics**: Simple directory structure with enforced import consistency
```
jacklionheart, line 34: Does anyone know if this structure will support "Go to Definition" in VSCode/Cursor? That worked for me 6 weeks ago but does not since rejoining.

relh: It should, this is something we've also been emphasizing along with moving to typed configs instead of YAML (can click through to where a value is set (or at least a default) in a typed config). 

If this plan doesn't support it, the backup with `src/` in the python packages should and we'd switch to that

PR #1388, opened 2025-07-09
refactor: improve naming consistency in MettaGrid config

rjwalters:
This PR improves code readability by establishing clear naming conventions for Python and C++ types in the mettagrid configuration system. 

### Changes
- Renamed all C++ imports to use `Cpp` prefix (e.g., `GameConfig` → `CppGameConfig`)
- Added `Py` prefix to all Python config classes for consistency with existing `PyGameConfig`
- Standardized variable naming throughout the conversion function:
  - Variables containing C++ types/params use `cpp_` prefix
  - Python config objects remain unprefixed (as they're the default in a Python file)
  - Simplified loop variables from generic names to `cfg` for clarity

### What's NOT changed
- No changes to actual code execution or logic
- C++ stub files (`mettagrid_c.pyi`) remain unchanged
- The conversion function (`from_mettagrid_config`) works exactly as before

### Future considerations
In a future PR, we might consider consolidating `mettagrid_c_config.py` into `mettagrid_config.py`, but keeping them separate for now makes it easier to verify this PR contains only naming changes with no behavioral modifications.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210756693242079)


sasmith:
Please split this into commits that only do one thing.

Also, I'm going to give the layer invarience PR to Dave, since I think we should kill layers until they actually mean something, which they currently don't. I think he agrees, but we had some waffling.

mettagrid/src/metta/mettagrid/mettagrid_c_config.py
```diff
@@ -48,75 +55,68 @@ def from_mettagrid_config(mettagrid_config_dict: dict[str, Any]) -> GameConfig_c
...
             ),
-            "resource_rewards": dict(
-                (resource_ids[k], v) for k, v in merged_config.get("rewards", {}).items() if not k.endswith("_max")
-            ),
+            "resource_rewards": dict((resource_ids[k], v) for k, v in rewards_dict.items() if not k.endswith("_max")),
             "resource_reward_max": dict(
-                (resource_ids[k[:-4]], v)
-                for k, v in merged_config.get("rewards", {}).items()
-                if k.endswith("_max") and v is not None
+                (resource_ids[k[:-4]], v) for k, v in rewards_dict.items() if k.endswith("_max") and v is not None
```
rjwalters, line 60: I think this is the only substantive change proposed

mettagrid/src/metta/mettagrid/mettagrid_c_config.py
```diff
@@ -1,126 +1,122 @@
...
     # Group information is more specific than the defaults, so it should override
-    for group_name, group_config in mettagrid_config.groups.items():
+    for group_name, group_config in game_config.groups.items():
+        agent_group_props = copy.deepcopy(default_agent_config_dict)
+
         group_config_dict = group_config.model_dump(by_alias=True, exclude_unset=True)
-        merged_config = copy.deepcopy(agent_default_config_dict)
-        # update, but in a nested way
-        for key, value in group_config_dict.get("props", {}).items():
-            if isinstance(value, dict):
```
sasmith, line 35: I don't know why I split these out to begin with.

sasmith (approved):

PR #1389, opened 2025-07-09
mettascope webgl backend

monofuel:
- add a new `ContextWebgl` for using webgl instead of webgpu
- backends can be switched by commenting lines on common.ts
- attached ctx to the window for easier debugging

- also tested that webgl works in headless chrome for testing (webgpu did not)
smoketest


monofuel:
re-opening from a new branch not on a fork https://github.com/Metta-AI/metta/pull/1391

PR #1394, opened 2025-07-09
Fixing install for both OpenHands and Cursor Bckg Agents

relh:
Trying to more robustly find UV for cursor background agents

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210757391851892)


install.sh
```diff
@@ -52,28 +52,56 @@ check_cmd() {
...
 
+# More robust uv detection that checks direct paths
+find_uv() {
+    # First try command -v
+    if check_cmd uv; then
+        return 0
+    fi
+
+    # Check common installation paths directly
+    for path in "$HOME/.local/bin/uv" "$HOME/.cargo/bin/uv" "/opt/homebrew/bin/uv" "/usr/local/bin/uv"; do
```
nishu-builder, line 63: could put these things in a list since we're restating them a few times

nishu-builder (approved):
seems a little ugly but also not dangerous; in a perfect world we'd have a `POSSIBLE_UV_LOCS` or something that we use for both `find_uv` and the section that is possibly-adding uv, and the latter only adds the things not already in your path, and then before we do the funal "$UV" items we add in robb's suggestion: `export UV_PROJECT_ENVIRONMENT=${REPO_ROOT}/.venv`

PR #1417, opened 2025-07-10
Add more tests for specific curriculum scenarios

jacklionheart:


mettagrid/tests/test_curriculum.py
```diff
@@ -1,13 +1,37 @@
...
+Progressive Curriculum Tests:
+1. Monotonic linear signal -> should advance through tasks correctly
+2. Always 0 signal -> should stay on first task
+3. Random signal -> should still progress with right parameters
+
+Learning Progress Tests:
+4. Mixed impossible/learnable tasks -> should weight learnable evenly, ignore impossible
+5. Threshold dependency -> should first weight primary, then secondary after milestone
+
+Low Reward Curriculum Tests:
```
mstormbull, line 16: Should be prioritized regression. Is the impossibility outcome still true?

mettagrid/tests/test_curriculum.py
```diff
@@ -165,3 +197,1055 @@ def complete_task(self, id, score):
...
+    """
+
+    def get_score(self, task_id: str) -> float:
+        return 0.0
+
+
+class RandomScores(ScoreGenerator):
+    """Returns random scores in [min_val, max_val].
+
+    Useful for testing curriculum behavior under noisy conditions.
```
mstormbull, line 260: When using the noisy sampling is there a 'repeatability' condition that is worth enforcing?

mstormbull (approved):
Great tests. On the name change, more accurate name sounds good. In general maximums suffer from sometimes non intuitive extreme value statistics and so I wonder if there is another way to stabilize this approach in the face of noisy signal. Overall lgtm.

mettagrid/tests/test_curriculum.py
jacklionheart: Good catch, only changed the test not the top comment. An argument against maintaining the top comment but Ill leave it for now.

PR #1429, opened 2025-07-11
refactor: _compute_observation

rjwalters:
This PR refactors the observation computation logic:

1. Full support with unit tests for rectangular observation views in MettaGrid
```py
    obs_width: Literal[3, 5, 7, 9, 11, 13, 15]
    obs_height: Literal[3, 5, 7, 9, 11, 13, 15]
```
2. Enforced observation shape constraints in our trainer via pydantic override
```py
class PyPolicyGameConfig(PyGameConfig):
    obs_width: Literal[11]
    obs_height: Literal[11]
```
3. Manhattan distance iterator broken out to allow tests in `mettagrid/tests/test_observations.cpp`

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210769366241380)


sasmith:
Have you perf tested this? At a glance, it looks like you're adding complexity to add caching. I've tried some caching that seemed "obviously good" around observations, and it's made things slower.

Moreover, when I glance at this, it looks like "this is going to take work to understand", and that makes me nervous, and I really want to make sure I understand why we're proposing this. It does not look like an improvement in clarity, despite what the LLM summary says. Maybe I'll change my mind when I dig in more?

sasmith:
Handing off to @daveey. I don't have a strong reason to believe we need to support rectangular observation fields, but it's not a question I want to have to consider at the moment. Similarly, I think the logic you're replacing was hardly beautiful logic, but it's not logic that I think need to be supplanted at the moment.

Separately, I don't think the PR description is doing as much to help as it could. My assumption is that it's AI generated, and this makes it harder for me to trust claims that it makes about the impetus for the PR. I'd prefer shorter but more information dense descriptions.

rjwalters:
image

maybe slightly better perf? I think the main goal for this PR is code clarity. This gets rid of:

```c
  if (r_dist > distance) continue;
  int c_dist = distance - r_dist;
  // This is a bit ugly. We want to run over {c_dist, -c_dist}, but only do it once if c_dist == 0.
  // Here's how we're trying to do that, and to be performant (e.g., not re-allocating a set).
  for (int i = 0; i < 2; i++) {
    if (c_dist == 0 && i == 1) continue;
    int c_offset = i == 0 ? c_dist : -c_dist;
    int c = static_cast<int>(observer_col) + c_offset;
    // c could still be outside of our bounds.
    if (c < c_start || c >= c_end) continue;
 ```

File: mettagrid/src/metta/mettagrid/mettagrid_c.cpp [hunk 1]
```diff
@@ -260,65 +257,71 @@ void MettaGrid::_compute_observation(GridCoord observer_row,
...
-          int obs_c = object_loc.c + obs_width_radius - observer_col;
-
-          uint8_t location = PackedCoordinate::pack(obs_r, obs_c);
-
-          attempted_tokens_written += _obs_encoder->encode_tokens(obj, agent_obs_tokens, location);
-          tokens_written = std::min(attempted_tokens_written, static_cast<size_t>(observation_view.shape(1)));
-        }
-      }
+  // Helper lambda to process a single grid location - force inline for performance
+  const auto process_location = [&](int r, int c) __attribute__((always_inline)) -> void {
```
daveey, line 287: i'm not sure this approach is actually cleaner. @sasmith  do you have an opinion?

mettagrid/src/metta/mettagrid/packed_coordinate.hpp
```diff
@@ -68,5 +68,120 @@ inline bool is_empty(uint8_t packed_data) {
...
+/**
+ * Lightweight view into the observation pattern for a specific square window size.
+ *
+ * This class provides zero-allocation iteration over observation offsets for square
+ * windows. The allowed window sizes are 1x1, 3x3, 5x5, 7x7, 9x9, 11x11, 13x13, and 15x15.
+ *
+ * The offsets are pre-sorted so that each window size corresponds to a contiguous
+ * prefix of the global pattern, enabling simple pointer-based iteration.
+ */
+class ObservationSearchPattern {
```
daveey, line 150: is this approach faster? it's not simpler.

daveey (changes requested):

agent/src/metta/agent/metta_agent.py
```diff
@@ -35,7 +35,7 @@ def make_policy(env: "MettaGridEnv", cfg: DictConfig) -> "MettaAgent":
     return MettaAgent(
         obs_space=obs_space,
         obs_width=env.obs_width,
-        obs_height=env.obs_height,
+        obs_height=env.obs_width,
```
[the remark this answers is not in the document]
rjwalters: why would we want or need a rectangular view window? we have only used square ones so far

mettagrid/src/metta/mettagrid/packed_coordinate.hpp
rjwalters: I think it is significantly simpler and less error prone to enumerate all the squares, sort them by distance, and then iterate over that list vs trying to be clever with nested for loops. There is a modest performance increase in this approach

mettagrid/src/metta/mettagrid/mettagrid_c.cpp
rjwalters: the process location lambda could easily be inlined and I think this is a style preference -- I think it is easeier to understand the loop after breaking this out into the helper just to reduce things to essentials:

```c
 // Process locations in order using the search pattern
  for (const auto& offset : search_pattern) {
    int r = static_cast<int>(observer_row) + offset.r_offset;
    int c = static_cast<int>(observer_col) + offset.c_offset;

    // Skip if outside grid bounds
    if (r < r_start || r >= r_end || c < c_start || c >= c_end) {
      continue;
    }

    process_location(r, c);
  }
```

if you prefer the nested loops, we can keep the lambda or vice versa (drop the lambda, keep the simpler for loop)

daveey (approved):

PR #1435, opened 2025-07-11
feat: simplify `change_color` action

rjwalters:
## Core Idea

Agents can change their color to communicate with each other. We use HSV color space to preserve semantic distance as vocabulary grows.

## Why HSV Instead of Simple Numbering?

**With simple numbering (1,2,3,4...):**
- Adding new "words" breaks existing meanings
- No relationship between similar concepts

**With HSV (color wheel):**
- New colors fit between existing ones
- Similar colors = similar meanings
- Agents with different vocabulary sizes can still partially understand each other

## Example: Growing from 4 to 8 Colors

**4-color vocabulary:**
```
0: Red (0°)
1: Green (90°)
2: Cyan (180°)
3: Purple (270°)
```

**8-color vocabulary:**
```
0: Red (0°)
1: Orange (45°)
2: Green (90°)
3: Teal (135°)
4: Cyan (180°)
5: Blue (225°)
6: Purple (270°)
7: Magenta (315°)
```

The original 4 colors stay roughly in place - agents trained with 4 colors can still understand the "gist" of 8-color communication.

## Implementation

```cpp
// Map action argument to evenly distributed values [0, 255)
// For n colors: 0, 255/n, 2*255/n, ..., (n-1)*255/n
actor->color = (arg * 255) / max(max_arg() + 1, 1);
```

Priority set to 2 (higher than attacks) ensures color changes happen reliably for communication.

## Research Potential

- Will similar situations get similar colors?
- Can agents with different vocabulary sizes communicate?
- Do color "dialects" emerge in separate populations?

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210770402768736)


mettagrid/src/metta/mettagrid/actions/change_color.hpp
```diff
@@ -9,29 +9,19 @@
 
 class ChangeColorAction : public ActionHandler {
 public:
-  explicit ChangeColorAction(const ActionConfig& cfg) : ActionHandler(cfg, "change_color") {}
+  explicit ChangeColorAction(const ActionConfig& cfg) : ActionHandler(cfg, "change_color") {
+    priority = 2;  // higher than attack
```
daveey, line 13: i don't think we want this. the only reason it would matter is if the agent was attacked and frozen on the same tick it changed color. otherwise priority doesnt do anything

daveey (approved):
remove the priority change then feel free to merge

mettagrid/src/metta/mettagrid/actions/change_color.hpp
rjwalters: that was the intent - maybe in step 1, a hostile agent steps into my field of view, in step 2 I prioritize alerting my companions via color change anticipating an attack -- but it color change has lower priority than attack then I will be frozen before I change color...  my companions could interpret me being frozen as an indicator of the advancing hostile agent, but in a slightly modified scenario maybe I would want to first communicate that multiple agents appeared or that there are valuable resources over there... 

Anyway, the priority does not feel as critical -- do you want  me to pull it or leave it at 2?

rjwalters:
After some conversation with @bluewin4 and @treeform, some notes for the future:

1. "Color" has a lot of emotional loading and may lead to researchers introducing subtle bias if we use it as our semantic model for communication. Better to switch to signalling with glyphs or possibly oriented arrows (semaphore style)

2. Whether our instantaneously changed persistent communication channel is interpreted as a continuum or discrete signal is up to the policy - if the communication signal is not produced by anything other than policies then it may be less important to try to impose structure

In any case I think there is agreement for changing the communication channel to be instantaneous rather than integrating action, so I will merge this and we can change it again later as needed.

PR #1434, opened 2025-07-11
Integrate replay generation with policy evaluation

daveey:
## Summary

This PR integrates replay generation directly into the policy evaluation process, creating a unified system for generating and displaying replays for both training and evaluation environments.

### 1. **Unified Replay Generation**
- All replays (including training) now go through the `evaluate_policy` function
- Training environment is treated as `"eval/training_task"` in the simulation suite
- Eliminates code duplication between training and evaluation replay generation

### 2. **Enhanced Data Structures**
- Added `replay_urls: dict[str, str]` to `EvalResults` to store multiple replay URLs
- Added `replay_urls` field to `SimulationResults` for passing URLs between components
- `SimulationSuite` now collects replay URLs from each simulation

### 3. **Improved WandB Integration**
- Creates a single HTML view (`replays/all_links`) with all replay links
- Organizes replays into "Training Environment" and "Evaluation Environments" sections
- Maintains backward compatibility with legacy `replays/link` for training replay

## Benefits

- **Automatic replay generation**: Replays are generated during evaluation without separate intervals
- **Code simplification**: Single code path for all replay generation
- **Better organization**: All replays accessible from one consolidated HTML view
- **Improved maintainability**: Less duplicate code, easier to debug and extend

## Implementation Details

The key changes are:
1. `evaluate_policy` now accepts and passes through `replay_dir` parameter
2. `SimulationSuite.simulate()` collects replay URLs from individual simulations
3. `_generate_and_upload_replay` uses `evaluate_policy` instead of creating its own simulation
4. HTML generation separates training and evaluation replays for clarity

## Test Plan

- [ ] Verify training continues to work normally
- [ ] Check that evaluation replays are generated correctly
- [ ] Confirm all replay links appear in WandB HTML view
- [ ] Test backward compatibility with existing replay consumers

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1437, opened 2025-07-11
metta cli wraps tools

nishu-builder:
Reading all the `uv run ./tools/{x}.py {args}` commands [here](https://github.com/Metta-AI/metta/pull/1434/files#diff-e0d9c3f046dccba6f00e7cdeb50bfc30603c1a808a5b5ba27fd623a330d3bab9R24) inspired this

`metta tool {x} {args}` seems a bit cleaner, and moves us towards having a single entrypoint for developer commands in the repo

this change doesn't do much substantively. in the future, maybe `metta_cli` could:
- be the home for managing user configs (what we currently have in `configs/user/{name}.yaml`) and could automatically pass those along to tool commands
- parse arguments (e.g. `wandb` enabled) to these commands, check them against what is installed and connected, and prompt installation/authentication where not

### DIFF
--- a/metta/setup/metta_cli.py
+++ b/metta/setup/metta_cli.py
@@ -204,6 +204,19 @@ class MettaCLI:
         except subprocess.CalledProcessError as e:
             sys.exit(e.returncode)
 
+    def cmd_tool(self, tool_name: str, args: list[str]) -> None:
+        """Run a tool from the tools/ directory using uv run."""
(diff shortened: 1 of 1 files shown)


rjwalters (approved):
I guess we have a lot of tools so this makes sense...

Can we add opentelemetry to the metta tool so we can see what people actually use? That might inform promotion of `metta train` or `metta eval` to top level versus `metta tool train`

PR #1438, opened 2025-07-11
Cache heatmap and metrics requests

Barraketh:
The metrics caching is temporary until I migrate the episode_agent_metrics table to use metric ids

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210772461328844)

### DIFF
--- a/app_backend/src/metta/app_backend/metta_repo.py
+++ b/app_backend/src/metta/app_backend/metta_repo.py
@@ -198,16 +198,10 @@ MIGRATIONS = [
             """CREATE INDEX IF NOT EXISTS idx_episode_agent_metrics_metric_eiid_value
                 ON episode_agent_metrics(metric, episode_internal_id)
                 INCLUDE (value)""",
-            """DROP INDEX idx_episode_agent_metrics_metric_episode_value""",
-            """ALTER TABLE episode_agent_metrics DROP COLUMN episode_id""",
-        ],
-    ),
-    SqlMigration(
-        version=11,
-        description="Add episode_internal_id index to episode_agent_metrics table",
-        sql_statements=[
             """CREATE INDEX IF NOT EXISTS idx_episode_agent_metrics_episode_internal_id
                 ON episode_agent_metrics(episode_internal_id)""",
+            """DROP INDEX idx_episode_agent_metrics_metric_episode_value""",
+            """ALTER TABLE episode_agent_metrics DROP COLUMN episode_id""",
         ],
     ),
 ]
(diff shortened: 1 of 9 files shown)


app_backend/src/metta/app_backend/metta_repo.py
```diff
@@ -198,6 +198,8 @@
             """CREATE INDEX IF NOT EXISTS idx_episode_agent_metrics_metric_eiid_value
                 ON episode_agent_metrics(metric, episode_internal_id)
                 INCLUDE (value)""",
+            """CREATE INDEX IF NOT EXISTS idx_episode_agent_metrics_episode_internal_id
```
nishu-builder, line 201: to make GET_ALL_METRICS_QUERY fast we could also add an index on (metric, episode_internal_id)

app_backend/src/metta/app_backend/routes/dashboard_routes.py
```diff
@@ -59,10 +62,102 @@ class TrainingRunTagsUpdate(BaseModel):
...
+
+@dataclass
+class CachedMetrics:
+    """Cached metrics data with metadata."""
+
+    last_episode_id: int
+    metrics: List[str]
+
+
+class MetricsCache:
```
nishu-builder, line 87: do we want some max size for eviction in case of memory issues?

app_backend/src/metta/app_backend/routes/heatmap_routes.py
```diff
@@ -1,412 +1,449 @@
...
+        self._store(suite, metric, entry)
+
+        # Apply selector and build heatmap
+        if selector == "best":
+            selected_evals = await select_best_per_run(evaluations, eval_names_set)
+        else:
+            selected_evals = select_latest_per_run(evaluations)
+        return build_heatmap(selected_evals, eval_names)
+
+    async def _update_cache(
```
nishu-builder, line 345: not a huge deal but maybe want thread safety here -- could grab a `threading.Lock` specific to the `suite, metric`?

app_backend/src/metta/app_backend/metta_repo.py
nishu-builder: ah youre right! nice

app_backend/src/metta/app_backend/routes/dashboard_routes.py
Barraketh: Nah - there are not that many distinct metrics, and i want all of them available

PR #1439, opened 2025-07-11
mettascope npm run format

monofuel:
- mettascope has some linting drift, ran `npm run format` to fix things

### DIFF
--- a/mettascope/biome.jsonc
+++ b/mettascope/biome.jsonc
@@ -1,13 +1,8 @@
 {
   "$schema": "https://biomejs.dev/schemas/1.9.4/schema.json",
-  "extends": [
-    "../biome.jsonc"
-  ],
+  "extends": ["../biome.jsonc"],
   // MettaScope uses all the workspace defaults
   "formatter": {
-    "ignore": [
-      "./index.css",
-      "./index.html"
-    ]
+    "ignore": ["./index.css", "./index.html"]
   }
 }
--- a/mettascope/src/common.ts
+++ b/mettascope/src/common.ts
@@ -6,8 +6,7 @@ import { HoverPanel } from './hoverpanels.js'
 
 // The 3D context, used for nearly everything.
 export const ctx = new Context3d(find('#global-canvas') as HTMLCanvasElement)
-
-; (window as any).ctx = ctx
+;(window as any).ctx = ctx
 
 // Constants
 export const MIN_ZOOM_LEVEL = 0.025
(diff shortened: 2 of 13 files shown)


mettascope/src/demomode.ts
```diff
@@ -79,12 +78,7 @@ export function doDemoMode() {
   if (shot == null || shot.startTime + shot.duration < epochTime()) {
     if (Math.random() < 0.1) {
       // Create a shot that just shows the overall view.
-      shot = new Shot(
-        ShotType.SHOW_OVERALL_VIEW,
-        0,
-        3,
-        0
-      )
+      shot = new Shot(ShotType.SHOW_OVERALL_VIEW, 0, 3, 0)
```
monofuel, line 81: I don't really like how the linter wants to smush everything on one line when it seemed more readable before. might need to change settings

mettascope/src/common.ts
```diff
@@ -135,14 +134,7 @@ export const html = {
...
 
-  speedButtons: [
-    find('#speed1'),
-    find('#speed2'),
-    find('#speed3'),
-    find('#speed4'),
-    find('#speed5'),
-    find('#speed6'),
-  ],
+  speedButtons: [find('#speed1'), find('#speed2'), find('#speed3'), find('#speed4'), find('#speed5'), find('#speed6')],
```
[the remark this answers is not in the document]
monofuel: I can't find any way to get biome to format this nicely without changing`lineWidth` from 120 to something like 90 or 100.

mettascope/biome.jsonc
```diff
@@ -1,13 +1,14 @@
...
-  ],
+  "extends": ["../biome.jsonc"],
   // MettaScope uses all the workspace defaults
   "formatter": {
-    "ignore": [
-      "./index.css",
-      "./index.html"
-    ]
+    "ignore": ["./index.css", "./index.html"],
+    "lineWidth": 110
```
monofuel, line 7: I couldn't find any good settings to handle multiline arrays or function arguments better. 110 seems like a good number for our current code.. 100 seemed too aggressive and 120 made lines too long.

treeform:
@nishu-builder eventually we will, but we had issues with the formatter breaking our code for no reason. We will ease into it. Not this PR though...

mettascope/biome.jsonc
monofuel: got feedback that we aren't picky  about formatting, will remove my tweaks.

PR #1443, opened 2025-07-11
[MettaScope] fix broken url param updating

monofuel:
- mettascope was acting pretty slow / broken because it was hammering url updates every frame.
- chrome was throttling this call, so mettascope still acted ok
- firefox really slows down when this api is hammered

### DIFF
--- a/mettascope/src/main.ts
+++ b/mettascope/src/main.ts
@@ -277,44 +277,84 @@ document.addEventListener('blur', function (e) {
   requestFrame()
 })
 
+/** Track previous URL parameter values to avoid unnecessary updates. */
+let previousUrlState = {
+  step: -1,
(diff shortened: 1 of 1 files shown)


mettascope/src/main.ts
```diff
@@ -277,44 +277,84 @@ document.addEventListener('blur', function (e) {
   requestFrame()
 })
 
+/** Track previous URL parameter values to avoid unnecessary updates. */
+let previousUrlState = {
```
monofuel, line 281: alternatively we could:
- update once every 60 frames (that doesn't feel great assuming fps)
- instead of tracking previous state as an object, we could read in the url params and check for diff? feels less great in case we have some sort of diff logic bug, then we end up hammering it again.

PR #1447, opened 2025-07-11
feat: add `change_glyph` action for communication

rjwalters:
This PR introduces a new `change_glyph` action that enables agents to communicate through visual symbols (glyphs). Unlike the existing `change_color` action which uses a continuous color space, glyphs provide a discrete communication channel with a configurable number of distinct symbols.

## Motivation

Multi-agent environments benefit from diverse communication mechanisms. While the `change_color` action provides a way for agents to signal using colors, it has limitations:
- Colors exist on a continuum, which can make discrete signaling ambiguous
- The color channel is already used for team/group identification in some scenarios

Glyphs offer a complementary communication method:
- Discrete, unambiguous symbols (e.g., 4, 8, or 16 distinct glyphs)
- Clear semantic separation from color-based team identification
- More intuitive for certain communication protocols (e.g., signaling specific strategies or states)

### Core Implementation
- **New action**: `ChangeGlyph` action handler that allows agents to set their glyph to one of N configurable values
- **Agent state**: Added `glyph` property to agents (defaults to 0)
- **Observations**: Added `Glyph` observation feature (type 12) so agents can observe each other's glyphs
- **Configuration**: Added `ChangeGlyphActionConfig` with configurable `number_of_glyphs` parameter

### Configuration Updates
- Added `change_glyph` action to all environment configs with sensible defaults (4 glyphs)
- Updated Python bindings and configuration classes to support the new action

### Other Improvements
- Renamed action files for clarity: `put_recipe_items.hpp` → `put_items.hpp`, `get_output.hpp` → `get_items.hpp`
- Updated `change_color` action to use increment/decrement/double/half operations instead of direct mapping

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210778321667358)


mettagrid/src/metta/mettagrid/actions/get_items.hpp
```diff
@@ -10,9 +10,9 @@
 #include "objects/converter.hpp"
 #include "types.hpp"
 
-class GetOutput : public ActionHandler {
+class GetItems : public ActionHandler {
 public:
-  explicit GetOutput(const ActionConfig& cfg) : ActionHandler(cfg, "get_output") {}
+  explicit GetItems(const ActionConfig& cfg) : ActionHandler(cfg, "get_items") {}
```
daveey, line 15: i think this will break mettascope, did you test it?

mettagrid/src/metta/mettagrid/actions/get_items.hpp
```diff
@@ -1,5 +1,5 @@
-#ifndef ACTIONS_GET_OUTPUT_HPP_
-#define ACTIONS_GET_OUTPUT_HPP_
+#ifndef ACTIONS_GET_ITEMS_HPP_
```
daveey, line 1: we named it this way for a reason, dont rename it please

mettagrid/src/metta/mettagrid/actions/put_items.hpp
```diff
@@ -1,5 +1,5 @@
-#ifndef ACTIONS_PUT_RECIPE_ITEMS_HPP_
-#define ACTIONS_PUT_RECIPE_ITEMS_HPP_
+#ifndef ACTIONS_PUT_ITEMS_HPP_
```
daveey, line 1: don't rename this

daveey (changes requested):

rjwalters:
@daveey - thanks for the initial feedback

1. just floating the shorter names as a suggestion; I'm happy to revert
2. I think it might better to reserve glyph zero as "no glyph" and omit the token?
3. I wish the glyph space was more of a policy property than an env property so that policies could have different vocabularies -- I think this would probably require us to rethink our action space so I think it is beyond scope here but maybe something for the future
4. I fixed metttscope but biome ended up formatting a bunch of files so it was hard to tell what was changing; I wanted to check with @treeform on that and won't merge this until after

mettagrid/src/metta/mettagrid/actions/put_items.hpp
rjwalters: the idea here was to match the python convention:

    put_items:
      enabled: false
    get_items:
      enabled: true
      
in order to unify our configuration interface

rjwalters: as per conversation, (1) we will eventually change the python code to match and (2) we are going to leave the mismatch for now until we settle on a stable plan for actions at some point in the future

daveey (approved):

PR #1457, opened 2025-07-14
v1 researcher notebook

nishu-builder:
Basic notebook with support for launching skypilot jobs, handling met…rics from wandb runs in dataframes, and viewing replays inline

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210784855454929)


Barraketh (commented):
So, this notebook really seems to be aimed at running a training run and then analyzing that training run.  I think based on what we talked about yesterday we need a notebook that's just focused on analysis of an existing run (or set of runs).  I have nothing against committing this one, but we do need the analysis one

.github/workflows/checks.yml
```diff
@@ -140,11 +140,11 @@ jobs:
 
       - name: Run Ruff linter
         run: |
-          ruff format --check .
+          uv run ruff format --check .
```
nishu-builder, line 143: @rjwalters just checking this is cool with you

PR #1460, opened 2025-07-14
Add systematic exploration memory configs and maps

georgejwdeane:
- systematic exploration eval set, minimal_memory_set


daphnedemekas (changes requested):
Please make the following changes:

you_shall_not_pass : too long, should be like 110
which_way - max 50 timesteps (are they supposed to get all three or just 2. if all 3, then what is the point of this eval?)
systematic_memory: not quite enough time. maybe 350
spacey_memory: too long, make it 200
lobster legs: max 210
little_landmark_hard: way too much time, make it 100
journey_home: like 110
hall_of_mirrors: 150
choose_wisely: 200
boxout: 150
access_cross: delete this one

daphnedemekas (approved):
Hey, this looks good. It would be nice if you could move num_agents: 1 into the defaults - no need to have that in every file. After that feel free to merge!

PR #1463, opened 2025-07-14
Always evaluate policy at training end

nishu-builder:
To address confusion about the lack of generated eval metrics or replays, even though there are saved policy checkpoints, for runs with very few timesteps

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210794225364893)


rjwalters (approved):
fwiw I don't think we reach this code very often - most jobs are terminated by ctrl-c or skypilot job cancel far before they hit 10B steps.. but this could be useful for people who set a max steps and go to completion

PR #1468, opened 2025-07-14
fix tagging

bluewin4:
Add write permissions to workflow so it can tag properly

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210795259322625)


nishu-builder (approved):
nice ty :)

PR #1473, opened 2025-07-14
[MettaScope] Altar should only show hearts

monofuel:
- https://app.asana.com/1/1209016784099267/project/1209513563278644/task/1210610634771469

- currently, Altars always show a heart if there is any contents, even if it is not a heart.

- if a grid object has it's output in the inventory, we should show the item overlay.
- otherwise, we should not show the overlay, even if there are other contents present.

heart and battery present:
Screenshot From 2025-07-16 15-35-19

only battery present:
Screenshot From 2025-07-16 15-36-09

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210795553258681)


mettascope/src/worldmap.ts
```diff
@@ -177,7 +177,27 @@ function drawObject(gridObject: any) {
...
+        if (hasHeart) {
+          // Default overlay already depicts a heart.
+          ctx.drawSprite(state.replay.object_images[type][1], x * Common.TILE_SIZE, y * Common.TILE_SIZE)
+        } else if (hasBattery) {
+          // Display the battery resource on top of the altar
+          let batteryKey = 'inv:battery'
+          let pos = new Vec2f(x * Common.TILE_SIZE, y * Common.TILE_SIZE)
+          // bump the position up a bit to look right
+          pos.setY(pos.y() - Common.TILE_SIZE * 0.4)
+          ctx.drawSprite(state.replay.resource_inventory.get(batteryKey)[0], pos.x(), pos.y(), [1, 1, 1, 1], 0.5, 0)
```
monofuel, line 198: This feels more gross than the other overlays we have? but this is currently the only one-off case where an object can have different contents.

daveey: I agree. I don't love having a battery popping out of an altar. Can we just show buildings as empty if they don't have ready outputs? 

The next step would be to show a "partially-ready" indicator. Could be a pie that is missing segments. So a building that needs 3 batteries but only has 2, would show a pie that's 2/3 filled. If it needs 3 batteries and an ore, and has 5 batteries, it would show 3/4 filled. etc....

daveey (changes requested):

mettascope/src/worldmap.ts
monofuel: Whoops, Treeform only wanted it to show heart when it contains a heart, and not show input items, I misunderstood the ticket

mettascope/src/replay.ts
```diff
@@ -339,7 +339,9 @@ export function loadReplayStep(replayStep: any) {
 
 /** Get object config. */
 export function getObjectConfig(object: any) {
-  let typeName = state.replay.object_types[object.type]
+  let typeId = getAttr(object, 'type')
```
monofuel, line 342: Gv_L0JnakAAXFoy

monofuel: (jk jk)

PR #1479, opened 2025-07-15
Change episode truncation to termination in MettaGrid

daveey:
### TL;DR

Changed episode ending behavior to use termination instead of truncation when max steps are reached.

### What changed?

Modified the `_step` method in `mettagrid_c.cpp` to set the `_terminals` flag instead of `_truncations` when the maximum number of steps is reached. This changes how episode endings are signaled to the agent.

### How to test?

1. Run an environment with a set `max_steps` value
2. Verify that episodes end with termination signals rather than truncation signals when reaching the maximum steps
3. Check that agent behavior is appropriate for terminated episodes

### Why make this change?

The code comments explain the rationale: setting termination instead of truncation is considered "more realistic" for the agent experience. This change makes the agent recognize that episodes genuinely end rather than being artificially cut short, which better aligns with the intended simulation behavior.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210798388172596)

### DIFF
--- a/configs/env/mettagrid/curriculum/bbc/bbc.yaml
+++ b/configs/env/mettagrid/curriculum/bbc/bbc.yaml
@@ -27,3 +27,4 @@ tasks:
   /env/mettagrid/curriculum/bbc/agent_1.armor: 1
   /env/mettagrid/curriculum/bbc/agent_1.combat: 1
   /env/mettagrid/curriculum/bbc/agent_1.tag: 1
+  /env/mettagrid/curriculum/arena/learning_progress: 1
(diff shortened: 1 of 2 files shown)


sasmith:
I have a PR to make this configurable. How about I ship that with a default of "use terminal"?

My mental model was that we should use terminal for episodes with fixed reward, and use truncation for episodes with variable reward (like arena). But I don't know how much precision matters, and agree that terminal is probably a better default if we had to pick one.

sasmith:
Or better yet, let's merge this and my change and resolve conflicts if / when they matter.

PR #1482, opened 2025-07-15
bug: document hardcoded reward metric breaking sweep flexibility

akrbc9:
bug: document hardcoded reward metric breaking sweep flexibility

The trainer hardcodes 'reward' as the evaluation metric instead of
respecting cfg.sweep.metric, preventing optimization of other metrics
like episode_length or success_rate. Includes proposed fixes.

bug: add missing metadata score bug to evaluation report

Discovered that trainer never writes metadata['score'] at all, making
policy selection essentially random. This is CRITICAL as it breaks
all sweep checkpoint selection, not just non-reward optimization.

fix: save aggregated score to metadata for policy selection during sweep evaluation.

fix: implement validation logic with retries for batch size divisibility in protein suggestions

- Fixed backwards divisibility check: batch_size must be divisible by minibatch_size
- Added comprehensive unit tests for validation pipeline
- Tests cover all edge cases and validation scenarios

fix: pass sweep_id instead of wandb_path for run id generation

fix: resolve multiple issues in sweep initialization

- Fixed protein instance being created twice (once in init_run, once in generate_protein_suggestion)
- Made protein parameter required in generate_protein_suggestion to avoid duplication
- Fixed validation to uncomment and use correct config paths (trainer.ppo.*)
- Fixed sweep_rollout.sh to use correct path without ./ prefix
- Updated tests to match new function signatures

fix: use correct dict structure for suggestion parsing during validation + fix: 0 cost for failure to 0.001:
- add batch_size, minibatch_size, bppt_horizon back to sweeps
- fix SimulationSuite usage in sweep_eval.py
- update configs


tools/sweep_init.py
```diff
@@ -84,6 +85,7 @@ def create_run(cfg: DictConfig | ListConfig, logger: Logger) -> str:
     sweep_metadata = OmegaConf.load(os.path.join(cfg.sweep_dir, "config.yaml"))
 
     # Generate a new run ID for the sweep, e.g. "simple_sweep.r.0"
+    # TODO: Use sweep_id instead of sweep_path, currently very confusing.
```
akrbc9, line 88: NB: The use as-is is correct.

PR #1491, opened 2025-07-16
WandB summary now tracks the policy's parameters

Al-does:
I just tracked all params, didn't feel it was necessary if the param required grad or not as most or all will.


rjwalters (approved):
are there other model parameters we want to log?

PR #1494, opened 2025-07-16
Refactor proposal for `agents` subpackage

UmerHA:
Hi all, as discussed with Robb, here’s a PR to propose a refactor to simplify the agent subpackage.


relh:
Simplifying this sounds great! I like this proposal.

PR #1499, opened 2025-07-16
Upload fully-specified env configs per bucket to wandb

nishu-builder:
Jack wants an easy way to be able to give a run id, an epoch num, and get the set of tasks that were at play (and their probabilities, which can be inferred through completion rates already in wandb)

Right now, we upload bucket names, but not the full bucket specs to wandb

This means that in order to find out what the exact specs of a bucket was in a training run you have to also switch to the branch on which that run was run

This uploads the full env cfg for each bucket in the trainer's curriciulum. Can later be fetched using `metta.mettagrid.util.file`'s `read`

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210815074826584)

### DIFF
--- a/mettagrid/src/metta/mettagrid/curriculum/sampling.py
+++ b/mettagrid/src/metta/mettagrid/curriculum/sampling.py
@@ -17,15 +17,21 @@ class SamplingCurriculum(Curriculum):
     def __init__(self, env_cfg_template: str, env_overrides: Optional[DictConfig] = None):
         self._cfg_template = config_from_path(env_cfg_template, env_overrides)
 
+    @property
+    def _name(self) -> str:
+        return f"sample({self._cfg_template.sampling})"
+
     def get_task(self) -> Task:
         cfg = OmegaConf.create(copy.deepcopy(self._cfg_template))
         OmegaConf.resolve(cfg)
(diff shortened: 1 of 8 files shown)


bluewin4:
Thank you!

mettagrid/src/metta/mettagrid/curriculum/sampling.py
```diff
@@ -17,15 +17,21 @@ class SamplingCurriculum(Curriculum):
     def __init__(self, env_cfg_template: str, env_overrides: Optional[DictConfig] = None):
         self._cfg_template = config_from_path(env_cfg_template, env_overrides)
 
+    @property
+    def _name(self) -> str:
```
rjwalters, line 21: we could probably make the names a lot nicer in general - I am not super happy with the "short name" system... we don't have that many curricula defined so we could just make it abstract and require children to implement

rjwalters (approved):
lgtm

daveey (changes requested):
This approach doesn't make sense:

1. buckets are just a concept of bucketted curriculum
2. we are soon going to generate tasks via mutation (or other generative algorithms), so the space of configs will become huge. 

I think we should revert this and discuss a better approach

PR #1503, opened 2025-07-16
[MettaScope] hide actions for replays

monofuel:
https://app.asana.com/1/1209016784099267/project/1209513563278644/task/1210538273694971?focus=true

- Replays do not support actions. showing the memory interface is confusing since it doesn't do anything.
- if we loaded a replay, keep the actions hidden.

Screenshot From 2025-07-16 16-49-55

### DIFF
--- a/mettascope/src/hoverpanels.ts
+++ b/mettascope/src/hoverpanels.ts
@@ -68,7 +68,12 @@ hoverPanel.addEventListener('mousedown', (e: MouseEvent) => {
   // and if the websocket is connected.
   let actions = findIn(panel.div, '.actions')
   if (state.ws != null && panel.object.hasOwnProperty('agent_id')) {
-    actions.classList.remove('hidden')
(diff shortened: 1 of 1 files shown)


monofuel:
I wonder if we should also hide `id="controls-toggle"` and the controls panel for replays?

mettascope/src/hoverpanels.ts
```diff
@@ -68,7 +68,12 @@ hoverPanel.addEventListener('mousedown', (e: MouseEvent) => {
   // and if the websocket is connected.
   let actions = findIn(panel.div, '.actions')
   if (state.ws != null && panel.object.hasOwnProperty('agent_id')) {
-    actions.classList.remove('hidden')
+    if (state.replay != null) {
```
monofuel, line 71: Turns out this does not work- in play mode, replay is used to keep a historical log. need a better way to figure out play vs replay

PR #1507, opened 2025-07-16
add max_reward

sasmith:
Adds max_reward to terminate episodes when total reward threshold is reached

This PR adds a new `max_reward` parameter to the MettaGrid environment that terminates episodes when the total accumulated reward reaches this threshold. This is intended to represent scenarios where "all the reward on the map has been collected."

Key changes:
- Added `max_reward` field to GameConfig structure
- Implemented logic to check for episode termination when total reward exceeds max_reward
- Refactored navigation configs to use a centralized max_reward parameter instead of duplicating values
- Changed default number of agents from 4 to 1 in navigation configs

I'm currently testing a training run on this, both to test the max_reward and also to test the simpler 4 -> 1 agent switch.

This would be better with a unit test, but I'm submitting now since that pushes the code remotely, so I can start my training jobs.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210816370675249)

### DIFF
--- a/configs/env/mettagrid/navigation/training/defaults.yaml
+++ b/configs/env/mettagrid/navigation/training/defaults.yaml
@@ -4,7 +4,7 @@ defaults:
 
 sampling: 1
 game:
-  num_agents: 4
+  num_agents: 1
   max_steps: 1000
   actions:
     attack:
--- a/configs/env/mettagrid/navigation/training/defaults.yaml
+++ b/configs/env/mettagrid/navigation/training/defaults.yaml
@@ -5,6 +5,7 @@ defaults:
 sampling: 1
 game:
   num_agents: 1
+  max_reward: ${sampling:3, 50, 25}
   max_steps: 1000
   actions:
     attack:
@@ -24,7 +25,7 @@ game:
       border_width: 3
       agents: 1
       objects:
-        altar: ${sampling:3, 50, 25}
+        altar: ${game.max_reward}
   objects:
     altar:
       cooldown: 1000
--- a/mettagrid/src/metta/mettagrid/mettagrid_c.cpp
+++ b/mettagrid/src/metta/mettagrid/mettagrid_c.cpp
@@ -34,10 +34,11 @@
 namespace py = pybind11;
 
 MettaGrid::MettaGrid(const GameConfig& cfg, const py::list map, unsigned int seed)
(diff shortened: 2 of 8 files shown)


sasmith:
* **#1507** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1507?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1507?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

configs/env/mettagrid/navigation/training/defaults.yaml
```diff
@@ -4,7 +4,8 @@ defaults:
 
 sampling: 1
 game:
-  num_agents: 4
+  num_agents: 1
+  max_reward: ${sampling:3, 50, 25}
```
daveey, line 8: let's not sample in configs, that should now be up to curriculum

mettagrid/src/metta/mettagrid/mettagrid_c.cpp
```diff
@@ -417,7 +418,21 @@ void MettaGrid::_step(py::array_t<ActionType, py::array::c_style> actions) {
     episode_rewards_view(i) += rewards_view(i);
   }
 
-  // Check for truncation
+  // Check for episode end due to reward limit
+  if (max_reward > 0) {
```
daveey, line 422: max_reward doesn't communicate that the episode will end when reached, how about episode_termination_reward_threshold, or something?

daveey, line 422: max_reward doesn't communicate that the episode will end when reached, how about episode_termination_reward_threshold, or something?

configs/env/mettagrid/navigation/training/defaults.yaml
sasmith: Yes, but later.

mettagrid/src/metta/mettagrid/mettagrid_c.cpp
sasmith: Sure. I was mirroring `max_steps`, but am happy to be more verbose.

PR #1508, opened 2025-07-16
Remove and simplify tests

sasmith:
Simplifies reward testing by removing unnecessary test environment and tests.

The main change here is removing tests about reward accumulation. In these tests, the agent never gets any reward, so we're just ensuring that adding zero a bunch results in zero.

I considered updating these tests to use non-zero values, but `test_multiple_heart_collections` is testing accumulation anyways, so I instead deleted the test.

This also removes a few extra config params we don't need.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210816481162760)


rjwalters (approved):
we should turn on codecov soon

PR #1520, opened 2025-07-17
cp

daveey:
cp

cp

cp

cp

cp

cp


mettagrid/src/metta/mettagrid/mettagrid_c.cpp
```diff
@@ -916,6 +925,16 @@ PYBIND11_MODULE(mettagrid_c, m) {
            py::arg("number_of_glyphs"))
       .def_readonly("number_of_glyphs", &ChangeGlyphActionConfig::number_of_glyphs);
 
+  py::class_<GlobalObsConfig>(m, "GlobalObsConfig")
+      .def(py::init<>())
+      .def(py::init<bool, bool, bool>(),
+           py::arg("episode_completion_pct") = true,
```
sasmith, line 931: I'm undecided on whether defaults make sense here. Depends on how we expect these to be constructed.

So for now, sounds good.

mettagrid/src/metta/mettagrid/mettagrid_c_config.py
```diff
@@ -116,5 +126,7 @@ def from_mettagrid_config(mettagrid_config_dict: dict[str, Any]) -> CppGameConfi
 
     game_cpp_params["actions"] = actions_cpp_params
     game_cpp_params["objects"] = objects_cpp_params
+    # Note: global_observations configuration is handled internally by the C++ code
```
sasmith, line 129: I don't understand the comment.

sasmith (approved):

mettagrid/src/metta/mettagrid/mettagrid_c_config.py
daveey: leftover from previous diff, will remove

PR #1524, opened 2025-07-17
Add inventory rewards observation to MettaGrid

daveey:
# Add Inventory Rewards Global Observation to MettaGrid

This PR adds a new global observation feature that provides agents with information about which inventory items provide rewards. This allows agents to adapt their behavior based on which resources are valuable in the current game.

Key changes:
- Added `inventory_rewards` flag to global observation config (disabled by default)
- Implemented a packed representation that encodes up to 8 inventory items into a single byte
- Each bit represents whether an item has a positive reward (1) or not (0)
- Added comprehensive documentation in `global_observations.md` explaining the feature
- Added unit tests to verify correct functionality with different reward configurations

The inventory rewards observation is useful for multi-task learning scenarios where agents need to adapt to different reward structures, curriculum learning with gradually changing rewards, and creating game variants with the same environment but different objectives.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210820943572733)


mettagrid/src/metta/mettagrid/mettagrid_c.cpp
```diff
@@ -292,6 +315,11 @@ void MettaGrid::_compute_observation(GridCoord observer_row,
     global_tokens.push_back({ObservationFeature::LastReward, reward_int});
   }
 
+  // Add inventory rewards for this agent
+  if (_global_obs_config.resource_rewards && !_resource_rewards.empty() && agent_idx < _resource_rewards.size()) {
```
sasmith, line 319: The last check seems fine, but it also seems like it should be an assert. I.e., `agent_idx` should always be less than `_resource_rewards`.

rjwalters (commented):
I feel like the packed coordinate code should be refactored to "observation_tokens.hpp" and that we should centralize our complex packing logic there. Tokens are expensive so I imagine we are going to want to do this again. We could also generalize to allow more than 8 by using more token types (i.e. "resource_rewards b0:7", "resource_rewards b8:15") etc

PR #1532, opened 2025-07-17
minor-feat: Add a check to PRs if they are external to avoid failing tests

relh:
Currently external PRs hit this: 

0s
Run chmod +x .github/scripts/upload_codecov.py
❌ Error: CODECOV_TOKEN is required (pass via env or argv)
Error: Process completed with exit code 1.

---

Get this when running CI on external PRs from 

https://github.com/Metta-AI/metta/pull/1340

This PR is to skip it if doesn't exist

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210825786577910)


rjwalters:
is this the same issue we had with the asana task? I get that we might want to approve running CI on external PRs but maybe we should find a general way to grant permissions. How do the smoke tests run without wandb api for example

rjwalters (changes requested):
I think this can be done in a simpler way... I'll send you a pr

PR #1535, opened 2025-07-17
`metta run [component]`, `metta configure [component]`, and updated pre-commit hooks

nishu-builder:
Support run commands for components

Support configure commands for components (gets stored in the metta config file in ~/.metta)

Refactored other parts of the cli

pre-commit hooks now run `metta run githooks precommit`, which refer to a setting for how we want to lint. If you've selected to auto-fix, we also auto-re-add your staged files, so you shouldn't need to run the command twice.

`metta lint --fix` now flips the order of `format` and `check` so we format first, meaning you don't have to run it twice if there are fixable errors

### DIFF
--- a/metta/setup/utils.py
+++ b/metta/setup/utils.py
@@ -1,8 +1,11 @@
 import importlib
 import textwrap
 from pathlib import Path
+from typing import TypeVar
 
 from metta.common.util.colorama import Fore, blue, bold, colorize, cyan, green, red, yellow
+
+T = TypeVar("T")
 
 
 def _format_message(message: str) -> str:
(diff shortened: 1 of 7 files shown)


metta/setup/utils.py
```diff
@@ -34,6 +37,46 @@ def step(message: str) -> None:
     print(colorize(_format_message(message), Fore.WHITE))
 
 
+def prompt_choice(prompt: str, choices: list[tuple[T, str]], default: T | None = None, current: T | None = None) -> T:
```
rjwalters, line 40: some overlap here with `common/src/metta/common/util/cli.py`

rjwalters (approved):
👍

metta/setup/utils.py
nishu-builder: hmm true i will merge in another pr

PR #1536, opened 2025-07-17
[MettaScope] Mobile UI

monofuel:
https://github.com/user-attachments/assets/f6699347-6e94-437d-a6bc-74aa53925975

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210827085366136)


mettascope/src/common.ts
```diff
@@ -60,6 +60,15 @@ export const ui = {
   mainScrubberDown: false,
   mainTraceMinimapDown: false,
 
+  // Touch events for pinch-to-zoom
```
monofuel, line 63: Turns out multitouch is hard!

to zoom, we have to keep track of multiple touch points. it also feels intuitive to be able to pan while zooming.

mettascope/src/hoverpanels.ts
```diff
@@ -52,7 +52,7 @@ document.body.appendChild(hoverPanel)
 findIn(hoverPanel, '.actions').classList.add('hidden')
 hoverPanel.classList.add('hidden')
 
-hoverPanel.addEventListener('mousedown', (e: MouseEvent) => {
+hoverPanel.addEventListener('pointerdown', (e: PointerEvent) => {
```
monofuel, line 55: Pointer events work for a broad range of input methods, not just mouse.

mettascope/src/common.ts
monofuel: Safari has a special gesture API, but nothing else supports it afaik.

we only care about pinch to zoom so it's probably fine to just implement ourselves.

mettascope/src/panels.ts
```diff
@@ -74,13 +74,62 @@ export class PanelInfo {
...
+
     if (ui.mouseClick) {
       this.isPanning = true
     }
     if (!ui.mouseDown) {
       this.isPanning = false
     }
 
+    // Handle pinch-to-zoom gestures
+    if (ui.isPinching && ui.touches.length === 2) {
```
monofuel, line 96: This is the actual fun pinch to zoom + pan code

mettascope/style.css
```diff
@@ -155,3 +155,155 @@ html {
 input::placeholder {
   opacity: 0.25;
 }
+
+/* moderately sized tablet or phone screens */
```
monofuel, line 159: These CSS tweaks are enough to make the entire UI work, but there's more room for improvement.

I tested this 'moderately sized screen' css with my tablet in portrait mode.

mettascope/style.css
```diff
@@ -155,3 +155,155 @@ html {
...
+  #header input#search-input {
+    width: 150px !important;
+  }
+
+  #header #search-icon {
+    right: 215px !important;
+  }
+}
+
+/* very small phone sized screens */
```
monofuel, line 267: I tested this on my phone in portrait mode. room for improvement, but overall looks reasonable? all the inputs work.
Screenshot_20250717_173408_Firefox

treeform (approved):

PR #1537, opened 2025-07-17
Optional limit on Converter operations

lsoros:
- **Adding support for limiting the number of conversion operations that can be done by a converter. max_conversions can now be specified in config, and the converter will be set to permanent stop once that limit has been reached.**
- **Added optional limit for Converter unit operations**

In the object config files, there is now a field for specifying a max number of conversion operations that can be performed. Default value of -1 means unlimited conversions. 

Tested with values -1, 0, 1, and 2 on red mines.

### DIFF
--- a/configs/env/mettagrid/game/objects/advanced.yaml
+++ b/configs/env/mettagrid/game/objects/advanced.yaml
@@ -6,6 +6,7 @@ lab:
   output_resources:
     blueprint: 1
   max_output: 5
+  max_conversions: -1
   conversion_ticks: 1
   cooldown: 5
   initial_resource_count: 0
@@ -20,6 +21,7 @@ factory:
     armor: 5
     laser: 5
   max_output: 5
+  max_conversions: -1
   conversion_ticks: 1
   cooldown: 5
   initial_resource_count: 0
@@ -32,6 +34,7 @@ temple:
   output_resources:
     heart: 5
   max_output: 5
+  max_conversions: -1
   conversion_ticks: 1
   cooldown: 5
   initial_resource_count: 0
--- a/configs/env/mettagrid/game/objects/mines.yaml
+++ b/configs/env/mettagrid/game/objects/mines.yaml
@@ -4,8 +4,9 @@ mine_red:
     ore_red: 1
   color: 0
   max_output: 5
+  max_conversions: -1
   conversion_ticks: 1
-  cooldown: 50
+  cooldown: 2
   initial_resource_count: 0
 
 mine_blue:
@@ -14,8 +15,9 @@ mine_blue:
   output_resources:
     ore_blue: 1
   max_output: 5
+  max_conversions: -1
   conversion_ticks: 1
-  cooldown: 50
+  cooldown: 2
   initial_resource_count: 0
 
 mine_green:
@@ -24,6 +26,7 @@ mine_green:
     ore_green: 1
   color: 2
   max_output: 5
+  max_conversions: -1
   conversion_ticks: 1
-  cooldown: 50
+  cooldown: 2
   initial_resource_count: 0
--- a/mettagrid/src/metta/mettagrid/mettagrid_config.py
+++ b/mettagrid/src/metta/mettagrid/mettagrid_config.py
@@ -106,6 +106,7 @@ class PyConverterConfig(BaseModelWithForbidExtra):
     output_resources: dict[str, int] = Field(default_factory=dict)
     type_id: int = Field(default=0, ge=0, le=255)
     max_output: int = Field(ge=-1)
+    max_conversions: int = Field(ge=-1)
     conversion_ticks: int = Field(ge=0)
     cooldown: int = Field(ge=0)
     initial_resource_count: int = Field(ge=0)
--- a/mettagrid/src/metta/mettagrid/objects/converter.hpp
+++ b/mettagrid/src/metta/mettagrid/objects/converter.hpp
@@ -19,6 +19,7 @@ struct ConverterConfig : public GridObjectConfig {
                   const std::map<InventoryItem, InventoryQuantity>& input_resources,
                   const std::map<InventoryItem, InventoryQuantity>& output_resources,
                   short max_output,
+                  short max_conversions,
                   unsigned short conversion_ticks,
                   unsigned short cooldown,
                   InventoryQuantity initial_resource_count,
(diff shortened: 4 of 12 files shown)


mettagrid/src/metta/mettagrid/mettagrid_config.py
```diff
@@ -106,6 +106,7 @@ class PyConverterConfig(BaseModelWithForbidExtra):
     output_resources: dict[str, int] = Field(default_factory=dict)
     type_id: int = Field(default=0, ge=0, le=255)
     max_output: int = Field(ge=-1)
+    max_conversions: int = Field(ge=-1)
```
sasmith, line 109: Let's give this a default of -1. This is a perfectly good place for a default, and -1 is a perfectly good default.

mettagrid/src/metta/mettagrid/objects/converter.hpp
```diff
@@ -55,6 +58,15 @@ class Converter : public HasInventory {
     if (this->converting || this->cooling_down) {
       return;
     }
+
+    // Check if we've reached max_conversions
+    if (this->max_conversions >= 0 && this->conversions_completed >= this->max_conversions) {
+      this->cooldown = -1;  // Set to permanent stop
```
sasmith, line 64: This works and it's possibly the most elegant solution; but my instinct is that we should combine the "have you hit max conversions" check with line 58, and not change `this->cooldown` or `this->cooling_down` -- we just never start converting again if we hit max production (since we always fail this check).

mettagrid/src/metta/mettagrid/objects/converter.hpp
```diff
@@ -105,24 +117,28 @@ class Converter : public HasInventory {
...
         output_resources(cfg.output_resources),
         max_output(cfg.max_output),
+        max_conversions(cfg.max_conversions),
         conversion_ticks(cfg.conversion_ticks),
         cooldown(cfg.cooldown),
         color(cfg.color) {
     GridObject::init(cfg.type_id, cfg.type_name, GridLocation(r, c, GridLayer::ObjectLayer));
     this->converting = false;
     this->cooling_down = false;
+    this->conversions_completed = 0;
```
sasmith, line 141: Move this to line ~137 as `conversions_completed(0)`.

mettagrid/src/metta/mettagrid/objects/converter.hpp
```diff
@@ -153,15 +172,16 @@ class Converter : public HasInventory {
...
-      // Negative cooldown means never convert again
-      this->cooling_down = true;
-      stats.incr("conversions.permanent_stop");
     }
   }
 
   void finish_cooldown() {
     this->cooling_down = false;
+    // Only update stats if there was actually a conversion
+    if (this->max_conversions != 0) {
```
sasmith, line 181: I don't think we should be able to get here with `max_conversions == 0`. If we can get here, can we change it so we can't?

Also, finishing cooldown isn't the same as finishing a conversion. It just means we're (maybe) ready to convert again.

mettagrid/src/metta/mettagrid/objects/converter.hpp
```diff
@@ -137,7 +153,10 @@ class Converter : public HasInventory {
 
   void finish_converting() {
     this->converting = false;
-    stats.incr("conversions.completed");
+    if (this->max_conversions != -1) {
```
sasmith, line 156: Why only update if `max_conversions != -1`?

configs/env/mettagrid/game/objects/mines.yaml
```diff
@@ -4,8 +4,9 @@ mine_red:
     ore_red: 1
   color: 0
   max_output: 5
+  max_conversions: -1
   conversion_ticks: 1
-  cooldown: 50
+  cooldown: 2
```
sasmith, line 9: let's not change these

configs/env/mettagrid/game/objects/advanced.yaml
```diff
@@ -6,6 +6,7 @@ lab:
   output_resources:
     blueprint: 1
   max_output: 5
+  max_conversions: -1
```
sasmith, line 9: With a default of `-1` in the config, I think we can leave this out.

sasmith (changes requested):

mettagrid/src/metta/mettagrid/objects/converter.hpp
lsoros: If we combine with line 58, then we aren't able to update the stats for conversions.permanent_stop. However, I can remove the code that changes cooldown and cooling_down.

configs/env/mettagrid/game/objects/mines.yaml
lsoros: That was for testing, forgot to change back to original value.

mettagrid/src/metta/mettagrid/mettagrid_config.py
sasmith: I see you've resolved this, but I don't see that a default has been added. Can you say more?

configs/env/mettagrid/game/objects/advanced.yaml
sasmith: It doesn't look like you're providing a default in the pydantic file -- just a lower bound (i.e., `ge=-1`, aka the value needs to be `>= -1`). Can you try providing `default=-1` in the pydantic file and see if that lets you drop these?

If not, then don't worry about it, since it'll be easy to clean this up in the future as needed.

mettagrid/src/metta/mettagrid/objects/converter.hpp
```diff
@@ -55,6 +58,13 @@ class Converter : public HasInventory {
     if (this->converting || this->cooling_down) {
       return;
     }
+
+    // Check if we've reached max_conversions
+    if (this->max_conversions >= 0 && this->conversions_completed >= this->max_conversions) {
+      stats.incr("conversions.permanent_stop"); // if you don't do it this way, and here, mine produce an extra ore
```
sasmith, line 64: I think we can drop the inline comment.

Also, I think this stat may be a little hard to interpret, since we may hit this multiple times for the same converter (we can hit this check any time someone adds inventory to the converter, for example). For now that's just a note, and not something we obviously need to fix.

sasmith (approved):
Looks good. See if you can clean up the defaults and then merge it.

PR #1538, opened 2025-07-17
feat: add gdb debug launcher for mettagrid

rjwalters:
This PR adds the ability to debug MettaGrid's C++ code using VS Code's integrated debugger. It addresses the team's need to set breakpoints in `.hpp` files and inspect variable values during runtime debugging.

### Problem
- Team members were unable to debug C++ code effectively
- Breakpoints in `.hpp` files weren't working
- No way to inspect C++ variable values at runtime
- Existing printf-style debugging was slowing down development

### Solution
Adds a debug build configuration that:
- Builds the C++ extension with full debug symbols (`-g3 -ggdb -O0`)
- Integrates with VS Code's C++ debugger (GDB/LLDB)
- Maintains the existing `uv sync` workflow for normal development
- Provides clear platform-specific setup instructions

### Changes
- **`.vscode/launch.json`**: Added three C++ debug configurations for common workflows (train, test, play)
- **`.vscode/tasks.json`**: Added `install-debug` task to build with debug symbols
- **`mettagrid/install-debug.sh`**: New script that builds the extension with debug symbols
- **`mettagrid/GDB_DEBUG.md`**: Documentation for the debugging workflow

### Usage
```bash
# One-time setup
chmod +x mettagrid/install-debug.sh

# Build with debug symbols
./mettagrid/install-debug.sh

# Then in VS Code:
# 1. Set breakpoints in any .hpp file
# 2. Select "C++ Debug: Train Metta" from debug dropdown
# 3. Press F5 to start debugging
```

### Platform Support
- **Linux**: Uses GDB (requires `apt-get install gdb`)
- **macOS**: Uses LLDB (comes with Xcode Command Line Tools)
- Script automatically detects platform and validates debugger installation

### Performance Note
Debug builds are intentionally unoptimized (-O0) for accurate debugging. Use regular `uv sync` for normal development to maintain performance.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -192,6 +192,117 @@
       },
       "python": "${workspaceFolder}/.venv/bin/python",
       "preLaunchTask": "${defaultBuildTask}"
+    },
+    {
+      "name": "C++ Debug: Train Metta",
+      "type": "cppdbg",
+      "request": "launch",
+      "program": "${workspaceFolder}/.venv/bin/python",
+      "args": [
+        "-m", "tools.train",
+        "+hardware=macbook",
+        "+user=${env:USER}"
+      ],
+      "stopAtEntry": false,
+      "cwd": "${workspaceFolder}",
+      "environment": [
+        {
+          "name": "PYTHONPATH",
+          "value": "${workspaceFolder}"
+        },
+        {
+          "name": "UV_PROJECT_ENVIRONMENT",
+          "value": "${env:VIRTUAL_ENV}"
+        }
+      ],
+      "externalConsole": false,
+      "MIMode": "gdb",
+      "setupCommands": [
+        {
+          "description": "Enable pretty-printing for gdb",
+          "text": "-enable-pretty-printing",
+          "ignoreFailures": true
+        },
+        {
+          "description": "Set breakpoint pending",
+          "text": "set breakpoint pending on"
+        }
+      ],
+      "preLaunchTask": "install-debug"
+    },
+    {
+      "name": "C++ Debug: Test",
+      "type": "cppdbg",
+      "request": "launch",
+      "program": "${workspaceFolder}/.venv/bin/python",
+      "args": [
+        "-m", "pytest",
+        "-v", "-s",
+        "./tests"
+      ],
+      "stopAtEntry": false,
+      "cwd": "${workspaceFolder}",
+      "environment": [
+        {
+          "name": "PYTHONPATH",
+          "value": "${workspaceFolder}"
+        },
+        {
+          "name": "UV_PROJECT_ENVIRONMENT",
+          "value": "${env:VIRTUAL_ENV}"
+        }
+      ],
+      "externalConsole": false,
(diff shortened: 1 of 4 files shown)


teodorionita:
@rjwalters just had a look at this today and I am a bit stuck on the C++ symbol debugging

Report below:

# Debug Symbols Investigation Report

## Attempted Solutions

### 1. Environment Variables
**Approach:** Set CMAKE_BUILD_TYPE=Debug, CMAKE_C_FLAGS, CMAKE_CXX_FLAGS  
**Expected:** Force debug settings through environment  
**Result:** Failed - scikit-build ignored these settings

### 2. Modified pyproject.toml
**Approach:** Changed cmake.args = ["--preset", "release"] to ["--preset", "debug-gdb"]  
**Expected:** Force scikit-build to use our debug preset  
**Result:** Failed - Build worked but final binary had no debug info

### 3. Anti-Stripping Measures
**Approach:** Set CMAKE_STRIP=/usr/bin/true, --config-settings=install.strip=false  
**Expected:** Prevent symbol stripping during packaging  
**Result:** Failed - Symbols were still missing

### 4. Linker Flags
**Approach:** Added -g3 to CMAKE_SHARED_LINKER_FLAGS, CMAKE_MODULE_LINKER_FLAGS  
**Expected:** Preserve debug symbols during linking  
**Result:** Failed - Final binary still stripped

## Investigation Findings

### Debug Symbol Journey

1. **Source Code (.cpp)** - Contains all debug information 
2. **Compilation (.cpp → .o)** - Object files have debug symbols (-g3 works) 
3. **Linking (.o → .so)** - Debug symbols present but getting processed 
4. **Python Packaging** - Final .so has no debug symbols 

### Root Cause: Multi-Layer Stripping

#### Layer 1: scikit-build-core
- Builds in isolated temporary directories
- Optimizes for package size and distribution
- Has its own symbol management logic

#### Layer 2: Python Extension Pipeline
- Python extensions are heavily optimized by default
- Multiple tools in the chain (pybind11, setuptools, wheel) can strip symbols
- Final .so files are expected to be small and fast

#### Layer 3: macOS Toolchain
- Apple's linker strips symbols by default in many contexts
- Code signing and distribution requirements favor stripped binaries
- Different behavior than Linux/GCC toolchain

#### Layer 4: Temporary Build Cleanup
- scikit-build cleans up intermediate files after building
- Object files with symbols are deleted
- Only the final stripped .so remains

## Technical Evidence

### File Analysis

```bash
# Object files during build: Had debug symbols 

# Final .so file:
dwarfdump --debug-info mettagrid_c.so
# Result: .debug_info contents: (empty)

# File structure:
file mettagrid_c.so
# Result: Mach-O 64-bit bundle arm64 (no mention of debug info)
```

### Critical Finding

```bash
dsymutil mettagrid_c.so
# Warning: no debug symbols in executable
# Warning: object files not found (deleted from temp directory)
```

The debug symbols are successfully generated during compilation but are stripped during the Python packaging process, with intermediate files being cleaned up, preventing symbol recovery.

PR #1542, opened 2025-07-17
Let users tag episodes in observatory

Barraketh:
First step towards having tag-specific dashboard

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210827596499490)

### DIFF
--- a/app_backend/src/metta/app_backend/routes/episode_routes.py
+++ b/app_backend/src/metta/app_backend/routes/episode_routes.py
@@ -29,7 +29,7 @@ def _parse_filter_query(filter_query: str) -> str:
 
 # Request/Response Models for Episode Tagging
 class EpisodeTagRequest(BaseModel):
-    episode_ids: List[str]
+    episode_ids: List[uuid.UUID]
     tag: str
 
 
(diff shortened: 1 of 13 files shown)


app_backend/src/metta/app_backend/routes/episode_routes.py
```diff
@@ -0,0 +1,325 @@
...
+    - created_at > '2023-01-01'
+    - policy_name=test AND created_at > '2023-01-01'
+
+    This is a simple implementation that handles common cases.
+    """
+    if not filter_query.strip():
+        return "1=1"
+
+    # Simple mapping of filter fields to SQL columns
+    field_mapping = {
```
nishu-builder, line 27: this feels like the sort of thing we will eventually break by accidentally renaming the tables aliases we use in a query...maybe we leave a comment about that? also doesn't work with self-joins

app_backend/src/metta/app_backend/routes/episode_routes.py
```diff
@@ -0,0 +1,325 @@
...
+            raise HTTPException(status_code=500, detail=f"Failed to filter episodes: {str(e)}") from e
+
+    # Episode tagging routes
+    @router.post("/tags/add", response_model=EpisodeTagResponse)
+    @timed_route("add_episode_tags")
+    async def add_episode_tags(request: EpisodeTagRequest, user: str = user_or_token) -> EpisodeTagResponse:
+        """Add a tag to multiple episodes."""
+        try:
+            # Convert string UUIDs to UUID objects
+            episode_uuids = [uuid.UUID(episode_id) for episode_id in request.episode_ids]
```
nishu-builder, line 201: cool thing i discovered: you can give your EpisodeTagRequest a uuid type for the episode_id, not a string type, and parsing with fastapi will work well with serializing and deserializing it. so you don't need to do this manually

app_backend/src/metta/app_backend/routes/episode_routes.py
```diff
@@ -0,0 +1,325 @@
...
+            raise HTTPException(status_code=500, detail=f"Failed to get episode tags: {str(e)}") from e
+
+    @router.get("/tags/all", response_model=AllTagsResponse)
+    @timed_route("get_all_episode_tags")
+    async def get_all_episode_tags(user: str = user_or_token) -> AllTagsResponse:
+        """Get all distinct tags that exist in the system."""
+        try:
+            tags = await stats_repo.get_all_episode_tags()
+            return AllTagsResponse(tags=tags)
+        except Exception as e:
```
nishu-builder, line 246: check out timed_http_handler; might save you some lines in this file

PR #1549, opened 2025-07-18
feat: external pr detection in workflows

rjwalters:
This is a simpler approach and an alternative to #1532 - I'll let @relh pick the one he prefers

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210828503101069)


relh:
this is great I like yours better

PR #1562, opened 2025-07-18
Reduce the use of random resolvers

sasmith:
Refactors the MettagGrid navigation sparse config to use the SampledTaskCurriculum, to remove randomness from OmegaConf resolution.

The main questionable thing here is that the configuration is all done in place, versus having one yaml file with the config template and another yaml file with the sampling parameters, like we do with bucketing. This leads to defaults feeling roundabout, since we need to apply them a layer down. This is sort of excused since we're _already_ applying them in a targeted way. Also, we were including a lot of defaults we didn't need, and this lets us remove them.

Happy for a second set of eyes on whether we should just break things into more files.

The sampling is covered by existing unit tests. I'm also training against this as a smoke test.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210836462023612)


jacklionheart (approved):
submitter of PRs removing  `${sampling:` from the codebase = 🐐

🙇🙇🙇🙇🙇

PR #1563, opened 2025-07-18
Movement metrics 

bluewin4:
## Primary Purpose

To add metrics that track how many times an agent rotates in a direction, moves in a direction, and performs a "sequential rotation" where multiple rotations happen in a row.

## Metrics Added
- **Direction Distribution** (`movement/facing/[up|down|left|right]`) - Tracks time spent facing each direction
- **Sequential Rotation Behavior** (`movement/sequential_rotations`) - Identifies indecision/search patterns through consecutive rotations

- Configurable via `track_movement_metrics` (default: `true`)
- API support: `Environment(track_movement_metrics=True)`
- Environment variable: `MOVEMENT_METRICS=true`


metta/rl/functions.py
```diff
@@ -685,6 +685,33 @@ def compute_timing_stats(
     }
 
 
+def filter_movement_metrics(stats: Dict[str, Any]) -> Dict[str, Any]:
```
daveey, line 688: Can we just do this for all metrics? I find the .xxx metrics useless and they pollute Wandb

recipes/navigation.sh
```diff
@@ -4,10 +4,14 @@
 #We expect this recipe to achieve >88% on the navigation evals
 #For baseline to compare, see wandb run: https://wandb.ai/metta-research/metta/runs/daphne.navigation.low_reward.1gpu.4agents.06-25
 
+# Set default value for movement metrics
```
daveey, line 7: Don't use env for this, just pass the command line arg

metta/rl/functions.py
bluewin4: I would agree, although we would need some principled way of doing this. Probably with some form of flag of "extensive_wandb_metrics" or something if someone needs them.

mettagrid/src/metta/mettagrid/mettagrid_c.cpp
```diff
@@ -87,6 +90,7 @@ MettaGrid::MettaGrid(const GameConfig& cfg, const py::list map, unsigned int see
     } else if (action_name_str == "move") {
       _action_handlers.push_back(std::make_unique<Move>(*action_config));
     } else if (action_name_str == "rotate") {
+      _rotate_action_index = _action_handlers.size();  // Store index before adding
```
sasmith, line 93: Rather than track at the mettagrid_c level, it would seem much more natural if we tracked in the action class. Can we instead track inside of `rotate` and `move`?

Maybe there's a shared `action_metrics` object they share or something?

sasmith (changes requested):
Returning since I think this should be done within actions, to keep things cleaner.

I'll also want to take a look at the metric filtering, but @rjwalters has more recently owned "cleaning up metrics", so he might be a better choice.

mettagrid/src/metta/mettagrid/mettagrid_c.cpp
bluewin4: There's no `action_metrics` object I could find.

Movement tracking is easy, but the sequential rotations metric requires knowing other actions which is hard to do on the action layer without the object. I can try to make one if you think it's important to get sequential_rotation in actions

PR #1565, opened 2025-07-18
Cursor background agent compatibility

nishu-builder:
Yesterday, I snuck in a change to allow `softmax-docker` to be specifiable as a profile. It has [these](https://github.com/Metta-AI/metta/blob/main/metta/setup/config.py#L74) defaults. 

Fixes `install.sh` to be runnable on the default docker container for cursor agents. Also fixes some passing of profiles through `install.sh` that were previously broken.

the softmax-docker profile tries to install and connect to our aws and wandb accounts. If you don't provide corresponding WANDB_API_KEY and aws api keys as Background Agent secrets (which cursor injects as environment vars in the agent's runtime), then it won't be able to connect to those, but it should still get through everything else just fine; they are optional.


rjwalters (approved):
👍

PR #1573, opened 2025-07-20
imports to top

relh:
moves imports to the top where found because people like this

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210839876813225)


relh:
tested local, mettabox, distributed

PR #1589, opened 2025-07-21
Disable stats metadata collection to reduce memory usage

daveey:
### TL;DR

Disabled detailed statistics tracking to reduce memory usage and improve performance.

- system and memory stats were sufficiently slow as to contribute to training time
- the .avg / .stdev stats were never used and just poluted wandb

### What changed?

- Commented out the system and memory monitoring stats collection in `metta/rl/trainer.py`, replacing them with empty dictionaries
- Disabled the calculation and tracking of metadata statistics in `StatsTracker.to_dict()` method, including:
  - First/last seen timestamps
  - Update counts and rates
  - Min/max values
  - Average calculations
  - Activity rate calculations
- Updated corresponding tests to reflect the disabled functionality

### How to test?

1. Run the trainer with a memory-intensive workload and verify reduced memory usage
2. Check that basic statistics are still being collected and reported correctly
3. Verify that the tests pass with the updated expectations

### Why make this change?

The detailed statistics tracking was consuming excessive memory and processing resources during training. This change reduces overhead by temporarily disabling non-essential statistics calculations while preserving core functionality. This should improve performance and reduce memory pressure for large-scale training runs.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210849040011815)

### DIFF
--- a/metta/rl/trainer.py
+++ b/metta/rl/trainer.py
@@ -859,14 +859,17 @@ class MettaTrainer:
 
         # Add hyperparameter values
 
+        system_stats = {}  # self._system_monitor.stats()
+        memory_stats = {}  # self._memory_monitor.stats()
+
         # Build complete stats dictionary for wandb
         all_stats = build_wandb_stats(
             processed_stats=processed_stats,
             timing_info=timing_info,
             weight_stats=weight_stats,
             grad_stats=self.grad_stats,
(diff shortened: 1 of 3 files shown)


sasmith (commented):
what's the plan for follow up? Should we just delete the excess stats?

PR #1590, opened 2025-07-21
Add kubectl to dockerfile that will be used for orchestrator

nishu-builder:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210849708423730)

### DIFF
--- a/app_backend/Dockerfile
+++ b/app_backend/Dockerfile
@@ -1,16 +1,18 @@
 # Use Python 3.11 slim image as base
 FROM python:3.11-slim
 
-# Install system dependencies
+# Install system dependencies and add Kubernetes apt repository
 RUN apt-get update && apt-get install -y \
     curl \
     git \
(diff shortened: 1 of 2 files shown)


app_backend/Dockerfile
```diff
@@ -7,6 +7,11 @@ RUN apt-get update && apt-get install -y \
     git \
     && rm -rf /var/lib/apt/lists/*
 
+# Install kubectl for orchestrator
```
Barraketh, line 10: Is there any way we can install kubernetes using apt-get like above? In particular, I don't love the '/amd64' dependency - I am not sure if we'll ever want arm images for local development, but this prevents it

nishu-builder: surely we can. :) sec

nishu-builder: better now @Barraketh

PR #1597, opened 2025-07-21
Eval stats worker authenticates with stats server

nishu-builder:
Worker posts to specified stats server uri and uses token

Orchestrator collects and passes machine token

`metta local kind {x}` replaces `kind.sh`, adds new quality of life commands, and automatically passes tokens

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210852033370090)


app_backend/src/metta/app_backend/container_managers/base.py
```diff
@@ -27,6 +27,7 @@ def start_worker_container(
         git_hash: str,
         backend_url: str,
         docker_image: str,
+        machine_token: str | None = None,
```
Barraketh, line 30: When can machine token be None?

nishu-builder: you're right there's no good case for it being None, will fix

Barraketh (approved):

PR #1602, opened 2025-07-22
fix: max reward calculation in Agent::_update_resource_reward

rjwalters:
Fixes a bug where per-item reward caps were incorrectly applied to the total reward instead of individual item contributions, causing incorrect reward calculations when agents have multiple resource types.

## Bug Description
```cpp
// BEFORE (incorrect):
this->current_resource_reward += reward_it->second * (new_amount - old_amount);
if (this->resource_reward_max.count(item) > 0) {
  // BUG: This caps the TOTAL reward from all items, not just this item
  this->current_resource_reward = std::min(current_resource_reward, this->resource_reward_max.at(item));
}
```

This would cause scenarios like:
- Agent has 10 hearts (no cap) = 10.0 reward
- Agent gains 100 ore (capped at 2.0) = should add 2.0 reward
- Bug: Total reward would be capped to 2.0 instead of 12.0

## Solution
1. Calculate the old contribution (with cap applied)
2. Calculate the new contribution (with cap applied)
3. Update rewards by the difference

This ensures each item's reward contribution respects its individual cap while maintaining correct total rewards.

## Added new tests:
- ✅ Basic capping behavior when adding/removing items
- ✅ **edge case**: Removing items when over the cap keeps reward at cap
- ✅ Multiple item types with different caps interact correctly
- ✅ Edge case: Removing all items brings reward to zero

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210852662941444)


daveey (approved):
Nice find, thanks for fixing!

PR #1603, opened 2025-07-22
Disable unneeded actions from nav evals

sasmith:
Disables unnecessary actions in navigation evaluation configurations.

Navigation tasks only require movement actions to complete their objectives. Disabling unnecessary actions simplifies the action space for navigation-specific evaluations, making them more focused and potentially improving agent performance by removing irrelevant action choices.

Mostly: it seems like "navigate while dealing with a bunch of irrelevant actions" is a different skill from "navigate", and so it's reasonable for us to separate these.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210853082082452)

### DIFF
--- a/configs/env/mettagrid/navigation/evals/defaults.yaml
+++ b/configs/env/mettagrid/navigation/evals/defaults.yaml
@@ -13,6 +13,16 @@ game:
     last_reward: true
     resource_rewards: false # This feature wasn't present when older policies were trained
 
+  # Disable unneeded actions from nav evals
+  actions:
+    put_items:
+      enabled: false
+    attack:
+      enabled: false
+    swap:
+      enabled: false
+    change_color:
+      enabled: false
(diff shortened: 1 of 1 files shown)


sasmith:
FWIW, I don't feel super strongly about this, but I've had this in a branch for a while and I'm trying to clean things out. And I think this is more good than bad.

sasmith:
> Is this going to make policies trained on navigation incompatible with other environments where these actions are enabled? Like, will I be able to evaluate a policy that has these actions enabled, on these evals, without getting an error?

No, that shouldn't be a problem. If it is, it's a bug.

What should happen is that the policy should be emitting some sort of "what I want to do" vector in latent space, and then we match it to the closest action that's available.

georgejwdeane (approved):

PR #1604, opened 2025-07-22
Policy evaluator helm charts

nishu-builder:
Works locally on kind. Tried to keep the difference between kind and eks versions minimal

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210853054436870)


devops/docker/Dockerfile.local
```diff
@@ -22,12 +22,6 @@ WORKDIR /workspace
 # Copy the entire repo
 COPY --chown=linuxbrew:linuxbrew . /workspace/metta
 
-# Copy .metta directory
```
Barraketh, line 25: Why are we doing this? I'm pretty sure this is gonna break co-pilot

devops/charts/orchestrator/templates/role.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
Barraketh, line 1: Do we need all these permissions?  I thought our kube manager only interacted with pods

devops/docker/Dockerfile.local
nishu-builder: Dockerfile.local was something I introduced a bit ago for local testing of policy evaluator; dont think it's in use anywhere else. It copied over my ~/.metta/ in part to get observatory tokens. But that isn't necessary now if we're passing machine token directly through secrets

devops/charts/orchestrator/templates/role.yaml
nishu-builder: good point! removed most. only getting secrets and pod interactions

Barraketh (approved):

PR #1606, opened 2025-07-22
chore: improve CI test parallelism

nishu-builder:
This switches unit tests to a matrix strategy which helps speed things up a bit.
We also adjusted some timeouts and improved the uv setup cache strategy -- although that did not make a bit difference.

We should revisit this after we finish up #1385 since the core unit tests are a clear bottleneck.

Edited: I've had CI fail because it exceeds 15 minutes several times. This attempts to speed it up

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210853241319543)

### DIFF
--- a/.github/actions/setup-uv/action.yml
+++ b/.github/actions/setup-uv/action.yml
@@ -35,13 +35,8 @@ runs:
         case "${{ inputs.install-mode }}" in
 
           "testing")
-            uv pip install --no-deps -e .
-            uv pip install --no-deps -e ./common
-            uv pip install --no-deps -e ./agent
-            uv pip install --no-deps -e ./app_backend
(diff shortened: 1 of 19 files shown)


.github/actions/setup-uv/action.yml
```diff
@@ -30,22 +30,6 @@ runs:
 
         # Handle different installation modes
         case "${{ inputs.install-mode }}" in
-          "no-mettagrid")
```
rjwalters, line 33: I think you will still need `uv pip install pytest pytest-cov gcovr` if you refactor how this works...

nishu-builder: thanks! gcovr is covered but added the others

PR #1607, opened 2025-07-22
Heavily reduce the amount of $sampling

sasmith:
Remove $sampling from most environment configs by replacing it with the provided mid value (or an appropriate seeming value when the parameters were invalid).

The direction we want to move is to have randomness at fewer layers. It should be controlled above for things we want to be able to adjust in curricula, and below for things we don't care about. Aka, randomness in BucketCurricula is okay, and randomness in mapgen is okay, and randomness in OmegaConf is not.

This change does not preserve any randomness in the files is touches. I want feedback on that! This doesn't touch the nav curricula, since I want to preserve randomness there, and so I'll want to convert things to bucketed curricula; but that takes more work. My hope here is to spend an amount of effort proportional to how much we care.

So, if any of these are places where we think the randomness is important, please speak up, and I'll happily do a more random-preserving change instead.

After this, I'll also need to update $choice.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210853702533283)

### DIFF
--- a/configs/env/mettagrid/extended_sequence/backchain1.yaml
+++ b/configs/env/mettagrid/extended_sequence/backchain1.yaml
@@ -17,12 +17,12 @@ game:
       battery_green: 1
     rewards:
       inventory:
-        ore_red: ${sampling:0.005,0.1,0.01}
-        battery_red: ${sampling:0.01,0.3,0.1}
-        ore_blue: ${sampling:0.005,0.1,0.01}
-        battery_blue: ${sampling:0.01,0.3,0.1}
-        ore_green: ${sampling:0.005,0.1,0.01}
-        battery_green: ${sampling:0.01,0.3,0.1}
+        ore_red: 0.01
+        battery_red: 0.1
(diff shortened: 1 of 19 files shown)


sasmith:
* **#1607** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1607?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1607?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

daphnedemekas (approved):
Hey, my thoughts on this are: 
For navigation sequence / multiagent, I was planning on merging in changes to our curricula there to move towards bucketed. I haven't gotten around to doing that yet since we aren't prioritizing these experiments at the moment, but once we are back to doing that, it's first order of business. So I'm happy with this basically, and also, we will still need to re-implement randomness using bucketed.

PR #1608, opened 2025-07-22
fix variable interpolation

sasmith:
Fixed altar configuration in sparse bucketed navigation config to use the `num_altars` variable.

In a last minute change to a prior PR, I had swapped the {num_altars} for ???. This is wrong in this context, as the bucketing is done over {num_altars}, not `game.map_builder.root...`. Moreover, I don't think the bucketing can be done over the latter, since the we need to go through a list (vs a dict).

Regardless, this is the version that currently works, and that's better than not working.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210855214856568)


daphnedemekas (approved):

sasmith:
Well I _meant_ to merge it!

PR #1611, opened 2025-07-22
fix: correct discord post character limit calculation

rjwalters:
The Discord webhook was failing to send messages because chunks were being split at exactly 2000 characters, but then a 10-character prefix (`"...\r\n   \r\n"`) was added to each chunk, causing them to exceed Discord's 2000 character limit.

### Solution
- Calculate the effective maximum chunk length as `2000 - len(MESSAGE_PREFIX)` before splitting
- Add a safety check to truncate any chunks that still exceed the limit after prefixing
- Define `MESSAGE_PREFIX` as a constant for better maintainability
- Update type hints to use built-in `list` and `dict` instead of `List` and `Dict` from typing

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210860699819477)


dominik3141 (approved):
Good find!

PR #1614, opened 2025-07-22
Load weights only on master rank

dominik3141:
## Summary
- avoid loading policy weights on worker ranks so DDP can broadcast them

## Testing
- `ruff format metta/rl/trainer.py`
- `ruff check metta/rl/trainer.py`
- `mypy metta/rl/trainer.py` *(fails: missing stubs and annotations)*
- `pytest -q` *(interrupted after ~100s)*

------
https://chatgpt.com/codex/tasks/task_e_688009a682d48325b5bfae3277815fd2

### DIFF
--- a/metta/rl/trainer.py
+++ b/metta/rl/trainer.py
@@ -21,7 +21,6 @@ from metta.agent.policy_store import PolicyStore
 from metta.app_backend.stats_client import StatsClient
 from metta.common.profiling.memory_monitor import MemoryMonitor
 from metta.common.profiling.stopwatch import Stopwatch, with_instance_timer
-from metta.common.util.fs import wait_for_file
 from metta.common.util.heartbeat import record_heartbeat
 from metta.common.util.system_monitor import SystemMonitor
 from metta.common.wandb.wandb_context import WandbRun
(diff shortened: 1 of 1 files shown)


daveey (commented):
I'll review this in a bit, but meanwhile, what do you think about not passing PolicyStore to the MettaTrainer when it's not master?

dominik3141:
> I'll review this in a bit, but meanwhile, what do you think about not passing PolicyStore to the MettaTrainer when it's not master?

Good idea. Also no need to review this yet (which is why the PR is still a draft and not assigned to anyone). This is a first version by CodeX but I think it isn't optimal yet. Gonna assign it to you once I'm happy with it.
(It might be confusing that our GitHub action requests reviews even if the PR is just a draft, maybe we should change that).

metta/rl/trainer.py
```diff
@@ -177,77 +176,51 @@ def __init__(
                 self.timer.load_state(checkpoint.stopwatch_state, resume_running=True)
 
         # Load or create policy with distributed coordination
-        policy_record = self._load_policy(checkpoint, policy_store)
+        distributed = torch.distributed.is_initialized()
+        load_here = not (distributed and not self._master)
```
daveey, line 180: i think self._master is sufficient

metta/rl/trainer.py
```diff
@@ -177,77 +176,51 @@ def __init__(
...
-                and "original_feature_mapping" in policy_record.metadata
-            ):
-                self.policy.restore_original_feature_mapping(policy_record.metadata["original_feature_mapping"])
-                logger.info(f"Rank {self._rank}: Restored original_feature_mapping")
-
-            # Initialize the policy to the environment
-            self._initialize_policy_to_environment(self.policy, metta_grid_env, self.device)
-
-            self.initial_policy_record = policy_record
+        # Broadcast metadata from master to workers so they can restore mappings
```
daveey, line 197: can we move as much of this into rl/policy.py:load_or_create_policy() or something like that? reducing trainer.py code is godliness

daveey (changes requested):
thanks for doing this, sorry to make you do more cleanup, but i think it'll be nicer.

metta/rl/trainer.py
dominik3141: I will. There is too much happening in trainer right now.

dominik3141:
> I'll review this in a bit, but meanwhile, what do you think about not passing PolicyStore to the MettaTrainer when it's not master?

I'm gonna do that in a separate PR tomorrow as it requires another large set of changes and I don't want this PR to be too much of a pain to review.

PR #1617, opened 2025-07-23
Make example notebook slightly more interesting

nishu-builder:
Very simplistic, maybe a hair better than what was there before

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210863587299280)


rjwalters:
these diffs are really ugly... we should look into using https://jupytext.readthedocs.io/en/latest/

relh:
I think we should turn the `run.py` into an example notebook as it's a great use-case for showing researchers how everything works without them getting their hands dirty in the codebase

PR #1640, opened 2025-07-23
mv notebooks experiments/notebooks

jacklionheart:
# Move notebooks/ under experiments/

This moves notebooks/ under experiments/ with the idea that over time we will also move configs into experiments (while converting yaml -> python) and build tooling that marries the "input" of configs into the "output" of notebooks.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210874558484694)


nishu-builder (approved):
Nice thank you! Maybe want to update experiments/README.md real quick

jacklionheart:
in the next revision i have something

PR #1642, opened 2025-07-24
Converter limits

lsoros:
This pull request replaces https://github.com/Metta-AI/metta/pull/1537, which will be closed. 

In the object config files, there is now an optional field for specifying a max number of conversion operations (param: max_conversions) that can be performed. Default value of -1 means unlimited conversions.

Tested with values -1, 0, 1, and 2 on red mines.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210875413677367)


mettagrid/src/metta/mettagrid/objects/converter.hpp
```diff
@@ -172,10 +187,6 @@ class Converter : public HasInventory {
     } else if (this->cooldown == 0) {
       // No cooldown, try to start converting again immediately
       this->maybe_start_converting();
-    } else if (this->cooldown < 0) {
```
sasmith, line 175: Was this intentionally removed? It seems unrelated.

sasmith (changes requested):
Looks good except for the code around cooldown being removed.

sasmith:
The failing tests come from a couple of places where we instantiate ConverterConfigs in cpp tests. You'll need to change them to include the new property.

mettagrid/src/metta/mettagrid/objects/converter.hpp
lsoros: I will revert that edit. 

It was because of a misunderstanding / mis-scoping of https://github.com/Metta-AI/metta/pull/1537#discussion_r2214417469 when I was still getting used to the codebase.

lsoros:
The requested change reverting the cooldown issue was also added in during this series of commits. The rest was modifying the tests.

sasmith (approved):

mettagrid/src/metta/mettagrid/objects/converter.hpp
rjwalters: this entire case should be removed - cooldown is uint so this is just unreachable code. alternatively if we want to preserve this meaning we should change the parameter to be short and the pydantic to have different restrictions...

PR #1653, opened 2025-07-24
pytorch out of bounds error fix. vibecoded this. hope it helps

zfogg:
it actually fixes the problem!

```
./tools/sim.py +hardware=macbook stats_server_uri=http://localhost:8000policy_uri=wandb://run/relh.skypilot.fff.j20.666:v11

^ causes a crash because of tensor indices out of bounds
```

this pull request fixes that

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210883457503416)


relh:
I think we should tackle this in the more complicated/involved rehydrate/dehydrate feature we're working on

agent/src/metta/agent/metta_agent.py
```diff
@@ -242,22 +242,23 @@ def _create_feature_remapping(self, features: dict[str, dict], is_training: bool
 
     def _apply_feature_remapping(self, features: dict[str, dict], unknown_id: int):
         """Apply feature remapping to observation component and update normalizations."""
-        # Update observation component if it supports remapping
-        if "_obs_" in self.components and hasattr(self.components["_obs_"], "update_feature_remapping"):
```
relh, line 246: I think we liked this if statement to make this not always happen. don't worry you don't have to address these zach

relh:
Good job finding this, closing because https://github.com/Metta-AI/metta/pull/1764 fixes

PR #1654, opened 2025-07-24
Switch navigation training to be bucketed curricula instead of using random resolvers

sasmith:
The main thing this does is to switch all of the navigation training environments to be bucketed curricula; so their randomness can exist there instead of in $sampling resolver.

This also renames `defaults` to be `terrain_from_numpy_defaults`, since I've gotten confused about this for some time.

Smoke tested by training and playing localled.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210883516905365)

### DIFF
--- a/configs/env/mettagrid/curriculum/all.yaml
+++ b/configs/env/mettagrid/curriculum/all.yaml
@@ -2,7 +2,7 @@ _target_: metta.mettagrid.curriculum.random.RandomCurriculum
 
 tasks:
   /env/mettagrid/curriculum/arena/learning_progress: 5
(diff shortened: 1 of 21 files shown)


bluewin4:
I think this makes sense, but one thing I would push on is where this is recorded for other people to know how to adapt to this change? 

Like, if I'm trying to make a new config, beside combing through the ones modified and trying to copy the structure, is there an explanation as to what goes where and why?

bluewin4 (approved):
Looks good to me. We really need a proper readme for these configs and some examples.

PR #1656, opened 2025-07-24
Split up core tests

nishu-builder:
"Core" tests (remainder of tests not within the recognized namespace packages) take the longest.

This splits those up into sub-groups based on functionality

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210883994152984)


rjwalters (commented):
image

I think this approach could speed things up a bit while we are waiting on the repo reorg... but it's also probbaly more complex than it needs to be

my suggestion would be to just manually break core and/or mettagrid into two parts so that the test-matrix job is about 4 min to make the smoke-test

rjwalters (commented):
4m 48s

PR #1658, opened 2025-07-24
Request policy evals with ./tools/sim.py remote=true

nishu-builder:
- Supports ./tools/request_eval.py --policy policy1 --policy policy2 --eval eval_name, with other optional arguments for git_hash. Equivalently, `metta tool run request_eval {args}`
- Fixes worker orchestrator to operate on version of codebase within git-hash-specific checkout
- Refactors a few semi-redundant entrypoints for fetching policies from wandb and loading them into the policy store
- Updates `metta local load-policies`
- Emits a link to the (local or prod) eval tasks page when done

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210884462208038)

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210884467102090)


app_backend/src/metta/app_backend/stats_client.py
```diff
@@ -267,3 +270,7 @@ def record_episode(
         response_data = response.json()
         episode_id_uuid = uuid.UUID(response_data["id"])
         return ClientEpisodeResponse(id=episode_id_uuid)
+
+    def create_task(self, request: TaskCreateRequest) -> TaskResponse:
```
nishu-builder, line 274: alternatively we can merge StatsClient and EvalTaskClient

Barraketh: The other possibility is moving create_task into 'StatsClient'.  I know that sounds somewhat weird, but it actually makes sense from the point of view that there's a set of requests that are made from the outside (including create_eval_task), and a set of requests that are made from the inside by task workers. I was even considering putting them on different route prefixes - having 'create_task' live on an authenticated route, but having all of the worker endpoints live on un-authenticated routes that are not exposed externally by the ingress

nishu-builder: That's smart actually! I'll do that

nishu-builder: I will do this in another PR because I had some other thoughts on common code for the clients :)

PR #1660, opened 2025-07-24
[MettaScope] trace panel caching

monofuel:
- add flags for `cachable` and `isDirty` to try to help trace panel performance
- this solves some performance problems when we are not interacting with the trace panel. but it's still bad when panning the panel. needs more thinking

- `cacheable` meshes get preserved between frames and do not get cleared automatically.
- `isDirty` gets set on mesh modification, to tell us to update the GPU buffer.
- cacheable non-dirty buffers do not update on GPU and save us much time and cpu.

- also included some performance fixes to getAttr().


mettascope/src/context3d.ts
```diff
@@ -4,31 +4,31 @@ const VERTEX_SHADER_SOURCE = `
   attribute vec2 a_position;
   attribute vec2 a_texcoord;
   attribute vec4 a_color;
-  
+
```
monofuel, line 7: my vscode keeps removing these spaces for some reason....

mettascope/src/replay.ts
```diff
@@ -6,18 +6,10 @@ import { updateAgentTable } from './agentpanel.js'
...
-    return defaultValue
-  } else if (obj[attr] instanceof Array) {
-    return obj[attr][atStep]
-  } else {
-    // This must be a constant that does not change over time.
-    return obj[attr]
-  }
+  const prop = obj[attr]
+  if (prop === undefined) return defaultValue
+  if (!Array.isArray(prop)) return prop // This must be a constant that does not change over time.
```
monofuel, line 11: I read that Array.isArray() is faster because it avoids prototype chain traversal (not sure if it actually results in performance benefits in this specific case)

mettascope/src/context3d.ts
treeform: do we need these spaces?

mettascope/src/traces.ts
```diff
@@ -61,9 +74,34 @@ export function drawTrace(panel: PanelInfo) {
...
+    lastCachedState.selection = state.selectedGridObject
+    lastCachedState.zoomLevel = panel.zoomLevel
+    lastCachedState.panPos = new Vec2f(panel.panPos.x(), panel.panPos.y())
   }
 
+  // Clear and regenerate ALL content only when needed
+  if (!shouldRegenerate) {
+    return
+  }
+  console.log('Regenerating trace')
```
treeform, line 100: remove log

treeform (approved):
Please next PR abstract the camera matrix into a field on mesh and pass it as a uniform to a shader so that it does not need to re-render if you pan or zoom.

PR #1664, opened 2025-07-25
Recreate the <Heatmap /> component from Observatory, but as a Juypter notebook widget

zfogg:
It went well. Next up is to import and use the <Heatmap /> component for real, so gonna try to change all of this. Hope it works.

Here's what it looks like inside of Jupyter
image

I'm gonna keep committing and pushing to this branch so no one try to merge this just yet

The main problem right now is that it looks bad in Cursor.app and I don't know why. Can't do anything about it apparently. ipywidgets look bad too.

Jupyter looks like this:
image

Cursor looks like this:
image

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210885046079715)


zfogg:
@nishu-builder all webview content has the same problem in Cursor's / VSCode's Jupyter webview renderer right now.. can barely get it to render, it will have the wrong styles or just not show up while other things do show up

If we wanna provide this to people we have to accept that right now until they fix widgets / webviews. they're broken in Cursor i guess.

well maybe we could try to fix it in vscode and send a pull request, as vscode IS open source :P then cursor will get it downstream

experiments/notebooks/utils/heatmap_widget/heatmap_widget/metta_client.py
```diff
@@ -0,0 +1,285 @@
...
+from typing import List, Optional
+
+import httpx
+
+from metta.common.util.stats_client_cfg import get_machine_token
+
+from .HeatmapWidget import HeatmapWidget, create_heatmap_widget
+
+
+class MettaAPIClient:
```
nishu-builder, line 11: this can borrow some behavior from EvalTaskClient (could subclass it) to get auth/etc behavior out of the box

experiments/notebooks/utils/heatmap_widget/heatmap_widget/metta_client.py
```diff
@@ -0,0 +1,285 @@
+# 🔧 Fixed API Client Implementation
+from typing import List, Optional
+
+import httpx
+
+from metta.common.util.stats_client_cfg import get_machine_token
+
+from .HeatmapWidget import HeatmapWidget, create_heatmap_widget
```
nishu-builder, line 8: we should use absolute imports

experiments/notebooks/utils/heatmap_widget/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+from .heatmap_widget.HeatmapWidget import (
+    HeatmapWidget,
+    create_demo_heatmap,
+    create_heatmap_widget,
+    create_multi_metric_demo,
+)
+from .heatmap_widget.metta_client import MettaAPIClient, fetch_real_heatmap_data
+
+__all__ = [
```
nishu-builder, line 11: dont love the __all__ pattern and we tend not to use i think

experiments/notebooks/utils/heatmap_widget/heatmap_widget/metta_client.py
```diff
@@ -0,0 +1,285 @@
...
+                print(f"🔄 Trying with just the first eval name: {eval_names[0]}")
+                result = await client.get_available_metrics(training_run_ids, [], [eval_names[0]])
+                print(f"✅ Single eval name worked! Got {len(result)} metrics")
+                return result
+            except Exception as e2:
+                print(f"❌ Single eval name also failed: {e2}")
+
+        print("💡 Falling back to common metrics...")
+        # Return some common metrics as fallback
+        return ["reward", "heart.get", "ore_red.get", "action.move.success"]
```
nishu-builder, line 285: dont think we want to be taking an opinion on what common metrics are like this

experiments/notebooks/research/example.ipynb
```diff
@@ -1,199 +1,193 @@
...
-            "mimetype": "text/x-python",
-            "name": "python",
-            "nbconvert_exporter": "python",
-            "pygments_lexer": "ipython3",
-            "version": "3.11.7"
-        }
-    },
-    "nbformat": 4,
-    "nbformat_minor": 4
+ "cells": [
```
nishu-builder, line 2: is this change the result of linting? if so let's revert; if you rebase on main it should ignore linting on ipynb files, which is what we want I think

experiments/notebooks/mettabook-template.ipynb
```diff
@@ -34,7 +34,6 @@
    "source": [
     "%load_ext autoreload\n",
     "%autoreload 2\n",
-    "import os\n",
```
nishu-builder, line 37: same here (linting comment)

nishu-builder (dismissed):
Hard to comment directly on the heatmap_widget_...ipynb file so I'm putting those comments here: 
- I think we should extract the relevant sections of this and put them in mettabook-template.ipynb
- Shouldn't have an Installation section; we should just copy what's in mettabook-template.ipynb
- I think we can remove the demo heatmap with fake data, and can remove the fallback-to-demo-heatmap behavior
- We're not using eval_filter in the "Example: simply get metrics for policies from the API" section
- I think we do'nt need the multiple-examples or the comments explaning much of the bbehavior in the "Multiple metrics with working selectedMetric" section -- I think it's probably self-explanatory
- I think for now we remove the custom callbacks thing because researchers aren't going to use it until we have a slightly more built out use case, and until then it'll probably just be confusing/dirty their working dirs
- the "Data Cell Format Reference" could probably be moved up to the "Custom Metrics" section

I prefer fewer logs and emojis, but that's just my personal preference

experiments/notebooks/utils/heatmap_widget/src/Heatmap.tsx
```diff
@@ -0,0 +1,221 @@
+import { useState } from 'react'
```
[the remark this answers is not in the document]
zfogg: @nishu-builder right now Pasha said to just copy it. it doesn't work with a symlink, and it needs to be in this directory for the build system to process it properly. Pasha and I were talking about turning Observatory into an NPM package to solve this. we could use GitHub's package registry or just npm itself. all the other ways i tried to get this to work error on build :/

experiments/notebooks/utils/heatmap_widget/heatmap_widget/metta_client.py
zfogg: the code doesn't actually work with an absolute import here. only relative imports work. i think it might because of what i'm doing in __init__.py.... but it's allowed and "good practice" in python code to do it this way (i think). Cursor keeps correcting it to a relative import and said this is the way to do it with the current setup

experiments/notebooks/utils/heatmap_widget/heatmap_widget/metta_client.py
zfogg: removed

experiments/notebooks/research/example.ipynb
zfogg: done

experiments/notebooks/mettabook-template.ipynb
zfogg: done

experiments/notebooks/utils/heatmap_widget/heatmap_widget/metta_client.py
```diff
@@ -0,0 +1,285 @@
...
+    try:
+        client = MettaAPIClient(api_base_url)
+
+        # Debug: Let's try with a simplified version first
+        print(f"🔍 Attempting to get metrics for {len(eval_names)} eval names...")
+        print(f"📋 Eval names: {eval_names}")
+
+        return await client.get_available_metrics(training_run_ids, [], eval_names)
+
+    except Exception as e:
```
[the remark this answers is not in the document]
zfogg: fixed

experiments/notebooks/utils/heatmap_widget/heatmap_widget/metta_client.py
```diff
@@ -0,0 +1,285 @@
+# 🔧 Fixed API Client Implementation
```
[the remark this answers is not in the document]
zfogg: ok i put the file in there in its own module

experiments/notebooks/utils/heatmap_widget/__init__.py
zfogg: ok i went ahead and removed those and rewrote all the necessary `import` lines of code

experiments/notebooks/utils/heatmap_widget/heatmap_widget/util.py
```diff
@@ -0,0 +1,129 @@
...
+    # TODO: backend should be doing the filtering, not frontend
+    policies_data = await client.get_policies(page_size=100)
+
+    # Find training run IDs that match our training run names
+    training_run_ids = []
+    for policy in policies_data.policies:
+        if policy.type == "training_run" and any(
+            run_name in policy.name for run_name in training_run_names
+        ):
+            training_run_ids.append(policy.id)
```
zfogg, line 42: @Barraketh this logic should go in heatmap_routes.py probably... we're doing this in the python notebook but it should be done by the API.

Barraketh:
I kinda like it as it's own file too, though I don't have strong feelings about it

Barraketh (approved):

PR #1682, opened 2025-07-25
Support pnpm in metta.setup

berekuk:
This PR improves the setup system for JavaScript dependencies by:

- Creating a dedicated `pnpm` setup component that handles JS dependencies installation
- Adding component dependency management with topological sorting to ensure proper installation order
- Moving the package manager specification to a root-level package.json
- Removing redundant package manager specifications from individual package.json files
- Consolidating frontend setup by replacing component-specific modules with the centralized pnpm approach
- Adding a `dependencies()` method to the base setup module class to declare dependencies

The changes ensure that components requiring pnpm (like mettascope) will have their dependencies properly installed first.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210891574874706)


nishu-builder (approved):
this is so great. do we still want to build gridworks and observatory-fe? i suppose we really dont have to in `metta install`; that could be left to e.g. other commands like `metta observatory run`

berekuk:
> i dont know enough about how pnpm works but just flagging that the observatory dockerfile uses `npm run build`

Ooh thanks for flagging. I updated that to `pnpm run build` (in #1681), but I almost forgot that there's still more to do there. With the top-level lockfile the docker build is more involved, the package's dir content is not enough context.

PR #1685, opened 2025-07-25
Fix trusted keys in flake file

monofuel:
- fix the flake to use extra-trusted-public-keys to match extra-substituters
- I think the first time I tested this I had added the trusted public key to my system outside of the flake, and didn't notice issues with it missing.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210891878782332)


treeform (approved):
nix is nix

PR #1686, opened 2025-07-25
added action target arrows to pair with thought bubbles

lcodes:


mettascope/src/worldmap.ts
```diff
@@ -488,74 +496,73 @@ function drawThoughtBubbles() {
...
-        keyActionStep = actionStep
-        break
+      const actionName = state.replay.action_names[action[0]]
+      if (actionName == 'noop' || actionName == 'rotate' || actionName == 'move') {
+        continue
       }
+      keyAction = action
+      keyActionStep = actionStep
+
+      actionIsTarget = actionName == 'swap'
```
lcodes, line 518: only part I'm unsure of; do we select only the actions which have a target, or filter out the ones which do not have one?

mettascope/src/worldmap.ts
```diff
@@ -476,6 +476,14 @@ function drawTrajectory() {
   }
 }
 
+function scaleTileX(x: number) {
```
treeform, line 479: Could you do a simpler thing here and just do that inline. We will fix it by scaling the world soon.

mettascope/src/worldmap.ts
```diff
@@ -488,74 +496,74 @@ function drawThoughtBubbles() {
...
+        if (actionStep !== state.step) { // Avoid drawing the arrow if target is the same as the agent
+          switch (getAttr(state.selectedGridObject, 'agent:orientation', actionStep)) {
+            case 0: targetY -= 1; break // North
+            case 1: targetY += 1; break // South
+            case 2: targetX -= 1; break // West
+            case 3: targetX += 1; break // East
+          }
+          const targetTileX = scaleTileX(targetX)
+          const targetTileY = scaleTileY(targetY)
+          const angle = Math.atan2(targetTileX - tileX, targetTileY - tileY)
```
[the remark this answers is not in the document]
treeform: it works, you are wrong

mettascope/src/worldmap.ts
```diff
@@ -488,74 +496,74 @@ function drawThoughtBubbles() {
...
       const x = getAttr(state.selectedGridObject, 'c')
       const y = getAttr(state.selectedGridObject, 'r')
+      const tileX = scaleTileX(x)
+      const tileY = scaleTileY(y)
+      if (actionHasTarget) {
+        // Draw an arrow on a circle around the target, pointing at it.
+        let targetX = getAttr(state.selectedGridObject, 'c', actionStep)
+        let targetY = getAttr(state.selectedGridObject, 'r', actionStep)
+        if (actionStep !== state.step) { // Avoid drawing the arrow if target is the same as the agent
+          switch (getAttr(state.selectedGridObject, 'agent:orientation', actionStep)) {
```
treeform, line 530: factor this out into its own function next to `function attackGrid(orientation, idx)`

treeform (changes requested):

treeform (approved):

PR #1689, opened 2025-07-25
[MettaScope] fix speed keybinds

monofuel:
- fix . and , to use the Common.SPEEDS speed array and update the UI, rather than directly manipulate the speed state value.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210892502576639)

### DIFF
--- a/mettascope/src/main.ts
+++ b/mettascope/src/main.ts
@@ -1017,6 +1017,7 @@ window.addEventListener('load', async () => {
   ui.dpr = ctx.dpr
 
   await parseUrlParams()
+  setPlaybackSpeed(0.1)
 
   requestFrame()
 })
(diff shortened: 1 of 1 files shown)


mettascope/src/main.ts
```diff
@@ -733,6 +755,7 @@ export function setIsPlaying(isPlaying: boolean) {
 /** Sets the playback speed and updates the speed buttons. */
 function setPlaybackSpeed(speed: number) {
   state.playbackSpeed = speed
+  localStorage.setItem('playbackSpeed', state.playbackSpeed.toString())
```
monofuel, line 758: whoops, didn't mean to add this if it didn't exist already

PR #1693, opened 2025-07-25
Categorical sampling for bucketed curriculum

daphnedemekas:
Changing from values: to just specifying a list of parameters. We make a bucket per item in the list. If the item is a list, then we do categorical sampling. 

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210893953077541)


daphnedemekas:
@bluewin4 

This is ready to be used! But I think @sasmith has some hesitation on the use case. 
I think we'll hold off on merging until there is an experiment that leverages it, so please let us know when that is the case and we can combine it with these changes :)

sasmith (approved):

PR #364, opened 2025-05-04
Replace setup_sso.py with aws sso login command

daveey:
### TL;DR

Replace custom SSO setup script with direct AWS CLI command for authentication.

### What changed?

- Updated the authentication method from a custom Python script (`setup_sso.py`) to the standard AWS CLI command `aws sso login --profile softmax
- Modified the setup instructions in the README to use the AWS CLI directly and copy a config file
- Updated error messages and instructions in the codebase to reference the new authentication method

### How to test?

1. Ensure AWS CLI is installed
2. Run `aws sso login --profile softmax`
3. Verify you can access AWS resources with the authenticated profile
4. Check that the error messages correctly guide users when credentials are missing

### Why make this change?

Simplifies the authentication workflow by leveraging the built-in AWS CLI functionality instead of maintaining a custom script. This provides a more standard approach to AWS authentication that follows AWS best practices and reduces maintenance overhead.

PR #1699, opened 2025-07-25
Make our check for if awscli is installed more general

nishu-builder:


berekuk (approved):
I'm confused how this could help. Even if there could be some other configuration source for awscli, `setup_aws_profiles.sh` writes directly to `~/.aws/config`, so it must be there.

But no harm in trying, so approving.

berekuk:
> david did not initially install via `setup_aws_profiles.sh`

I'm pretty sure David wrote that script... #364

PR #1715, opened 2025-07-26
feat: Cleanup legacy mettagrid_env.py

teodorionita:
# feat: Cleanup legacy `mettagrid_env.py` and fix adapter compatibility

## Summary

Refactored MettaGrid environment class hierarchy. Created proper separation of concerns between core functionality and training features, while fixing framework adapter compatibility issues and making C++ the source of truth for data types.

### New Class Architecture
Created a clean inheritance hierarchy with proper separation of concerns:

```
MettaGridCore (base functionality, no training features)
    ↓
MettaGridPufferBase (+ PufferLib integration)
    ↓
MettaGridEnv (+ training features)

Framework Adapters → MettaGridCore (clean separation)
```

### Core Changes
- **Created MettaGridCore**: Base class with pure C++ wrapper functionality, no training features
- **Created MettaGridPufferBase**: Shared PufferLib integration to avoid duplicate imports
- **Refactored MettaGridEnv**: Now inherits from MettaGridPufferBase for training functionality
- **Made C++ source of truth**: Export data types from `mettagrid_c.cpp` instead of duplicating in Python
- **Added PufferLib compatibility comment** to `types.hpp` per PR feedback

### Framework Adapter Fixes
- **Gym, PettingZoo, PufferLib adapters**: Now use MettaGridCore only (no training features)
- **Fixed inheritance conflicts**: Proper multiple inheritance handling for framework compatibility
- **Added property functions**: Converted repeated None checks to `@property` methods per PR feedback
- **Fixed type assertions**: Added dtype compatibility checks

### Files Added
- `mettagrid/src/metta/mettagrid/puffer_base.py` - MettaGridPufferBase integration

### Files Modified
- `mettagrid/src/metta/mettagrid/core.py` - MettaGridCore base class
- `mettagrid/src/metta/mettagrid/mettagrid_env.py` - Simplified, now inherits from MettaGridPufferBase
- `mettagrid/src/metta/mettagrid/gym_env.py` - Uses MettaGridCore base
- `mettagrid/src/metta/mettagrid/pettingzoo_env.py` - Uses MettaGridCore base
- `mettagrid/src/metta/mettagrid/puffer_env.py` - Uses MettaGridCore base
- `mettagrid/src/metta/mettagrid/mettagrid_c.cpp` - Exports data types
- `mettagrid/src/metta/mettagrid/types.hpp` - Added PufferLib compatibility comment

### Files Removed
- `mettagrid/src/metta/mettagrid/base_env.py` - Consolidated into new architecture

## Benefits

✅ **Proper separation of concerns**: Core functionality vs training features
✅ **No duplicate PufferLib imports**: Single MettaGridPufferBase class
✅ **Framework compatibility**: Clean adapters without training dependencies
✅ **C++ as source of truth**: Eliminates Python type duplication
✅ **Maintainable inheritance**: Clear hierarchy without multiple inheritance conflicts
✅ **All CI tests passing**: Fixed compatibility issues across all test suites

## Addresses PR Feedback

- [x] Import data types from C++ instead of duplicating them in Python
- [x] Move PufferLib compatibility comment to types.hpp
- [x] Fix class hierarchy so MettaGridEnv doesn't directly inherit from both PufferEnv and GymEnv
- [x] Make self._c_env a property function
- [x] Add type assertions for dtype compatibility
- [x] Push Curriculum up a level
- [x] Create MettaGridPufferBase to avoid importing PufferLib in two places
- [x] Keep test imports using direct module imports to reduce PR noise

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210895351420089)


mettagrid/src/metta/mettagrid/base_env.py
```diff
@@ -2,7 +2,7 @@
 MettaGridEnv - Base Python environment class.
 
 This class provides the common functionality for all framework-specific adapters:
```
daveey, line 4: Seems cleaner to put this into mettagrid_env.py:MettaGridEnv

And call the cpp class MettaGridCpp

mettagrid/src/metta/mettagrid/base_env.py
```diff
@@ -210,15 +222,44 @@ def reset_base(self, seed: Optional[int] = None) -> Tuple[np.ndarray, Dict[str,
...
-        buffer allocation and setup specific to their framework.
+        Allocates buffers and sets them in the core environment,
+        then returns the initial observations.
         """
-        pass
+        if self._core_env is None:
+            raise RuntimeError("Core environment not initialized")
+
+        # Allocate buffers
+        self._allocate_buffers()
```
daveey, line 250: We should make sure that buffers are only allocated once, not per-reset. The trainer will allocate buffers per-env, and we then continue to re-use them.

mettagrid/src/metta/mettagrid/base_env.py
```diff
@@ -243,17 +284,14 @@ def step_base(self, actions: np.ndarray) -> Dict[str, Any]:
         # Record step for replay
         if self._replay_writer and self._episode_id:
             with self.timer("_replay_writer.log_step"):
-                rewards = self._core_env._reward_buffer
-                if rewards is not None:
-                    self._replay_writer.log_step(self._episode_id, actions, rewards)
+                if self._rewards is not None:
```
daveey, line 287: Why would they ever be None? We should always log

mettagrid/src/metta/mettagrid/base_env.py
```diff
@@ -243,17 +284,14 @@ def step_base(self, actions: np.ndarray) -> Dict[str, Any]:
...
+                    self._replay_writer.log_step(self._episode_id, actions, self._rewards)
 
         # Check for episode completion
         infos = {}
-        terminals = self._core_env._terminal_buffer
-        truncations = self._core_env._truncation_buffer
 
-        if terminals is not None and truncations is not None:
-            if terminals.all() or truncations.all():
+        if self._terminals is not None and self._truncations is not None:
```
daveey, line 293: Why would the be none

daveey (changes requested):
Let's make sure we have a test that:

1. Creates an MettaGridEnv
2. Sets buffers on it
3. Resets it
4. Makes sure it has a new cpp env, but uses the same buffers

mettagrid/src/metta/mettagrid/base_env.py
teodorionita: fixed now

mettagrid/src/metta/mettagrid/base_env.py
teodorionita: fixed now!

mettagrid/src/metta/mettagrid/base_env.py
teodorionita: renamed base_env to mettagrid_env and also the cpp class to MettaGridCpp

mettagrid/src/metta/mettagrid/__init__.py
```diff
@@ -2,50 +2,49 @@
...
+
+# Import all environment adapters
+from metta.mettagrid.mettagrid_env import MettaGridEnv as MettaGridBaseEnv
 from metta.mettagrid.pettingzoo_env import MettaGridPettingZooEnv
 from metta.mettagrid.puffer_env import MettaGridPufferEnv
 from metta.mettagrid.replay_writer import ReplayWriter
 from metta.mettagrid.stats_writer import StatsWriter
 
+# Additional data types (for backward compatibility)
+dtype_masks = dtype_terminals  # masks same as terminals
```
rjwalters, line 40: the idea here is that the C environment is the source of truth on types - better to import

mettagrid/src/metta/mettagrid/mettagrid_env.py
```diff
@@ -18,40 +26,34 @@
...
 from metta.mettagrid.curriculum.core import Curriculum
 from metta.mettagrid.level_builder import Level
-from metta.mettagrid.mettagrid_c import MettaGrid
+from metta.mettagrid.mettagrid_c import MettaGrid as MettaGridCpp
 from metta.mettagrid.mettagrid_c_config import from_mettagrid_config
 from metta.mettagrid.replay_writer import ReplayWriter
 from metta.mettagrid.stats_writer import StatsWriter
 from metta.mettagrid.util.dict_utils import unroll_nested_dict
 
-# These data types must match PufferLib -- see pufferlib/vector.py
```
rjwalters, line 27: can you make sure this comment is kept somewhere... maybe move to types.hpp

mettagrid/src/metta/mettagrid/mettagrid_env.py
```diff
@@ -18,40 +26,34 @@
...
-dtype_success = np.dtype(bool)
+dtype_actions = np.dtype(np.int32)
 
 logger = logging.getLogger("MettaGridEnv")
 
 
-def required(func):
-    """Marks methods that PufferEnv requires but does not implement for override."""
-    return func
+class MettaGridEnv(PufferEnv, GymEnv):
```
rjwalters, line 45: I thought we were trying to get away from inheriting from puffer and gym?

rjwalters (changes requested):
You might be able to have the "core" python class effectively be `class MettaGrid` as defined in the stubs at `mettagrid/src/metta/mettagrid/mettagrid_c.pyi` ... but the class hierarchy as you have it now seems incorrect... I would guess you want something like:

class MettaGridEnv(MettaGridCore):
class MettaGridPufferEnv(MettaGridCore, PufferEnv):
class MettaGridPettingZooEnv(MettaGridCore, ParallelEnv):

MettaGridCore would manage a MettaGrid c_env and not include any of the Metta training specific code like wandb stats tracking or curriculum

mettagrid/src/metta/mettagrid/mettagrid_env.py
teodorionita: moved it to types.hpp

mettagrid/src/metta/mettagrid/__init__.py
teodorionita: I iexported/mported from cpp directly

teodorionita:
> You might be able to have the "core" python class effectively be `class MettaGrid` as defined in the stubs at `mettagrid/src/metta/mettagrid/mettagrid_c.pyi` ... but the class hierarchy as you have it now seems incorrect... I would guess you want something like:
> 
> class MettaGridEnv(MettaGridCore): class MettaGridPufferEnv(MettaGridCore, PufferEnv): class MettaGridPettingZooEnv(MettaGridCore, ParallelEnv):
> 
> MettaGridCore would manage a MettaGrid c_env and not include any of the Metta training specific code like wandb stats tracking or curriculum

This is the new class inheritance structure and logic (I have kept core.py separate as it makes more sense since the Cpp wrapper will be used across the downstream adapters):

📁 core.py
  ├── MettaGridCore (pure C++ wrapper, no training features)

  📁 mettagrid_env.py  
  ├── MettaGridEnv (MettaGridCore + PufferEnv + training features)
  │   └── Used by our training system (train.py)

  📁 User-facing adapters (inherit from MettaGridCore only):
  ├── puffer_env.py → MettaGridPufferEnv (clean PufferLib adapter)
  ├── gym_env.py → MettaGridGymEnv (Gymnasium adapter)
  └── pettingzoo_env.py → MettaGridPettingZooEnv (PettingZoo adapter)

mettagrid/src/metta/mettagrid/core.py
```diff
@@ -1,239 +1,353 @@
...
+
+    @property
+    def core_env(self) -> Optional[MettaGridCpp]:
+        """Get core environment instance."""
+        return self._c_env
+
+    # Properties that delegate to core environment
     @property
     def max_steps(self) -> int:
+        if self._c_env is None:
```
rjwalters, line 227: I think it might be better to make self._c_env a property function that manages this check rather than repeating it so many times

mettagrid/src/metta/mettagrid/core.py
```diff
@@ -1,239 +1,353 @@
...
     def __init__(
         self,
-        game_config: GameConfig_cpp,
-        map_grid: List[List[str]],
-        seed: int = 0,
-        observation_buffer: Optional[np.ndarray] = None,
-        terminal_buffer: Optional[np.ndarray] = None,
-        truncation_buffer: Optional[np.ndarray] = None,
-        reward_buffer: Optional[np.ndarray] = None,
+        curriculum: Curriculum,
```
rjwalters, line 36: I think we want to push Curriculum up a level - we can instead pass the level in to the base class
`def _create_c_env(self, level: Level, seed: Optional[int] = None) -> MettaGridCpp:`

mettagrid/src/metta/mettagrid/gym_env.py
```diff
@@ -77,59 +64,6 @@ def __init__(
...
-        # Create initial core environment for property access
-        self._core_env = self._create_core_env(0)
-
-        # Buffers for environment data
-        self._observations: Optional[np.ndarray] = None
-        self._terminals: Optional[np.ndarray] = None
-        self._truncations: Optional[np.ndarray] = None
-        self._rewards: Optional[np.ndarray] = None
-
-    def _allocate_buffers(self) -> None:
```
rjwalters, line 89: are you sure we don't need this for Gym? I think the plan is to always use shared memory in the C object -- in the pufferlib vectorization subclasses (puffer and metta) puffer will manage setting up these arrays but I think we need to do it here for gym as you had it before

mettagrid/src/metta/mettagrid/gym_env.py
```diff
@@ -252,34 +178,12 @@ def __init__(
...
-            stats_writer=stats_writer,
-            replay_writer=replay_writer,
-            is_training=is_training,
             single_agent=True,
             **kwargs,
         )
-
-    @property
-    @override
-    def observation_space(self) -> spaces.Box:
```
rjwalters, line 273: I think these are puffer env specific -- but we should check to see if the Gym parent class has any particular functions we have to provide/override

mettagrid/src/metta/mettagrid/mettagrid_c.cpp
```diff
@@ -1059,4 +1059,13 @@ PYBIND11_MODULE(mettagrid_c, m) {
   // This can be fixed, but until we do that, we're not exposing these.
   // .def_readwrite("actions", &GameConfig::actions)
   // .def_readwrite("objects", &GameConfig::objects);
+
+  // Export data types from types.hpp
```
rjwalters, line 1063: this is great - I guess we were not actually checking the types for compatibility at runtime. This way we can add assertions to enforce the type compatibility as required

mettagrid/src/metta/mettagrid/mettagrid_env.py
```diff
@@ -1,280 +1,256 @@
...
-#
-# Important:
-#
-# In PufferLib's class Multiprocessing, the data type for actions will be set to int32
-# whenever the action space is Discrete or Multidiscrete. If we do not match the data type
-# here in our child class, then we will experience extra data conversions in the background.
-# Additionally the actions that are sent to the C environment will be int32 (because PufferEnv
-# controls the type of self.actions) -- creating an opportunity for type confusion.
-
-dtype_observations = np.dtype(np.uint8)
```
rjwalters, line 37: maybe we keep the comment and change these to assertions

mettagrid/src/metta/mettagrid/puffer_env.py
```diff
@@ -12,26 +12,18 @@
...
 from typing_extensions import override
 
-from metta.mettagrid.base_env import MettaGridEnv
+from metta.mettagrid.core import MettaGridCore
 from metta.mettagrid.curriculum.core import Curriculum
 from metta.mettagrid.level_builder import Level
-from metta.mettagrid.replay_writer import ReplayWriter
-from metta.mettagrid.stats_writer import StatsWriter
 
-# Data types must match PufferLib -- see pufferlib/vector.py
```
rjwalters, line 21: we can have the same type assertions as in mettagridenv here

rjwalters (changes requested):
getting there!

mettagrid/tests/renderer/test_nethack.py
```diff
@@ -11,9 +11,9 @@
 import pytest
 from omegaconf import OmegaConf
 
+from metta.mettagrid import MettaGridEnv
 from metta.mettagrid.char_encoder import CHAR_TO_NAME
 from metta.mettagrid.curriculum.core import SingleTaskCurriculum
-from metta.mettagrid.mettagrid_env import MettaGridEnv
```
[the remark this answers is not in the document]
teodorionita: it's just moving from `metta.mettagrid.mettagrid_env` to `metta.mettagrid`, no other changes on any imports or anything else! I would keep it here as it does touch a lot of files but it is essentially an organizational not logic change, but as you wish! will fix the rest of the issues first

mettagrid/src/metta/mettagrid/puffer_env.py
```diff
@@ -12,26 +12,18 @@
...
     PufferLib adapter for MettaGrid environments.
 
-    This class combines the base MettaGridEnv functionality with PufferLib's
-    vectorized environment interface, including proper buffer management.
+    This class provides a clean PufferLib interface for users who want to use
+    MettaGrid environments with their own PufferLib training setup.
+    No training features are included - this is purely for PufferLib compatibility.
     """
 
     def __init__(
```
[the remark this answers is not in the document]
teodorionita: I had a thought of not having two imports from Pufferlib and these are the pros and cons I have sort of come up with: 
Benefits:
  - Remove duplicate PufferEnv inheritance
  - Avoid importing PufferLib in two places
  - DRY principle - single PufferLib integration point
  - Cleaner dependency structure

  Potential concerns:
  - MettaGridEnv (training) would inherit from MettaGridPufferEnv (user adapter)
  - Training environment depending on user adapter (conceptually backwards?)
  - MettaGridEnv adds training features on top of user adapter
  
This is an Opus summary I got after doing a bit of digging: 

```
MettaGridEnv has significant training-specific features:
  - Episode ID generation and tracking
  - Stats writing integration
  - Replay writing integration
  - Curriculum management and episode completion processing
  - Performance timing
  - Episode desyncing for training

  These are definitely training-specific and shouldn't be in the user adapter.

  I think the cleaner approach would be to extract the PufferLib integration into a base 
  class:

  MettaGridCore (framework-agnostic)
      ↓
  MettaGridPufferBase (adds PufferLib compatibility)
      ↓
  ├── MettaGridPufferEnv (user adapter - clean, no training features)
  └── MettaGridEnv (training environment - adds training features)

  This way:
  - Single PufferLib import/integration point
  - Clear separation of concerns
  - Training environment doesn't depend on user adapter
  - Both inherit from the same PufferLib base
```
@rjwalters I think we can do the above and keep the user adapter separate from the puffer training logic that we are using in MettaGridEnv?

mettagrid/src/metta/mettagrid/gym_env.py
teodorionita: checked and solved

mettagrid/src/metta/mettagrid/pettingzoo_env.py
```diff
@@ -52,95 +46,42 @@ def __init__(
...
+        if self._c_env is None:
             raise RuntimeError("Core environment not initialized")
 
         # Create agent names
-        num_agents = self._core_env.num_agents
+        num_agents = self._c_env.num_agents
         self.possible_agents = [f"agent_{i}" for i in range(num_agents)]
         self.agents = self.possible_agents.copy()
 
-    def _allocate_buffers(self) -> None:
```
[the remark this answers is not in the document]
teodorionita: added in again now

mettagrid/tests/renderer/test_nethack.py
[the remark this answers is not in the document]
teodorionita: fixed now, will have a separate PR to properly import from innit.py for the tests!

teodorionita:
@rjwalters I think I have covered all of the reviews, fixed the inheritance and only kept actual logical changes (no aesthetic/import related changes)!

PR #1740, opened 2025-07-28
fix: mettagrid build output capture in CI

rjwalters:
This PR addresses problems with the MettaGrid C++ build system that were preventing error detection in CI. The changes fix segmentation faults due to global destructors, improve build output capture, and resolve several code quality warnings.

### Problems Addressed
1. **Build output not being captured**: The CI system was failing to capture compiler warnings and errors, making it impossible to track code quality
2. **Runtime segfaults**: AddressSanitizer was detecting memory issues that were causing test failures
3. **Global destructor issues**: Static initialization order problems were causing undefined behavior
4. **Missing error detection**: The build checker wasn't properly identifying runtime failures

#### CI/Build System Improvements
- Enhanced `check_mettagrid_build.py` to properly capture and parse build output
- Added `VERBOSE=1` flag to ensure compiler messages are captured
- Implemented detection for runtime issues (segfaults, test failures)
- Added full build output to GitHub Actions for debugging
- Separated test and benchmark configurations for cleaner builds

#### Code Quality Fixes
- Replaced static global maps with function-local statics to avoid destructor ordering issues
- Fixed uninitialized member variables in multiple classes
- Added proper member initialization lists
- Removed unnecessary static members that could cause linking issues

#### Testing Infrastructure
- Split `BUILD_TESTS` and `BUILD_BENCHMARKS` options for better control
- Disabled sanitizers for benchmarks to avoid performance impact
- Fixed parameterized test issues that were causing failures
- Added warning suppressions for expected test framework warnings

### Build Configuration Changes
- Reorganized CMake presets for clarity:
  - `debug`: Unit tests with coverage (no benchmarks)
  - `release`: Production build (no tests)
  - `benchmark`: Release build with benchmarks only
- Updated Makefile targets to match new configurations

### Impact
- CI now properly detects and reports both compile-time and runtime issues
- Developers get clear visibility into build health
- Memory safety issues are caught early through proper sanitizer integration
- Benchmarks run without sanitizer overhead for accurate performance measurements

### Testing
- All unit tests pass with AddressSanitizer enabled
- Benchmarks run successfully without sanitizers
- CI workflow correctly captures and reports all warnings and errors

https://app.asana.com/1/1209016784099267/project/1209041170403492/task/1210889942431074?focus=true

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210909300728501)


nishu-builder (approved):

mettagrid/src/metta/mettagrid/action_handler.hpp
```diff
@@ -115,13 +115,14 @@ class ActionHandler {
 
   // Get the last action name for an agent across all handlers
   static std::string get_last_action_name(size_t agent_id) {
-    auto it = _global_last_actions.find(agent_id);
-    return it != _global_last_actions.end() ? it->second : "";
+    auto& actions = get_global_last_actions();
```
sasmith, line 118: Let's move this to be a property on the agent instead. We don't need the map to be on action handlers at all. Prefer for that to be in a separate PR.

mettagrid/src/metta/mettagrid/grid.hpp
```diff
@@ -137,8 +141,8 @@ class Grid {
     const int r = static_cast<int>(loc.r);
     const int c = static_cast<int>(loc.c);
 
-    int new_r;
-    int new_c;
+    int new_r = loc.r;
```
sasmith, line 144: No. We never want `new_r` to be `loc.r`. It should always be set by the switch statement below. Setting it here is wrong. Instead change the assert on line 165 to throw an exception, versus being an assert that might get removed.

mettagrid/src/metta/mettagrid/grid_object.hpp
```diff
@@ -60,6 +60,8 @@ class GridObject {
   TypeId type_id{};
   std::string type_name;
 
+  GridObject() : id(0), location(), type_id(0), type_name() {}
```
sasmith, line 63: Is there anywhere that we would want to use the default constructor?

mettagrid/src/metta/mettagrid/mettagrid_c.cpp
```diff
@@ -828,7 +825,7 @@ unsigned int StatsTracker::get_current_step() const {
 }
 
 const std::string& StatsTracker::inventory_item_name(InventoryItem item) const {
-  if (!_env) return StatsTracker::NO_ENV_INVENTORY_ITEM_NAME;
+  if (!_env) return get_no_env_inventory_item_name();
```
sasmith, line 828: why?

mettagrid/src/metta/mettagrid/mettagrid_env.py
```diff
@@ -429,7 +429,9 @@ def done(self):
 
     @property
```
sasmith, line 430: It was bad of me to make this a property.

Also, I think we can probably kill usage of feature_normalizations; replacing with `spec` as needed. Maybe just leave a note for now?

mettagrid/src/metta/mettagrid/objects/constants.hpp
```diff
@@ -62,42 +62,54 @@ constexpr ObservationType ObservationFeatureCount = 14;
...
-    {ObservationFeature::Color, "agent:color"},
-    {ObservationFeature::ConvertingOrCoolingDown, "converting"},
-    {ObservationFeature::Swappable, "swappable"},
-    {ObservationFeature::EpisodeCompletionPct, "episode_completion_pct"},
-    {ObservationFeature::LastAction, "last_action"},
-    {ObservationFeature::LastActionArg, "last_action_arg"},
-    {ObservationFeature::LastReward, "last_reward"},
-    {ObservationFeature::Glyph, "agent:glyph"},
-    {ObservationFeature::ResourceRewards, "resource_rewards"}};
+// Use functions to avoid global constructors/destructors
```
sasmith, line 65: You still have global constructors / destructors, as I understand it. In particular, you still have to clean up the local static const during the global destructor phase. Can you say more about what this is protecting against?

(I'm learning about this from Claude.)

mettagrid/src/metta/mettagrid/objects/converter.hpp
```diff
@@ -190,10 +190,6 @@ class Converter : public HasInventory {
     } else if (this->cooldown == 0) {
       // No cooldown, try to start converting again immediately
       this->maybe_start_converting();
-    } else {
```
sasmith, line 193: Was this removal a mistake?

mettagrid/src/metta/mettagrid/packed_coordinate.hpp
```diff
@@ -83,7 +83,7 @@ struct ObservationPattern {
     bool done = false;
 
     Iterator(int h, int w, bool is_end = false)
-        : max_emitted(h * w), row_min(-h / 2), row_max(h / 2), col_min(-w / 2), col_max(w / 2) {
+        : max_emitted(h * w), row_min(-h / 2), row_max(h / 2), col_min(-w / 2), col_max(w / 2), current{0, 0} {
```
sasmith, line 86: I don't know what this does.

sasmith (changes requested):

mettagrid/src/metta/mettagrid/grid.hpp
rjwalters: I get annoying compiler warnings about these being possibly used without being initialized. I'll see if there is a different fix

mettagrid/src/metta/mettagrid/objects/converter.hpp
rjwalters: no, cooling is an unsigned int

mettagrid/src/metta/mettagrid/packed_coordinate.hpp
rjwalters: Adding current{0, 0} to the initializer list explicitly initializes the current member variable (`std::pair<int, int> current;`) with the values {0, 0}.

By explicitly adding current{0, 0}, the code:

- Makes the initialization explicit and clear
- Ensures current starts at {0, 0} regardless of its type
- Follows best practices by initializing all member variables in the constructor
- Prevents potential undefined behavior from uninitialized values

This is likely either fixing a bug (if current was previously uninitialized) or improving code clarity by making the initialization explicit.

-- I could also try to turn down compiler warnings about completely initializing things

mettagrid/src/metta/mettagrid/grid_object.hpp
rjwalters: rm'd

mettagrid/src/metta/mettagrid/grid.hpp
rjwalters: I think the issue is that the assertion can be compiled out depending on optimizations

mettagrid/src/metta/mettagrid/objects/converter.hpp
sasmith: Huh. Good removal then.

sasmith (approved):

PR #1749, opened 2025-07-28
get claude to write your sql for you

nishu-builder:
- add new AIQueryBuilder component with natural language to SQL conversion
- add /sql/generate-query endpoint in backend
- store Anthropic API key in localStorage with clear option
- update SQLQuery component to include AI query builder above query field

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210910729590409)

### DIFF
--- a/app_backend/src/metta/app_backend/routes/sql_routes.py
+++ b/app_backend/src/metta/app_backend/routes/sql_routes.py
@@ -3,6 +3,7 @@
 import asyncio
 from typing import Any, Dict, List
 
+import httpx
 from fastapi import APIRouter, Depends, HTTPException
 from psycopg import errors as pg_errors
 from pydantic import BaseModel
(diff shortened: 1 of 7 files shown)


app_backend/src/metta/app_backend/routes/sql_routes.py
```diff
@@ -34,6 +35,15 @@ class TableSchema(BaseModel):
     columns: List[Dict[str, Any]]
 
 
+class AIQueryRequest(BaseModel):
```
Barraketh, line 38: Should we really force the user to provide the api key? We can just provide the api key to observatory server...

nishu-builder: done. updated observatory-backend/values.yaml file to reflect that it expects a kubernetes secret anthropic-api-key, which we'd have to set. lmk if you prefer we supply it a different way (like through the secrets.tf file)

Barraketh (approved):

PR #1760, opened 2025-07-28
[MettaScope] fix pausing in demo mode, and fix hover bubbles

monofuel:
- adjusted demo mode to only set playing once, not every frame. it feels bad to not let the user pause while demo mode is running. this fix will make time controls work properly on demo mode.
- fixed issues around hover bubbles not going away when they should, also made the code a bit easier to read.
- I'm also sneaking in a flake change to add pnpm to the flake file

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210911412148749)

### DIFF
--- a/mettascope/src/main.ts
+++ b/mettascope/src/main.ts
@@ -252,7 +252,11 @@ onEvent('pointermove', 'body', (_target: HTMLElement, e: Event) => {
     onTraceMinimapChange(event)
   }
 
-  if (!ui.mouseTargets.includes('#worldmap-panel') && !ui.mouseTargets.includes('.hover-panel')) {
+  // Start hide timer when: completely off worldmap/hover-panel OR on worldmap but no object being hovered
+  const shouldHide = (!ui.mouseTargets.includes('#worldmap-panel') && !ui.mouseTargets.includes('.hover-panel')) ||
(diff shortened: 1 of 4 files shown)


mettascope/src/worldmap.ts
```diff
@@ -838,6 +838,10 @@ export function drawMap(panel: PanelInfo) {
             updateHoverBubble(ui.delayedHoverObject)
           }
         }, Common.INFO_PANEL_POP_TIME)
+      } else if (!objectUnderMouse && ui.hoverObject !== null) {
+        // Reset hover state when moving to empty space
```
treeform, line 842: minor nit, complete sentence.

mettascope/src/main.ts
```diff
@@ -252,7 +252,14 @@ onEvent('pointermove', 'body', (_target: HTMLElement, e: Event) => {
     onTraceMinimapChange(event)
   }
 
-  if (!ui.mouseTargets.includes('#worldmap-panel') && !ui.mouseTargets.includes('.hover-panel')) {
+  const isOverBubble = ui.mouseTargets.includes('.hover-panel')
```
treeform, line 255: I like this

mettascope/src/worldmap.ts
monofuel, line 842:
```suggestion
        // Reset hover state when moving to empty space.
```

PR #1790, opened 2025-07-29
configs/user/zfogg.yaml (i made a user config 🙈💁🏾)

zfogg:


relh:
We need more emojis in our PR names

PR #1808, opened 2025-07-30
Fix skypilot setup

berekuk:
`setup_script_location` is running with capture, so we don't use it anymore.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210923148257803)


zfogg (approved):
looks good to me. approved

PR #1824, opened 2025-07-30
comparable-ish SPS on metta training and puffer train

relh:
image


relh:
image

Pretty similar to PufferLib, gets slower over time

PR #1825, opened 2025-07-30
Reduce usage of SamplingCurriculum

sasmith:
Our resolvers should be deterministic at this point, so we shouldn't need a SamplingCurriculum vs a SingleTaskCurriculum. So let's get rid of the former!

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210931065762687)

### DIFF
--- a/mettagrid/src/metta/mettagrid/curriculum/util.py
+++ b/mettagrid/src/metta/mettagrid/curriculum/util.py
@@ -1,19 +1,21 @@
 import hydra
 from omegaconf import DictConfig, OmegaConf
 
-from metta.mettagrid.curriculum.core import Curriculum
-from metta.mettagrid.curriculum.sampling import SamplingCurriculum
+from metta.mettagrid.curriculum.core import Curriculum, SingleTaskCurriculum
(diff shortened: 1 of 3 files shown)


mettagrid/src/metta/mettagrid/curriculum/util.py
```diff
@@ -1,19 +1,21 @@
...
+    loaded_config = hydra.utils.instantiate(
+        # Don't recurse here. We want one level of instantiation so we get the curriculum object, but we don't
+        # want to instantiate sub-curricula or map builders, since we may still want to override their
+        # config.
+        config_from_path(config_path, OmegaConf.create({"env_overrides": env_overrides})),
+        _recursive_=False,
+    )
+    if isinstance(loaded_config, Curriculum):
+        return loaded_config
+    return SingleTaskCurriculum(config_path, loaded_config)
```
sasmith, line 21: Here we're using `config_path` as the task_id. This seems good, but it'll be a much longer string than things like "sample".

jacklionheart (approved):

PR #1826, opened 2025-07-30
Fix external example policy

harshbhatt7585:
This PR fixes the external example policy. It ensures that the return types and values are aligned with those in `MettaAgent` to run the pipeline.


metta/rl/puffer_policy.py
```diff
@@ -67,10 +68,6 @@ def load_pytorch_policy(path: str, device: str = "cpu", pytorch_cfg: DictConfig
         # Use the common instantiate utility
         policy = instantiate(pytorch_cfg, env=env, policy=None)
 
-    policy.load_state_dict(weights)
```
relh, line 70: This function used to be for loading weights, but it seems like that's not happening. Do you think the `instantiate` function from Hydra is doing this? Or did we maybe delete the weight loading

agent/src/metta/agent/external/example.py
```diff
@@ -1,20 +1,159 @@
...
+        if lstm_h is not None and lstm_c is not None:
+            lstm_h = lstm_h.to(self.device)
+            lstm_c = lstm_c.to(self.device)
+            # Ensure LSTM state shapes match
+            expected_num_layers = self.lstm.num_layers
+            if lstm_h.shape[0] != expected_num_layers:
+                if lstm_h.shape[0] > expected_num_layers:
+                    lstm_h = lstm_h[:expected_num_layers, :, :]
+                    lstm_c = lstm_c[:expected_num_layers, :, :]
+                else:
```
relh, line 81: When would this else be true? Since we're producers and consumers of our own LSTM state isn't this else unnecessary?

agent/src/metta/agent/external/example.py
```diff
@@ -1,20 +1,159 @@
...
+
 
 class Recurrent(pufferlib.models.LSTMWrapper):
-    def __init__(self, env, policy, input_size=512, hidden_size=512):
+    def __init__(self, env, policy=None, cnn_channels=128, input_size=512, hidden_size=512):
+        if policy is None:
+            policy = Policy(env, cnn_channels=cnn_channels, hidden_size=hidden_size, input_size=input_size)
         super().__init__(env, policy, input_size, hidden_size)
 
+    def activate_actions(self, action_names, action_max_params, device):
```
relh, line 19: By putting this in the external/example.py, we're basically forcing policies that want to use Metta to also have an activate actions function. However, maybe we can put this in the PyTorchAgent and/or MettaAgent that wraps external policies

metta/rl/util/policy_management.py
```diff
@@ -123,6 +123,8 @@ def validate_policy_environment_match(policy: Any, env: Any) -> None:
         agent = policy
     elif isinstance(policy, DistributedMettaAgent):
         agent = policy.module
+    elif type(policy).__name__ == "Recurrent":
```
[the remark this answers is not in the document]
harshbhatt7585: For now it was to make it run in short term. 
More ideal approach could be through the following ways:
1. There is an abstraction class which creates MettaAgent and External Policy and instance will get bypass if it is built from Abstraction.
2. MettaAgent is supreme class which wraps external policies too.
3. External Policy is built through inheriting MettaAgent.

still thinking what could be the most ideal approach. wdut?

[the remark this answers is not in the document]
harshbhatt7585: what I feel, there should be an Abstraction layer to build MettaAgent Policies. 
It should assign what method should be implemented and what values to be returned.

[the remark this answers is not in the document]
relh: I think the best approach is that we always have a MettaAgent. We could have a MettaAgentBuilder that either did the YAML/component way or did just a normal torch.nn.Module.

I was originally thinking of having a BrainPolicy and a PytorchPolicy be the two names for the ways of building things, but I think that's a bit complicated. 

Anyways, I think that we can pull out the "building" part of MettaAgent such that the real MettaAgent doesn't need to know how it was built. 

Basically, the building part of the components could be it's own thing. Let me know what you think. I think this get's complicated because MettaAgent has a bunch of functions/members related to building things with YAML but I'm not sure which of these matter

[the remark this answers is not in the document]
harshbhatt7585: I've prototype the idea of this in `external/new_agent.py` (checkout new commit). Let me know what do you think of that,

agent/src/metta/agent/external/example.py
harshbhatt7585: there was some mismatch happpening, now I simplified a little also.

agent/src/metta/agent/external/new_agent.py
```diff
@@ -0,0 +1,652 @@
...
+
+    def _instantiate_policy(self, policy_config: Union[DictConfig, Dict]) -> PolicyBase:
+        """Instantiate policy from configuration."""
+        if isinstance(policy_config, (DictConfig, dict)) and "_target_" in policy_config:
+            return instantiate(policy_config)
+        else:
+            raise TypeError("Policy config must have '_target_' field")
+
+
+class ComponentBasedPolicy(PolicyBase):
```
relh, line 281: We can just call this ComponentPolicy

agent/src/metta/agent/external/new_agent.py
```diff
@@ -0,0 +1,652 @@
...
+            log_prob = log_prob.sum(dim=-1)
+
+        entropy = dist.entropy()
+        if len(entropy.shape) > 1:  # Multi-dimensional continuous actions
+            entropy = entropy.sum(dim=-1)
+
+        return action, log_prob, entropy, value, new_state
+
+
+class MonolithicPolicy(PolicyBase):
```
relh, line 327: we can just call this PytorchPolicy but we shouldn't even need this to be a class. This can also come later

agent/src/metta/agent/external/new_agent.py
```diff
@@ -0,0 +1,652 @@
...
+            log_prob = log_prob.sum(dim=-1)
+
+        entropy = dist.entropy()
+        if len(entropy.shape) > 1:  # Multi-dimensional continuous actions
+            entropy = entropy.sum(dim=-1)
+
+        return action, log_prob, entropy, value, new_state
+
+
+class PufferlibRecurrentPolicy(PolicyBase):
```
relh, line 437: Does this have to be different from Monolithic policy? PufferLib uses normal Pytorch nn.modules just like the Monolithic/PytorchPolicy does

relh (changes requested):

PR #1828, opened 2025-07-30
fix: clean up stationary tracking

rjwalters:
This PR simplifies how we track whether agents are stationary:

- Removes the `ActionTrackingState` struct and associated tracking maps from ActionHandler
- Adds simple `prev_location` and `prev_action` fields directly to the Agent class
- Implements stationary status detection by comparing current and previous locations in handle_action()

The new approach is cleaner, more maintainable, and achieves the same functionality with less code complexity.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210931838115612)


bluewin4:
It looks like there is a bit of a problem in the test.

Aside from that we should have a quick chat about intent for this system and future extensions.

bluewin4 (approved):

PR #1834, opened 2025-07-30
Set updated_at on eval tasks

Barraketh:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210932793690504)


nishu-builder (approved):
lol this was the only one for which updated_at mattered and I missed it :P thanks for catching

PR #1835, opened 2025-07-30
[MettaScope] Trace cache fixes

monofuel:
- if the trace panel is resized, we need to re-draw it.
- If the trace panel is closed and opened again, we should track the state.

### DIFF
--- a/mettascope/src/traces.ts
+++ b/mettascope/src/traces.ts
@@ -12,6 +12,13 @@ const lastCachedState = {
   selection: null as any,
   zoomLevel: -1,
   panPos: null as Vec2f | null,
+  width: -1,
+  height: -1,
+}
+
+/** Forces the trace to redraw on next call to drawTrace. */
+export function invalidateTrace() {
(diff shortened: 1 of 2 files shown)


mettascope/src/traces.ts
```diff
@@ -12,6 +12,13 @@ const lastCachedState = {
   selection: null as any,
   zoomLevel: -1,
   panPos: null as Vec2f | null,
+  width: -1,
+  height: -1,
+}
+
+/** Invalidate the trace cache to force regeneration on next call to drawTrace. */
+export function invalidateTrace() {
```
monofuel, line 20: This doesn't feel great. but drawTrace() is only ran when the panel is visible, and drawing the panel is toggled with a boolean. if you open and close the panel, the panel thinks it hasn't changed from the last time it was opened.

treeform: have you tried just adding the open/closed boolean to the cache key? Maybe that is cleaner?

PR #1839, opened 2025-07-30
gridworks: correctly classify curriculums vs maps

sandover:
previously gridworks would try to open & render curriculum files (& show errors).  Fixed this.


berekuk (changes requested):
The improvements look good, but this breaks curriculum tests.
Try changing `test_validate_all_curriculums` code to call `curriculum_from_config_path`?

PR #1841, opened 2025-07-31
Move some metta.common things to correct namespace package

nishu-builder:
I accidentally put some things into metta/common/ instead of common/src/metta/common. this puts them in the right place

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210933736787502)


relh:
> I accidentally put some things into metta/common/ instead of common/src/metta/common. this puts them in the right place
> 
> [Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210933736787502)

Are you sure this is the right place? Almost seems like you might have preferred a shorter path

PR #1859, opened 2025-07-31
in-line rollout

relh:
I think this is cleaner because it reduces layers of indirection and helps people *understand whats happening in the rollout loop*

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210942306245700)


metta/rl/trainer.py
```diff
@@ -286,14 +286,43 @@ def train(
...
+                        dones=d,
+                        truncations=t,
+                        values=values,
+                        env_id=training_env_id,
+                        mask=mask,
+                        lstm_state=lstm_state_to_store,
+                    )
+
+                    # Send actions back to environment
+                    with timer("_rollout.env"):
```
nishu-builder, line 318: we talked about putting just this one part back in the other file

relh: ADdresseded!

nishu-builder (approved):

PR #1864, opened 2025-07-31
[MettaScope] Heatmaps

monofuel:
- add a heatmap to the world grid and the map view

Screenshot From 2025-07-31 16-43-35

Screenshot From 2025-07-31 16-43-40

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210942945213744)


mettascope/src/common.ts
```diff
@@ -116,6 +121,7 @@ export const state = {
   showGrid: true,
   showVisualRanges: true,
   showFogOfWar: false,
+  showHeatmap: true, // TODO we need to add a heatmap button
```
monofuel, line 124: TODO we need to do this, but this will involve figma work

monofuel, line 124:
```suggestion
  showHeatmap: false, // TODO we need to add a heatmap button
```

treeform (approved):

PR #1882, opened 2025-08-01
Create/load policy only on master and then distribute using NCCL

dominik3141:
This is a new version of the reverted PR #1614.

**Summary**
Make sure only the master rank loads/creates a new policy as DDP syncs the master policy to all other ranks at initialization time.

**Context**
From the [PyTorch docs](https://docs.pytorch.org/docs/stable/notes/ddp.html):
> Construction: The DDP constructor takes a reference to the local module, and broadcasts state_dict() from the process with rank 0 to all other processes in the group to make sure that all model replicas start from the exact same state.


dominik3141:
> My only worry, and the reason this is still a draft, is that we no longer provide non-master ranks with `initial_policy_record` and `latest_saved_policy_record`. This should be fine for almost everything, but it might be a problem for kickstarting.

@relh what do you think?

dominik3141:
Update regarding kickstarting: I talked to David yesterday and he wants kickstarting out of trainer so it can just be a loss as soon as losses are refactored by @Al-does.

Al-does:
Absolutely. I can make that happen in a few days. If it's in the way of anything then feel free to delete it.

PR #1822, opened 2025-07-30
feat: add common setup script for skypilot yaml config

rjwalters:
* **Centralized env setup** in `setup-common.sh`: writes Metta environment activation and all Skypilot‐derived exports into `~/.bashrc`.
* **Sandbox config** (`sandbox.yaml`) now runs `setup-common.sh` and the GPU trouble‑shooter during setup.
* **Training workflow** (`sk_train.yaml`) simplified “run” step to `source ~/.bashrc` (drops the six manual `export …` lines).
* **fix: Sandbox Git Ref** exported git ref from cli argument to ENV so it is actually used
* 
**Why**
Keeping all SKYPILOT‑based exports in one place (bashrc via `setup-common.sh`) makes the CI/cluster setup DRY and ensures any future env changes only need to live in one script. 

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210929537733468)

PR #1885, opened 2025-08-01
Fix cost monitor and skypilot latency script

dominik3141:
**Summary**
Cost monitor and skypilot latency script both have stopped working a few days ago. This PR fixes both scripts.

**Context**
One problem was #1822 which made the `NUM_NODES` env variable inaccessible to the script, but there must have been another change that I couldn't identify so far that prevented both scripts to be run by `uv`, which is a big problem as both scripts have dependencies that are added in the form of
```
#!/usr/bin/env python3
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "wandb",
# ]
# ///
```
which causes `uv` to then add these to the environment temporarily.

The logs also repeatedly show ```(setup pid=1958) warning: `VIRTUAL_ENV=/puffertank/venv` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead``` which leads me to believe that there still is a problem of some sorts.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210953407091208)

### DIFF
--- a/common/src/metta/common/util/cost_monitor.py
+++ b/common/src/metta/common/util/cost_monitor.py
@@ -131,9 +131,9 @@ def main():
     cost_info = get_cost_info()
     if cost_info:
         instance_hourly_cost = cost_info["instance_hourly_cost"]
-        num_nodes_env = os.environ.get("NUM_NODES")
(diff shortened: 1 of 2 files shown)


mstormbull (approved):
Lgtm

PR #1894, opened 2025-08-02
trainer: if evaluate_remote=True evaluate remote else evaluate local

zfogg:
this is needed so i can better test things as i'm working, and so i don't have to wait for evaluations to run locally as i train.

if you set evaluate_remote to true in the trainer config, local evaluations won't happen. the inverse will happen if you set it to false.

let me know what you think.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211002717239838)


zfogg:
@claude review this PR for me

zfogg:
@claude review this pr again. check if the `cfg=cfg` bug is gone, and the epoch tracking indentation bug

Barraketh (approved):
Explain why we have both 'evaluate_remote' and 'evaluate_local' ?

PR #1900, opened 2025-08-04
Pytorch Agents

harshbhatt7585:
This PR converts existing YAML agents to PyTorch agents. It loads the agent as an external policy and runs it.

To run the external policy, place your target policy configuration in a `{user}.yaml` file like this:
```
pytorch:
 _target_: metta.agent.external.latent_attn_med.Recurrent
 hidden_size: 512
 cnn_channels: 128
```

Make sure your initial policy uri set as `pytorch://checkpoints/metta-new/metta.pt`
like this:
```
trainer:
  initial_policy:
     uri: pytorch://checkpoints/metta-new/metta.pt
```


relh:
image
image
image
image

It will look like the normal MettaAgent is being used from the logs, but the new one is being used. We need to address this moving forward.

PR #1902, opened 2025-08-04
Google Drive upload of filesets

jewstinbeac:
## Summary
Introduce a CLI tool (`tools/gdrive_upload/export_fileset.py`) that exports YAML-defined code “filesets” to Google Drive as Google Docs for Asana AI.

## Highlights
- **CLI options**: `--all`, `--fileset <name>`, `--dry-run`  
- **Config**: `tools/gdrive_upload/filesets.yml` with auto-deduplication  
- **Drive integration**: OAuth-based upload, size limits, binary detection  

## Setup
1. `pip install -r tools/gdrive_upload/requirements.txt`  
2. Place `credentials.json` & `token.json` in `~/.config/metta-gdrive/`  
3. Run `python tools/gdrive_upload/export_fileset.py --all`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,5 @@ sweep_id
 # Notebooks
 experiments/notebooks/.ipynb_checkpoints
 experiments/notebooks/utils/scorecard_widget/scorecard_widget/static/index.js
+tools/gdrive_upload/credentials.json
+tools/gdrive_upload/token.json
--- a/.gitignore
+++ b/.gitignore
@@ -85,5 +85,7 @@ sweep_id
 # Notebooks
 experiments/notebooks/.ipynb_checkpoints
 experiments/notebooks/utils/scorecard_widget/scorecard_widget/static/index.js
+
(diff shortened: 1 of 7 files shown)


jacklionheart (changes requested):
Let's discuss on discord. But the broad scope is:

* I'd like to see you update `file.py` to support google drive and then we use that api rather than having everything be a separate file 
* we should be using `codeclip` to generate the actual text to submit, which will handle things like ignoring for you
* I'd like you to produce a basic report (dont need to check this in, but just something you share and we discuss) on how google drive and asana handle different sizes of docs
* we shouldnt need any separate requirements here; if there are requirements to add (probably the main google ones) to the main repo requirements that is ok to add there
* generally speaking I think we should be able to have a much much simpler and shorter "instructions" and test script files if we make these changes. LLMs tend to be a little verbose in what they generate and I would rather have code that is so simple it essentially documents itself with ~10-100lines of basic usage demos rather than a small one off script like this requiring its own setup_instructions file

PR #1903, opened 2025-08-04
Add new movement keys to support new movement directions.

treeform:


treeform:
https://github.com/Metta-AI/metta/pull/2073

PR #1917, opened 2025-08-05
Add notebook support to metta CLI with kernel registration and status…

sandover:
- add "notebook" as a component for metta install and metta status
- update README
-

### DIFF
--- a/experiments/notebooks/README.md
+++ b/experiments/notebooks/README.md
@@ -1,17 +1,32 @@
 ## Mettabook - Research Analysis Notebooks
 
-This is experimental. The hopes are that it makes analysis easier - providing a unified interface for launching training runs, monitoring them, and analyzing results all in one place.
+This is experimental. The hopes are that it makes analysis easier - providing a unified interface for launching training
+runs, monitoring them, and analyzing results all in one place.
 
 More tools will be added over time, including integration with Observatory
 
-We expect that this won't be perfect at first but want to gauge its viability. Please post in #researcher-tools for anything that is painful or that you need!
+We expect that this won't be perfect at first but want to gauge its viability. Please post in #researcher-tools for
+anything that is painful or that you need!
 
-### How to Use
+### Setup
 
-- Copy `mettabook-template.ipynb` to your own file in the research/ folder
-- Check you're up to date with `metta install`
-- Open your notebook in Cursor. When you open the file or try to run a cell (shift+enter by default), you should be prompted to select a Python kernel. Select the one from your uv-managed `.venv`
+1. **Install notebook support**:
+
+   ```bash
+   metta install notebook
+   ```
+
+   Then `metta status` to check that it worked.
+
+2. **In Cursor or VS Code, select the correct environment**:
+
+   When you open a notebook file (`.ipynb` file), you'll need to choose a python environment for it. Just pick the same
+   environment you see listed in `metta status`, e.g. ".venv (Python 3.11.7)"
 
 ### Sharing results
-- Commit relatively freely to the `research/` subfolder for now. Don't hold yourself to a high quality bar: these are not intended to be depended on
+
+- Feel free to make copies of an existing notebook such as `mettabook-template.ipynb` into your own file in the
+  `research/` folder
+- Commit relatively freely to the `research/` subfolder for now. Don't hold yourself to a high quality bar: these are
+  not intended to be depended on
 - Add tools you think others will also want to`utils/` files. These will undergo a more typical dev process
--- a/experiments/notebooks/mettabook-template.ipynb
+++ b/experiments/notebooks/mettabook-template.ipynb
@@ -30,7 +30,18 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "ename": "",
+     "evalue": "",
+     "output_type": "error",
+     "traceback": [
+      "\u001b[1;31mRunning cells with 'Python 3.13.5' requires the ipykernel package.\n",
+      "\u001b[1;31mInstall 'ipykernel' into the Python environment. \n",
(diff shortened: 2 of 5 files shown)


metta/setup/components/notebook.py
```diff
@@ -0,0 +1,255 @@
...
+    # Kernel configuration constants
+    KERNEL_NAME = "metta-venv-kernel"
+    DISPLAY_NAME = "metta-venv-kernel"
+
+    @property
+    def description(self) -> str:
+        return "Jupyter notebook support with kernel registration"
+
+    def is_applicable(self) -> bool:
+        # This component is always applicable - all users need notebook support
```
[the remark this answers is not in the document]
nishu-builder: a true external version would be on google colab

metta/setup/components/notebook.py
```diff
@@ -0,0 +1,255 @@
...
+        This method uses 'jupyter kernelspec list --json' to find kernels,
+        which is cross-platform and handles different OS kernel locations automatically.
+
+        Args:
+            kernel_name: Name of the kernel to find (e.g., 'metta-notebook-kernel')
+
+        Returns:
+            Path to kernel directory if found, None otherwise
+        """
+        try:
```
nishu-builder, line 48: we want errors to propagate and not fail silently

metta/setup/components/notebook.py
```diff
@@ -0,0 +1,255 @@
...
+            # Use Jupyter's built-in discovery to find kernels
+            # This works on all platforms (macOS, Linux, Windows)
+            result = self.run_command(["jupyter", "kernelspec", "list", "--json"], check=False)
```
nishu-builder, line 51: if we want `jupyter` to be available, we should add it to Brewfile. alternatively it might just be availble given what we have in pyproject.toml via `uv run jupyter`, in which case we should use that

metta/setup/components/notebook.py
```diff
@@ -0,0 +1,255 @@
...
+                kernels = json.loads(result.stdout)
+                if kernel_name in kernels["kernelspecs"]:
+                    return Path(kernels["kernelspecs"][kernel_name]["resource_dir"])
+        except Exception:
+            # If Jupyter discovery fails, return None
+            # This could happen if jupyter command is not available
+            pass
+        return None
+
+    def _validate_kernel_spec(self, kernel_path: Path) -> bool:
```
nishu-builder, line 62: hmm not sure we need this; i think we just advise peopel do `uv run jpyter`. this is going to have trouble with multiple git checkouts otherwise (like git worktrees)

metta/setup/components/notebook.py
```diff
@@ -0,0 +1,255 @@
...
+        2. Removes any existing kernel with the same name (idempotent)
+        3. Registers a new kernel using ipykernel install
+
+        The kernel is registered as 'metta-venv-kernel' with display name 'metta-venv-kernel',
+        making it available in Jupyter notebooks and labs.
+        """
+        try:
+            # Step 1: Ensure dependencies are synced
+            self.run_command(["uv", "sync"])
+            success("✓ Dependencies synced")
```
nishu-builder, line 149: dont want to add this extra log; i know `metta install` is currently noisy, but we don't want it to be

nishu-builder: well in particular if they're a bit vague. also `uv sync` will already have been run: `metta install`, even `metta install {component}`, always runs the core.py thing first, which does `uv sync`

metta/setup/components/notebook.py
```diff
@@ -0,0 +1,255 @@
...
+        Check if the kernel is properly registered.
+
+        This method uses Jupyter discovery to find our kernel and returns
+        a status string if found. This is used by 'metta status' to show
+        the connection status in the "Connected As" column.
+
+        Returns:
+            ".venv (Python X.X.X)" if kernel is registered, None otherwise
+        """
+        try:
```
nishu-builder, line 246: there shouldn't be room for errors here, so no need for try/catch. uv also already guarantees that your python version is in sync with what's in our pyproject.toml, so i dont think that checking python version here is needed in general. totally fine to look for the existence of the path though

pyproject.toml
```diff
@@ -56,6 +56,7 @@ dependencies = [
   "widgetsnbextension>=4.0.14",
   "anywidget>=0.9.18",
   "traitlets>=5.14.3",
+  "ipykernel>=6.29.5",
```
nishu-builder, line 59: any reason for this? training boxes/etc don't need it

experiments/notebooks/mettabook-template.ipynb
```diff
@@ -30,7 +30,18 @@
...
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "ename": "",
+     "evalue": "",
+     "output_type": "error",
+     "traceback": [
+      "\u001b[1;31mRunning cells with 'Python 3.13.5' requires the ipykernel package.\n",
```
nishu-builder, line 39: let's clear outputs for this before committing

experiments/notebooks/README.md
```diff
@@ -1,17 +1,32 @@
...
-### How to Use
+### Setup
 
-- Copy `mettabook-template.ipynb` to your own file in the research/ folder
-- Check you're up to date with `metta install`
-- Open your notebook in Cursor. When you open the file or try to run a cell (shift+enter by default), you should be prompted to select a Python kernel. Select the one from your uv-managed `.venv`
+1. **Install notebook support**:
+
+   ```bash
+   metta install notebook
```
nishu-builder, line 16: we need to install more than just notebook -- we need the ability to launch skypiilot, hit observatory (for which we need observatory auth), hit wandb (for which we need wandb auth), etc; i would, for now, just keep it where it was before

nishu-builder (changes requested):
left some comments :)

PR #1921, opened 2025-08-05
Remove unused generation stats, simplify load_or_initialize_policy and improve its docstring

nishu-builder:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210971151343703)

### DIFF
--- a/metta/rl/policy_management.py
+++ b/metta/rl/policy_management.py
@@ -100,40 +100,40 @@ def load_or_initialize_policy(
     metta_grid_env: MettaGridEnv,
     is_master: bool,
     rank: int,
-) -> tuple[PolicyAgent, PolicyRecord, PolicyRecord]:
-    """Load or initialize policy with distributed coordination."""
+) -> PolicyRecord:
+    """
+    Load or initialize policy with distributed coordination.
+
+    First, checks if there is an existing policy at any of:
+        - checkpoint.policy_path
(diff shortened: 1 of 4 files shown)


metta/rl/trainer.py
```diff
@@ -399,7 +399,7 @@ def train(
                             if hasattr(policy, "clip_weights"):
                                 policy.clip_weights()
 
-                            if str(device).startswith("cuda"):
+                            if device.type == "cuda":
```
relh, line 402: are we sure this is always a torch device? I remember this being dubious a month or so ago?

metta/rl/policy_management.py
```diff
@@ -156,27 +156,18 @@ def log_progress(elapsed: float, status: str) -> None:
...
-
-        policy = policy_record.policy
-        initial_policy_record = policy_record
-        latest_saved_policy_record = policy_record
-
-        logger.info(f"Rank {rank}: Successfully loaded policy from {default_path}")
     else:
         # Master creates new policy
         logger.info("No existing policy found, creating new one")
-        name = policy_store.make_model_name(0)
```
relh, line 168: the reason we were doing this weird thing is the old `torch.package` code required us to basically re-instantiate the policy entirely and then load the weights in every time.

relh (approved):

PR #1922, opened 2025-08-05
Misc improvements: remove unused advantages param, add torch.device types, improve type hints

nishu-builder:


metta/rl/advantage.py
```diff
@@ -39,7 +36,7 @@ def compute_advantage(
     values, rewards, dones, importance_sampling_ratio, advantages = tensors
 
     # Create context manager that only applies CUDA device context if needed
-    device_context = torch.cuda.device(device) if str(device).startswith("cuda") else nullcontext()
+    device_context = torch.cuda.device(device) if device.type == "cuda" else nullcontext()
```
relh, line 39: ahh i see that this PR solves the device quandry I had

relh (approved):
nice removal of advantages when we have minibatch. I think originally there were some initial advantages being used there but glad to see it all cleaned up

PR #1929, opened 2025-08-05
Imporve Pytorch Agent Initialization

harshbhatt7585:
This PR improves the initialization of pytorch agent by adding another flag `+py_agent='*.py'`.
It can be use like this:
```
 ./tools/train.py run=my_experiment142 py_agent=fast.py
``` 
Atm, there are two flags available for loading agent `agent` and `py_agent`. `agent` loads the path yaml based agent available in `config/agents/`, if `py_agent` is present then it overrides the `agent`.

It also fixes Logging of the correct agent is being loaded.

Limitations:
1. It still logs incorrect agent if loading through external policy (using yaml file +user=), as the identification of loading the external policy done later than logging the hydra config, therefore it logs the default version.


relh (approved):
made some small changes so we don't have to change types places

PR #1650, opened 2025-07-24
mapgen num_agents parameter

berekuk:
mapgen num_agents parameter

implement intrinsic_size in inline_ascii

refactor mapgen tests

test num_agents

comments

PR #1930, opened 2025-08-05
mapgen instance_map param

berekuk:
This PR is for the sake of integrating `TerrainFromNumpy` with MapGen.

While porting it, I realized that `TerrainFromNumpy` can't be represented as a scene, because scenes don't have sizes (there's some rudimentary support for `intrinsic_size` but it's limited and it's not a good fit for `TerrainFromNumpy`).

So I had to do it in a different way: now, when configuring mapgen, you have a choice between the existing syntax:

```
_target_: metta.map.mapgen.MapGen
width: 10
height: 10
instances: 4
root:
  type: metta.map.scenes.my_scene.MyScene
  ...
```

And using `instance_map` instead of `root`, where `instance_map` points to another `LevelBuilder`, that produces its own grid:

```
_target_: metta.map.mapgen.MapGen
instances: 4
instance_map:
  type: metta.map.terrain_from_numpy.TerrainFromNumpy
  ...
```

Similar to [num_agents](https://github.com/Metta-AI/metta/pull/1650) parameter, when using `instance_map`, MapGen instantiates the instance grids separately at first, and then copies them into the final grid.

---

The behavior of borders around instances in this PR is slightly different from what we had before.

Before:

After (all empty space around instances is filled):

I'm 99% sure this won't affect anything.

---

More thoughts on this:

I don't love the amount of complexity that I had to introduce to achieve this, but it's the best tradeoff I could find.

Why bother with this at all? Could I keep `TerrainFromNumpy` separate from mapgen? I wasn't sure at first, but:
- `instances` and `num_agents` in mapgen are still useful, and we want to retire `mettagrid.room` completely as a matter of principle, so it's still a minor win compared to `mettagrid.room.multi_room` that we used before
- I hope to trickle down rng seeds from mapgen to instance maps, so that maps will stay fully deterministic (this PR doesn't implement this yet)
- I hope to transplant the scene tree from `instance_map` to the main grid when the instance map has its own scene (not implemented yet; minor but will make some APIs more consistent)
- I'm considering removing `Scene.intrinsic_size` API, so it's less of "one more extra workaround for mapgen limitations", and more like "replacing one workaround with another that's more flexible"

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210977451949448)

### DIFF
--- a/metta/map/mapgen.py
+++ b/metta/map/mapgen.py
@@ -1,10 +1,12 @@
-from typing import cast
+from typing import Any, cast
 
 import numpy as np
 from omegaconf import DictConfig, OmegaConf
+from pydantic import model_validator
 
 from metta.common.util.config import Config
-from metta.map.scene import Scene, load_class, make_scene, scene_cfg_to_dict
+from metta.map.scene import load_class, make_scene, scene_cfg_to_dict
+from metta.map.scenes.copy_grid import CopyGrid
 from metta.map.scenes.room_grid import RoomGrid, RoomGridParams
 from metta.map.scenes.transplant_scene import TransplantScene
 from metta.map.types import MapGrid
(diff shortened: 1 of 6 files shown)


berekuk:
* **#1931** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1931?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1930** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1930?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1930?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1933, opened 2025-08-05
Port VariedTerrain to mapgen

berekuk:
Migrated the VariedTerrain room implementation to the new map system architecture.

### What changed?

- Moved `VariedTerrain` to `metta/map/scenes/varied_terrain.py`
- Refactored `VariedTerrain` to inherit from `Scene` instead of `Room`
- Updated all configuration files to use the new `MapGen` target instead of `MultiRoom`
- Removed some unused config options (`teams`, `occupancy_threshold`)

---

Note that it's still a monolith. The current layout algorithm in VariedTerrain works by looking for free rectangular zones of different sizes incrementally. It's possible that no one would notice if I did something slightly different that would fit mapgen's scene tree approach better, but I didn't want to take that risk.

The code for placing objects and agents could be easy to port to MapGen's `Random` scene, but... `children` are harder to override in curriculum configs (because OmegaConf can't merge into lists, only into dicts), so I want to figure out the solution for that before I remove this feature from `VariedTerrain`.

We don't currently need to override `VariedTerrain`'s `objects` param in curricula, but I don't want to risk the situation where someone needs it and can't do it at all. I've filed this as a [separate task](https://app.asana.com/1/1209016784099267/project/1210006613945419/task/1210971517432516?focus=true) to figure out.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210978786941023)

### DIFF
--- a/metta/map/scenes/varied_terrain.py
+++ b/metta/map/scenes/varied_terrain.py
@@ -19,12 +19,18 @@ The build order is:
 from typing import List, Optional, Tuple
 
 import numpy as np
-from omegaconf import DictConfig
 
-from metta.mettagrid.room.room import Room
+from metta.common.util.config import Config
+from metta.map.scene import Scene
 
 
-class VariedTerrain(Room):
+class VariedTerrainParams(Config):
+    objects: dict[str, int]
+    agents: int = 1
+    style: str = "balanced"
+
+
+class VariedTerrain(Scene[VariedTerrainParams]):
     # Base style parameters for a 60x60 (area=3600) grid.
     # These counts are intentionally moderate.
     STYLE_PARAMETERS = {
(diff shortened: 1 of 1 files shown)


berekuk:
* **#1933** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1933?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/1933?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sasmith (approved):
Very much just skimmed.

PR #1966, opened 2025-08-06
Reset the single trial task without having to reinitialize it

daphnedemekas:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1210992420834124)

### DIFF
--- a/mettagrid/src/metta/mettagrid/curriculum/core.py
+++ b/mettagrid/src/metta/mettagrid/curriculum/core.py
@@ -124,7 +124,3 @@ class SingleTaskCurriculum(Curriculum):
 
     def get_task_probs(self) -> dict[str, float]:
         return {self._task_id: 1.0}
-
-    def env_cfg(self) -> DictConfig:
-        assert self._task_cfg is not None, "Task has no environment configuration"
-        return self._task_cfg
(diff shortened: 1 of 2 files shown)


mettagrid/src/metta/mettagrid/curriculum/core.py
```diff
@@ -89,6 +89,15 @@ def __init__(self, id: str, curriculum: Curriculum, env_cfg: DictConfig):
         # We may have been lazy about instantiation up to this point, since that allows us to
         # override the config. Now we complete the instantiation.
         self._env_cfg = hydra.utils.instantiate(env_cfg)
+        self._env_cfg.current_trial = self._current_trial
+
+    def reset(self, env_cfg):
```
daveey, line 94: I don't think tasks get reset, we should be getting a new task

daveey (changes requested):
This seems wrong, tasks don't get reset. They get completed.

mettagrid/src/metta/mettagrid/curriculum/core.py
daphnedemekas: Idk let's talk about it. If we reinitialize a task, current_trial and total_reward continuously get reset to zero. For a particular episode, since the environment is the only thing that changes, and we want to be able to track those variables, this made more sense to me.

PR #1983, opened 2025-08-07
[MettaScope] frontend replay format validation

monofuel:
- validate the replay object, and each play step
- manually tested on play mode, a new replay, and an old replay. all seem good.
- clicking on the modal will make it go away

Screenshot From 2025-08-08 14-45-33

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211002621547492)


mettascope/src/replay.ts
```diff
@@ -571,7 +575,6 @@ function loadReplayJson(url: string, replayJson: any) {
     html.fileName.textContent = url.split('/').pop() || 'unknown'
   }
 
-  Common.closeModal()
```
monofuel, line 574: Moved this up to before validating replay data

otherwise it was closing the error modal

PR #1989, opened 2025-08-07
ASCII map -> thumbnail png

lcodes:
Reuses the existing gen_thumb.py, adds an intermediate repr to switch formats


mettascope/tools/gen_thumb.py
```diff
@@ -69,12 +69,77 @@ def gen_thumb(input, step, size, output):
...
+            match chr(input[offset + x]):
+                case "@" | "A" | "1" | "2" | "3" | "4" | "p" | "P":
+                    type_id = 0
+                    agent_id = num_agents
+                    num_agents += 1
+                case "#" | "W" | "s":  # TODO unsure about s
+                    type_id = 1
+                case "m" | "R" | "G" | "B":  # TODO split colors?
+                    type_id = 2
+                case "_" | "a":
```
lcodes, line 108: Ok so only use I see for these cases is to map them to the same color scheme per type_id as replays

otherwise this all collapses into type_id = 2 as the default case

PR #1990, opened 2025-08-07
Make episode desync configurable

daphnedemekas:
If you train with trainer.env_overrides.episode_desync=false then the episodes will all be in sync


daveey (changes requested):
Don't we need to set it in mettagrid as the default?

daveey (approved):
Oh I see, I think it's better to set it in mettagrid.yaml so people can see it as a valid param

PR #1997, opened 2025-08-08
devops/skypilot/launch.py git status check bugfix

zfogg:
this fixes something i only halfway fixed earlier today.

this was returning a tuple which evaluated to `True` when we should have been checking the first value of the tuple which would be `False` often when you want code to run.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211004111034058)

### DIFF
--- a/common/src/metta/common/util/git.py
+++ b/common/src/metta/common/util/git.py
@@ -86,10 +86,12 @@ def get_commit_message(commit_hash: str) -> str:
 
 
 def has_unstaged_changes(allow_untracked: bool = False) -> tuple[bool, str]:
-    """Return (is_dirty_tracked, status_output) considering only tracked file changes.
+    """Returns if there are any unstaged changes in the local git checkout.
+
+    If allow_untracked is True, then unstaged changes in the form of new untracked files will be ignored.
 
     Interpretation of porcelain codes:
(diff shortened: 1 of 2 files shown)


daveey (approved):

rjwalters (changes requested):
I would rather you just unpack the tuple in launch.py:

```
def check_git_state(commit_hash: str) -> str | None:
    error_lines = []

    is_dirty_tracked, status_output = has_unstaged_tracked_changes()
    if is_dirty_tracked:
        error_lines.append(red("❌ You have uncommitted changes that won't be reflected in the cloud job."))
        error_lines.append("Options:")
        error_lines.append("  - Commit: git add -u && git commit -m 'your message'")
        error_lines.append("  - Stash: git stash")
        error_lines.append("\nDebug:\n" + status_output)  # optional
        return "\n".join(error_lines)

    if not is_commit_pushed(commit_hash):
        commit_display = commit_hash[:8]
        error_lines.append(
            red(f"❌ Commit {commit_display} hasn't been pushed and won't be reflected in the cloud job.")
        )
        error_lines.append("Options:")
        error_lines.append("  - Push: git push")
        return "\n".join(error_lines)

    return None
```

PR #1923, opened 2025-08-05
feat: switch SkyPilot CI test to multi-node training

rjwalters:
### Summary
This PR switches our GitHub Actions SkyPilot CI test from single-node to multi-node training by default, and fixes several infrastructure issues that were preventing multi-node training from working correctly.

### Background
Recently we reverted code that broke multi-node training in #1916. In debugging the problem, we discovered that our SkyPilot setup scripts had bugs and were not correctly configuring environment variables needed for distributed training. The root issue is that SkyPilot exposes different environment variables in the `setup` vs `run` phases.

### The Problem
SkyPilot provides cluster-specific environment variables (`SKYPILOT_NUM_GPUS_PER_NODE`, `SKYPILOT_NODE_IPS`, `SKYPILOT_NODE_RANK`, etc.) that are **only available during the `run` phase**, not during the `setup` phase.

This causes issues when:
1. You try to use these variables in the `setup` phase - they're empty/undefined
2. You try to write them to files (like `.bashrc`) during setup - they get written as empty values
3. Multi-node distributed training fails because nodes can't discover each other without these variables

#### 1. **CI Configuration**
- Changed default CI test configuration to use 2 nodes with 2 GPUs each (previously single node)
- Set shorter timeout (10 minutes) for faster CI feedback
- CI now tests multi-node distributed training on every push to main

#### 2. **Environment Configuration Refactoring**
- Created `configure_environment.sh` to centralize all environment setup
- Moved environment configuration from `setup` to `run` phase where SkyPilot variables are available
- Added proper NCCL configuration for multi-node communication
- Introduced `LazyPath` utility to handle file creation on-demand

#### 3. **NCCL and Shared Memory Improvements**
- Added comprehensive NCCL diagnostics and testing script (`test_nccl.py`)
- Enabled shared memory (SHM) by default with proper Docker configuration
- Added IPC_LOCK capability and proper ulimits for NCCL performance
- Set optimal NCCL parameters for multi-node training

#### 4. **Infrastructure Fixes**
- Fixed common setup script extraction to use curl from GitHub
- Added proper Docker run options for multi-node NCCL communication
- Improved error handling and debugging output
- Added pre-flight NCCL connectivity tests before training starts

### SkyPilot Documentation
The environment variables are documented here: [[SkyPilot Environment Variables](https://skypilot.readthedocs.io/en/latest/running-jobs/environment-variables.html)](https://skypilot.readthedocs.io/en/latest/running-jobs/environment-variables.html)

Key variables for distributed training:
[...]

PR #2000, opened 2025-08-08
feat: switch to METTA_ENV_FILE

rjwalters:
This contains some of the changes in progress in #1923 to make that PR more manageable.

This PR introduces `METTA_ENV_FILE` as a centralized location for runtime environment configuration in child scripts.

We also inline and remove the `devops/setup.env` file for clarity.

### 1. New `METTA_ENV_FILE` constant
- Added `METTA_ENV_FILE` to `constants.py` pointing to `~/.metta_env_path`
- Made `constants.py` executable to allow shell scripts to retrieve constant values
- Enables dynamic environment configuration based on runtime conditions
- 
### 2. Cost monitor integration
- Updated `cost_monitor.py` to write `METTA_HOURLY_COST` to `METTA_ENV_FILE`
- Enables parent processes to access cost information set by child processes

### 3. Removed static configuration
- Deleted `devops/setup.env` 
- Removed all `source ./devops/setup.env` calls from scripts and moved ENV setup inline

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211005095607490)


zfogg (approved):
i like it for skypilot. would be nice if the entire project supported this, i.e. our python coded always loaded and applied that file to the environment when it starts. because i would use it haha

common/src/metta/common/util/constants.py
```diff
@@ -18,3 +23,28 @@
 METTA_AWS_REGION = "us-east-1"
 
 METTA_SKYPILOT_URL = "skypilot-api.softmax-research.net"
+
+METTA_ENV_FILE = Path(os.path.expanduser("~/.metta_env_path"))
```
berekuk, line 27: The name of this file is confusing. Why "env **path**"?

Also, we have `~/.metta` dir that we use in `metta.setup`, it's better to store any global metta-related files under it.

The semantics of this file is confusing to me too. We had [this discussion](https://app.asana.com/1/1209016784099267/project/1209041170403492/task/1210887971625824?focus=true) about python-dotenv, but this file is not it. The use case of storing `METTA_HOURLY_COST` in this PR seems very specific, and I'm not even sure that we need `METTA_HOURLY_COST` (why don't we query AWS APIs in system_monitor directly, maybe with some caching if it's too slow? why do we need that separate step of writing it to the file, and then reading it?). Maybe there's a plan here to expand the file's usage, but I don't understand the scope.

devops/skypilot/config/sandbox.yaml
```diff
@@ -11,39 +11,58 @@ file_mounts:
...
+    source: s3://softmax-train-dir
     mode: MOUNT_CACHED
 
 setup: |
+  set -e
   cd /workspace/metta
-  git fetch
-  git checkout $METTA_GIT_REF
 
+  git fetch origin "$METTA_GIT_REF" || git fetch --depth=1000 origin
```
berekuk, line 21: I still don't understand how this is an improvement upon `git fetch`, which would always fetch everything.

berekuk (commented):
I know this is already merged, but I'll post this review anyway.

PR #2007, opened 2025-08-08
'make test-setup' for 'metta install'. pytests

zfogg:
pytests for the metta setup process.

also a few fixes in setup components to make it happen.

this PR does change a few setup components. it modifies them with checks to environment variables to see if we're in a mode where we shouldn't do interactive things, like wait for user input at the terminal or open the user's browser for input. this is so Docker containers and GitHub Actions workflows can install and test our code properly. these setup tests are a good example because they'll run in GitHub actions.

i added to `.github/workflows/checks.yml` to get install environments working as a github action.

`make test-setup` to test locally

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211008375122742)


conftest.py
```diff
@@ -1,20 +1,7 @@
-import sys
-from pathlib import Path
-
 import pytest
 
 from metta.common.tests.fixtures import docker_client_fixture
 
-print(f"\n===== applying conftest from {Path(__file__)} =====")
```
rjwalters, line 8: why are we removing this? just curious

devops/aws/setup_aws_profiles.sh
```diff
@@ -102,18 +87,54 @@ EOF
...
 
   echo "AWS profiles have been configured successfully."
-  grep -q '^export AWS_PROFILE=' ~/.zshrc 2> /dev/null || echo -e '\nexport AWS_PROFILE=softmax' >> ~/.zshrc
+
+  # Function to get the correct zshrc path based on ZDOTDIR
+  get_zshrc_path() {
+    if [ -n "$ZDOTDIR" ]; then
+      echo "$ZDOTDIR/.zshrc"
+    else
+      echo "$HOME/.zshrc"
```
rjwalters, line 96: do we want to support other shells here?

.github/workflows/checks.yml
```diff
@@ -377,9 +377,40 @@ jobs:
           chmod +x .github/scripts/upload_codecov.py
           SUBPACKAGES="${{ matrix.package }}" .github/scripts/upload_codecov.py
 
+  setup-tests:
```
rjwalters, line 380: this doesn't feel like it is on the critical path for the other tests in this workflow?  I think it might be slightly better to break it out into a new file at .github/workflows/setup-tests.yml? up to you

rjwalters (approved):
see comments but otherwise looks good

PR #2009, opened 2025-08-08
Fix to support torch compile

Al-does:
Adjusted how non-tensor information is stored in the tensordict. Code is longer but it's what's needed to support compile. Note that I tried NonTensorData but compile caused it to hang - no error message. Switched back to viewing the data as a tensor with the appropriate size. Requires more lines of code but works.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211010867626691)


zfogg:
@claude explain this PR to me please. make sure it's alright

PR #2019, opened 2025-08-09
Composable Losses

Al-does:
### Goals
Shift from where our codebase was, hardcoded for 1) LSTM memory, 2) PPO and 3) expecting that we only need logits, value, and entropy for all computations, to an RL platform affording rapid implimentation of highly-varied losses and architectures.

Refactoring the RL trainer without a presupposed agent architecture or loss algorithm has plenty of product design choices and challenges. All we know is that we have some parameterized model(s) that interacts with the environment and then wishes to be optimized based on some unknown loss.

### Strategy
Let each loss be a simulacra of trainer.py. We afford the researcher all the tools they need to write exotic losses in a single file. For instance, PPO is contained in a single class in a single file, allowing one to make a new version of that one file and tweak it however they wish. We therefore contain iterative, exploratory changes. This also means allowing losses to get messy in the search for RL breakthroughs while keeping the rest of the code clean and stable.

The result is lightweight but should allow the researcher to implement a great many configurations and combine them with others by simply writing a class or two and updating the agent's config.

### Issues 
- There's some awkwardness with control flow. Trainer is the de facto driver but curriculum might eventually be a better and more direct scheduler. I had to create a TrainerState class to pass necessary information to losses.
- Configs are clunky. I'm expecting major shifts in how we manage cfgs so I dumped everything loss-related into the agent cfg. This will need to move.
- Imminent to dos: eliminate Param Layer and replace with independent losses, add a way to save and reload loss state, add the ability to change losses attached to the policy when restarting training, update MettaAgent to gather loss specs from each layer.
- Future work could involve feeding agent stats into BaseLoss.

### Changes
The BaseLoss interface provides all the tools you need to build your losses. Subclassing it affords the ability to:
- control schedule including operating on different phases (e.g. Phasic Policy Gradient)
- spec what goes in the replay buffer
- take and use outputs from other losses ran before it
- reuse computations run on the agent by other losses and call different leaf nodes of the network
- load policies and hold its own policies (e.g. kickstarting, EMA/momentum/BYOL, First Explore, Then Exploit, etc.)
- spec which losses should be tracked and plotted
- block gradients both up and down
- change optimizer
- generally hack anything else you might need
- Updated LossTracker (renamed Loss) to be programmatically defined.

Outside of losses, the codebase no longer has the priors of PPO and LSTMs.

### Considerations
With composition, scheduling is a primary consideration. There are two senses:
1. At which agent_step count do our losses run rollouts or train? This is tentatively handled by the scheduler in the config but it should eventually be handled by curriculum and, eventually, the gardener. For now, each loss class determines should_run based on agent_step. A loss's schedule in rollout can be different than in train.
2. During a given rollout or minibatch update, which losses run first? Losses are run in the order they are listed in the config.
 This also means that they can share data or process the same data which provides opportunities for efficiencies. In training, your first loss samples experience and writes it to the shared loss minibatch TensorDict for other losses to use. This first loss should also include the policy output so that subsequent losses don't re-run parts of the network that already computed. If subsequent losses don't want to use that minibatch then they're free to instead sample their own and/or to re-run the network.

Using MettaAgent is slightly different. With enhanced neural architecture flexility we need to delegate the management of the agent's state to its layers. Now, we have methods to tell the agent the nature of our ask. Specifically, this means calling methods that tell it more about the type of computation we're trying to perform. There are currently four options: on_new_training_run, on_rollout_start, on_train_mb_start, and on_eval_start. This is hopefully the simplest set that is expansive enough to afford the various experiments we might want to run. As an example, on_eval_start includes things like setting policy.eval() and reset_memory(). on_rollout_start() could mean reset memory for certain layers or to store context in a transformer. Layers therefore have the flexibility to handle a lot of complexity that would have otherwise muddied the code external to the agent, requiring it handle various architectures.
When working with layers that require some state, use the layer to manage its own state if that state should be used in inference or eval. But if it's a concept that needs to be used across from rollout to training then use the experience buffer spec.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211061517827638)


metta/rl/trainer.py
```diff
@@ -324,14 +356,15 @@ def train(
...
+                    for _lname in list(policy_losses):
+                        loss_obj = loss_instances[_lname]
+                        loss_obj.rollout(td, trainer_state)
 
                     # Store experience
-                    experience.store(
-                        data_td=td,
-                        env_id=training_env_id,
-                    )
+                    experience.store(data_td=td, env_id=training_env_id)
```
Al-does, line 367: Currently working on moving this into the Loss classes so they can decide how to manipulate the experience buffer using the policy's output (and decide _which_ policy to update).

agent/src/metta/agent/component_policy.py
```diff
@@ -29,7 +30,7 @@ def __init__(
         action_space: Optional[gym.spaces.Space] = None,
         feature_normalizations: Optional[dict[int, float]] = None,
         device: Optional[str] = None,
-        cfg: DictConfig = None,
+        cfg: Optional[DictConfig] = None,
```
relh, line 33: why optional?

agent/src/metta/agent/component_policy.py
```diff
@@ -63,19 +64,14 @@ def __init__(
             comp_dict = dict(component_cfgs[component_key], **self.agent_attributes, name=component_name)
             self.components[component_name] = instantiate(comp_dict)
 
-        # Setup components
-        component = self.components["_value_"]
-        self._setup_components(component)
-        component = self.components["_action_"]
-        self._setup_components(component)
+        # Setup components by triggering leaf node setup
+        for component_name in self.cfg.leaf_components:
```
relh, line 68: do we have examples beyond value and action? maybe can comment # leaf components include value, action, etc

don't want to generalize too quickly.

should also call "heads" that's the typical neural network language (if they're outputs)

agent/src/metta/agent/component_policy.py
```diff
@@ -63,19 +64,14 @@ def __init__(
...
+            component = self.components[component_name]
+            self._setup_components(component)
 
-        # Track components with memory
-        self.components_with_memory = []
```
relh, line 73: do components not have memory anymore? or I guess maybe they handle it themselves?

agent/src/metta/agent/component_policy.py
```diff
@@ -232,41 +228,31 @@ def activate_action_embeddings(self, full_action_names: list[str], device):
...
-            comp.reset_memory()
-
-    def get_memory(self) -> dict:
-        """Get memory state from all components that have memory."""
-        memory = {}
-        for name in self.components_with_memory:
-            memory[name] = self.components[name].get_memory()
-        return memory
+    def get_agent_experience_spec(self) -> Composite:
+        return Composite(
```
relh, line 232: I guess this is saying that by default a component should have env_obs and dones in its experience spec?

agent/src/metta/agent/component_policy.py
```diff
@@ -232,41 +228,31 @@ def activate_action_embeddings(self, full_action_names: list[str], device):
...
+            env_obs=UnboundedDiscrete(shape=torch.Size([200, 3]), dtype=torch.uint8),
+            dones=UnboundedDiscrete(shape=torch.Size([]), dtype=torch.float32),
+        )
 
-    # ============================================================================
-    # Weight/Training Utility Methods
-    # ============================================================================
+    def on_new_training_run(self):
```
relh, line 237: as mentioned I am opposed to on_X where X is control-flow based and not semantic (but that's fine!).

Is there an example of this on_new_training_run that is different from an __init__() setup? Like what kind of things happen here (just two examples would be good)

agent/src/metta/agent/component_policy.py
```diff
@@ -232,41 +228,31 @@ def activate_action_embeddings(self, full_action_names: list[str], device):
...
+        for _, component in self.components.items():
+            component.on_new_training_run()
 
-    def clip_weights(self):
-        """Apply weight clipping if enabled."""
-        if self.clip_range > 0:
-            self._apply_to_components("clip_weights")
+    def on_rollout_start(self):
```
relh, line 241: can I have 1-2 examples of what happens here?

agent/src/metta/agent/component_policy.py
```diff
@@ -232,41 +228,31 @@ def activate_action_embeddings(self, full_action_names: list[str], device):
...
+        for _, component in self.components.items():
+            component.on_rollout_start()
+
+    def on_train_mb_start(self):
```
relh, line 245: can I have 1-2 examples of what happens here?

agent/src/metta/agent/component_policy.py
```diff
@@ -232,41 +228,31 @@ def activate_action_embeddings(self, full_action_names: list[str], device):
...
+        for _, component in self.components.items():
+            component.on_train_mb_start()
 
-    def l2_init_loss(self) -> torch.Tensor:
-        """Calculate L2 initialization loss for all components."""
-        losses = self._apply_to_components("l2_init_loss")
-        return torch.sum(torch.stack(losses)) if losses else torch.tensor(0.0, dtype=torch.float32)
+    def on_eval_start(self):
```
relh, line 249: can I have 1-2 examples of what happens here?

agent/src/metta/agent/component_policy.py
```diff
@@ -304,26 +290,3 @@ def update_normalization_factors(self, features: dict[str, dict], original_featu
                 # Update the component's normalization tensor
                 if hasattr(component, "update_normalization_tensor"):
                     component.update_normalization_tensor(norm_tensor)
-
-    # ============================================================================
-    # Helper Methods and Properties
-    # ============================================================================
-
-    def _apply_to_components(self, method_name, *args, **kwargs):
```
relh, line 312: are we deleting this because components handle things themselves now?

agent/src/metta/agent/lib/lstm.py
```diff
@@ -49,6 +49,18 @@ def __setstate__(self, state):
         self.lstm_h.clear()
         self.lstm_c.clear()
 
+    def on_new_training_run(self):
```
relh, line 52: I see these are all good places where memory should be reset. This far through reading I am curious if it would be cleaner to not have flow-based callbacks but off we go reading more yay

agent/src/metta/agent/lib/nn_layer_library.py
```diff
@@ -24,6 +24,42 @@
 
 from metta.agent.lib.metta_layer import LayerBase, ParamLayer
 
+# def initialize_weights(network: nn.Module, initialization: str, nonlinearity: str | None = None):
```
relh, line 27: Do we want to keep this commented out block?

agent/src/metta/agent/lib/obs_token_to_box_shaper.py
```diff
@@ -34,6 +34,17 @@ def __init__(self, obs_shape, obs_width, obs_height, feature_normalizations, **c
         self._out_tensor_shape = [self.num_layers, self.out_width, self.out_height]
 
     def _forward(self, td: TensorDict):
+        # if td.batch_dims > 1:
```
relh, line 37: do we want this comment?

agent/src/metta/agent/metta_agent.py
```diff
@@ -122,19 +123,47 @@ def forward(self, td: Dict[str, torch.Tensor], state=None, action: Optional[torc
...
-        """Reset memory - delegates to policy if it supports memory."""
-        if hasattr(self.policy, "reset_memory"):
-            self.policy.reset_memory()
+    def get_cfg(self) -> DictConfig:
+        return self.cfg
 
-    def get_memory(self) -> dict:
-        """Get memory state - delegates to policy if it supports memory."""
-        return getattr(self.policy, "get_memory", lambda: {})()
+    def on_new_training_run(self):
```
relh, line 129: Instead of using hasattr, this is something we COULD require of all MettaAgents. Do you think that would be worthwhile?

agent/src/metta/agent/metta_agent.py
```diff
@@ -122,19 +123,47 @@ def forward(self, td: Dict[str, torch.Tensor], state=None, action: Optional[torc
...
+            return self.policy.get_agent_experience_spec()
+        else:
+            return Composite(
+                env_obs=UnboundedDiscrete(shape=torch.Size([200, 3]), dtype=torch.uint8),
+                dones=UnboundedDiscrete(shape=torch.Size([]), dtype=torch.float32),
+            )
+
+    def attach_replay_buffer(self, experience: Experience):
+        """Losses expect to find a replay buffer in the policy."""
+        self.replay = experience
```
relh, line 166: Does it make sense for the agent to store a reference to the experience replay buffer? 

I do wonder if maybe the agents should store their state and then the Experience replay buffer should go to the agents and collect it from them to use.

I do kind of like the idea of agents knowing how to do their own training, but then all-agent-wide operations like prioritized experience replay and/or advantage filtering basically are global top-down operations, and might say something like "I dont care about Agent X experience".

What do you think?

agent/src/metta/agent/metta_agent.py
```diff
@@ -287,27 +316,6 @@ def lstm(self):
             return self.policy.lstm
         return None
 
-    def l2_init_loss(self) -> torch.Tensor:
```
relh, line 290: Did we delete this because they don't belong in MettaAgent or because they're not used even by ComponentPolicy? am curious

configs/agent/fast.yaml
```diff
@@ -12,6 +12,87 @@ observations:
...
+    clip_coef: 0.1
+    ent_coef: 0.0021
+    gae_lambda: 0.916
+    gamma: 0.977
+
+    # Training parameters
+    max_grad_norm: 0.5
+    vf_clip_coef: 0.1
+    vf_coef: 0.44
+    l2_reg_loss_coef: 0
```
relh, line 34: if we're going to split apart losses, the l2 regularization almost feels like it should be it's own.

however at the same time, "PPO" means something specific and entails a number of different operations that unless you do all of them don't mean "PPO", what do you think?

configs/agent/fast.yaml
```diff
@@ -12,6 +12,87 @@ observations:
...
+  #     train:
+  #       begin_at_epoch: 40
+  #       end_at_epoch: 900
+
+  #   anneal_ratio: 0.5 # ratio of kickstart_epochs to anneal to 0 coef
+  #   teacher_uri: wandb://run/m.av.student.probe.finalv2.001 #wandb://run/m.av.loss.leader.006
+  #   action_loss_coef: 1
+  #   value_loss_coef: 1
+
+optimizer:
```
relh, line 82: should an agent define its optimizer?

should an agent define its losses?

I think that the fast.yaml should define what it means to be a fast agent and not include other things that we may want to mix and match with the fast agent. 

what do you think?

configs/agent/fast.yaml
```diff
@@ -12,6 +12,87 @@ observations:
...
+  type: adam
+  beta1: 0.9
+  beta2: 0.999
+  eps: 1e-12
+  learning_rate: 0.000457
+  weight_decay: 0
+  clip_grad_norm: 0
+
+leaf_components: # output heads
```
relh, line 91: could make these output_heads

configs/agent/prototypes/fast_ks.yaml
```diff
@@ -0,0 +1,219 @@
...
+    # V-trace clipping parameters (for off-policy corrections)
+    vtrace:
+      rho_clip: 1.0 # Default: on-policy (no off-policy correction)
+      c_clip: 1.0 # Default: on-policy bootstrapping
+
+  # EMA:
+  #   decay: 0.996
+  #   loss_coef: 1.0
+
+  TLKickstarter:
```
relh, line 54: not sure what a prototype is, is this supposed to be like a fast agent in the above config but it also has changes that are related to kickstarting?

metta/rl/experience.py
```diff
@@ -145,35 +143,6 @@ def reset_for_rollout(self) -> None:
         self.ep_indices = self._range_tensor % self.segments
         self.ep_lengths.zero_()
 
-    def reset_importance_sampling_ratios(self) -> None:
-        """Reset the importance sampling ratio to 1.0."""
-        if "ratio" in self.buffer.keys():
-            self.buffer["ratio"].fill_(1.0)
-
-    def sample_minibatch(
```
relh, line 153: what happened to this function? maybe it just moved

metta/rl/loss/loss_tracker.py
```diff
@@ -0,0 +1,77 @@
+from __future__ import annotations
+
+from typing import Iterable
+
+
+class LossTracker:
+    """
+    Flexible dynamic loss/metric tracker. Interacts heavily with BaseLoss and leans on implemented losses.
+    TODO: investigate if this programmatic path has SPS implications. If so, delete LossTracker and have losses handle
```
relh, line 9: I think it might be nice to let Losses handle their own thing. That way one can pull up the loss class and immediately know all the things related to that loss (rather than other things existing in a tracker elsewhere)

metta/rl/trainer.py
```diff
@@ -192,54 +195,65 @@ def train(
...
+            weight_decay=int(policy_cfg.optimizer.weight_decay),
         )
     else:
         raise ValueError(f"Optimizer type must be 'adam' or 'muon', got {optimizer_type}")
 
+    # Instantiate configured losses dynamically by class name
+    loss_instances: dict[str, BaseLoss] = {}
+    for loss_instance_name, loss_config in policy_cfg.losses.items():
+        module_path, class_name = loss_config["path"].rsplit(".", 1)
+        module = importlib.import_module(module_path)
```
relh, line 224: I know there's probably not a better way to "dynamically/generally" do this but it's still kind of ugly :/

metta/rl/trainer_state.py
```diff
@@ -0,0 +1,20 @@
+from dataclasses import dataclass
+
+import torch
+
+
+@dataclass(slots=True)
+class TrainerState:
```
relh, line 7: should this be a class? do we need all of these variables? does the optimizer belong as part of the TrainerState here?

I know that elsewhere there is a TrainerCheckpointer and it might be the right place

metta/sim/simulation.py
```diff
@@ -350,9 +350,9 @@ def simulate(self) -> SimulationResults:
         """
         self.start_simulation()
 
-        self._policy_pr.policy.reset_memory()
+        self._policy_pr.policy.on_eval_start()
```
relh, line 353: I think this is the example of why I don't like control flow methods. They hide what's happening. If I was reading the simulation.py file, I could previously know that memory was getting reset here. Now I don't know that

relh (changes requested):
As I messaged you, right now it's a bunch of intense comments, but I want to just put my thoughts down comprehensively and then hear your thoughts. Then we can be free to dismiss a bunch of my opinions and get it merged~

metta/rl/experience.py
relh: it moced!

relh:
One other thing is that I like the "losses as a hologram" of the trainer, but it starts to seem weird when things like PPO might have a bunch of pieces in them (like regularization) that are what makes them PPO but would be probably different losses if they weren't already called PPO. What do you think?

configs/agent/prototypes/fast_ks.yaml
Al-does: I kind of need a folder for configs that are works in progress, in between "my own branch" and checked in. You're making me realize I should put an author name on prototypes so we know who to ask when we want to clean things up.

metta/rl/loss/loss_tracker.py
Al-does: Great call. I'll refactor the summing to happen at the loss class. I think it might be more computationally efficient too.

configs/agent/fast.yaml
Al-does: Yes good call.

configs/agent/fast.yaml
Al-does: Good questions and I have a proposal! But before getting into that, let me just say that the loss config is really rough and there should be a better paradigm for handling scheduling too. I'm waiting for the upcoming changes (and your council) to sort this one out.

Okay now for the proposal: I think that the vast majority of agent performance is currently defined by variables set in trainer.yaml and agent.yaml. I think that these variables are often interrelated and splitting this information across files causes confusion - when changing a config param in one file you may also need to change it in another. It's not like they are truly independent. Many of these hyperparameters are contingent on the agent architecture. For instance, I think optimizer params should be set by the agent, not by trainer, since we may want learning rates depending on the agent architecture and the other losses involved. And the optimizer is how we implement l2 renorm.

This would lead to a workflow whereby the researcher copies the entire config (an admitedly long one) and modifies what they wish. But here's the amazing thing with that approach: their entire training run is visible in front of them in one file, fewer than 200 lines. They don't need to know much about the repo to rearrange the architecture, identify that they don't like certain hypers and change them, add or delete losses, and have the whole thing in front of them at a glance.

Meanwhile, the code that implements these things is modular, affording interchangeability, composability, and the ability to readily add new modules.

metta/rl/trainer_state.py
Al-does: I don't know! I figured a class was nicer than just passing a list. But TrainerCheckpointer sounds much better. I'll look into that instead. Thanks.

agent/src/metta/agent/metta_agent.py
Al-does: This is a great question. I really just selected this approach on a hunch, namely, that some components of the agent might want to access their own replay buffer. With the component agent, each layer should be able to handle its own state. But perhaps there's something to being able to read the history from other layers. I then couldn't see many downsides with alternate approaches like leaving it a part of trainer. But I can see both ways working well and it might not be too hard to change back and forth.

I'm afraid I don't understand the "global top-down operations" thing. I'm trying to understand what is the top and what's operating on what. Can you give me an example or clarify for me?

If helpful, prio sampling is handled by PPO as I assumed it's a choice that the loss class should make, not the buffer. I ended up taking this approach that the buffer should really just be a dumb bucket but afford easy retrieval, handling shapes. Then losses or agents or whatever can query it however they like.

agent/src/metta/agent/metta_agent.py
Al-does: I don't feel strongly about this. Just figured that memory specific methods at the generic agent level might be a bit fraught: On one extreme, I recall Manuel testing with a memory-less agent. On the other (and I think this is the important point), getting and setting memory requires understanding the structure of the memory. With that in mind, I figured it would be better to expect the caller to know what they're doing which would mean knowing how to navigate the agent's component/attribute tree to find the memory layer and query its components where it might be lstm hidden and cell or obs sequence or latent tokens or some multitude of fast and slow LSTMs lol. Figured that if the caller can handle all those data structures then they would likely know how to get to memory. LMK what you prefer!

agent/src/metta/agent/lib/nn_layer_library.py
Al-does: Thank you. Half way between deleting ParamLayer which, heretofor, has handled initializing Linear and Conv2d.

agent/src/metta/agent/component_policy.py
Al-does: The idea is that it would be more general to tell each component which phase it's at and let it figure out what it needs to do.

agent/src/metta/agent/component_policy.py
Al-does: Yes, great call. "heads" is much better.

Other examples are predictive loss heads or decoders for reconstruction losses.

agent/src/metta/agent/component_policy.py
Al-does: Thanks for working with me on control flow stuff. I read your other comments about it -  great points. Let's start a thread here?

I'm now thinking that the right answer is a mix of _some_ control flow but, to your point, keeping a reset_memory.

Here are examples where control flow involves more than memory:
- use LSTMCell during rollout and eval but switch to LSTM during training
- consistent dropout masks: that layer needs to do one thing during rollout (create a new mask), another during train (read and apply a stored mask), and eval (don't use any mask at all)
- fast/slow RNNs where one might need to be reset more frequently than another.
- on training run start could make certain things more generic in a good way: for instance the obs sequence encoder can recalculate its Fourier sequencing scheme based on updated obs window.
- turning on/off autocast or changing precision at specified layers whether in train vs eval
- something causal transformer related??

I think that all of these have alternative implementations that could involve specific hooks in the rest of the training flow. I also don't want to over provision at the expense of clarity.

Discussing with you is certainly shifting my thinking - I'm now thinking that what's best is to have a reset memory method, delete the on_train_start method, but leave the on_rollout_start and the on_train_phase start methods. I could also see deleting even those later control flow methods on the policy too, LMK what you think. The only place where I think these methods are pretty critical is between BaseLoss and trainer, not the agent as much.

Al-does:
@relh to your comment 

> One other thing is that I like the "losses as a hologram" of the trainer, but it starts to seem weird when things like PPO might have a bunch of pieces in them (like regularization) that are what makes them PPO but would be probably different losses if they weren't already called PPO. What do you think?

I totally agree. I wanted to separate out anything that is an additive term in the loss function into its own class. My tentative hack is to set their coefficients to zero but that's inefficient since the computations still run. Just didn't get around to separating it from PPO.py. Will flag as a to do for me.

agent/src/metta/agent/metta_agent.py
relh: I agree that parts of a network may/may not have memory and they may/may not do things differently with it. I do agree that it makes sense for that part of the network itself to know what to do. Hmm hmm.

relh:
distributed training one node multi gpu works

PR #2034, opened 2025-08-10
fix: repair broken tests

rjwalters:
This PR fixes all failing tests.

## Key Changes

* **Test utilities refactor**

  * Introduced `TestEnvironmentBuilder`, `TokenTypes`, and `ObservationHelper` under `metta/m ettagrid/test_support` for reusable test environment creation and observation handling.
  * Removed ad-hoc `make_test_config` and duplicated token logic from individual tests.
* **Curriculum tests**

  * Added `mock_hydra_everywhere` fixture to bypass full Hydra config loading in unit tests, while preserving real config loading for specific YAML-based cases.
  * Fixed import path issues for `metta.map.mapgen`.
* **Environment tests**

  * Updated 8-way and cardinal movement tests to use the new `TestEnvironmentBuilder` API.
  * Added boundary and integration test cases for both movement modes.
  * Simplified map/object setup in observation-related tests and ensured type safety in object definitions.
* **Observation tests**

  * Centralized token lookups in `ObservationHelper` for consistency.
  * Fixed `test_global_tokens_update` to use `max_steps=10` for predictable completion percentages.
  * Added builder-based altar setup for edge and surround tests.
* **Miscellaneous**

  * Removed path manipulation in `metta/__init__.py` to avoid namespace conflicts.
  * Cleaned up unused `conftest.py` files and redundant imports.
  * Updated `pyproject.toml` and `.cspell.json` to include new settings and words.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211018504787833)


.github/actions/disk-cleanup/action.yml
```diff
@@ -0,0 +1,25 @@
+name: "Fast Disk Cleanup"
+description: "Quickly free up ~16GB by removing Android SDK and swap"
+author: "Your Organization"
```
daveey, line 3: ??

tests/test_validate_all_curriculums.py
```diff
@@ -33,6 +34,20 @@ class TestValidateAllCurriculums:
...
+            parts = type(curriculum).__module__.split(".")
+            core_module_path = ".".join(parts[:-1]) + ".core"
+            core_module = sys.modules.get(core_module_path)
+
+            if core_module and hasattr(core_module, "Curriculum"):
+                Curriculum = core_module.Curriculum
+                assert isinstance(curriculum, Curriculum)
+            else:
+                # Fallback: check if it's a Curriculum by checking the MRO
+                assert any(cls.__name__ == "Curriculum" for cls in type(curriculum).__mro__)
```
berekuk, line 53: This shouldn't be needed, revert. (I expect it'll pass after you bring back pkgutil code in `__init__.py` )

mettagrid/src/metta/mettagrid/test_support/token_types.py
```diff
@@ -0,0 +1,32 @@
+from dataclasses import dataclass
+
+
+# Constants from C++ code
+@dataclass
+class TokenTypes:
```
daveey, line 6: this is sus. tokens are computed dynamically based on enabled featues

mettagrid/pyproject.toml
```diff
@@ -70,11 +71,12 @@ wheel.packages = ["src/metta"]
 # So for now we're recompiling the entire package on every C++ code change.
 
 [tool.pytest.ini_options]
-# Configuring pytest here causes us to ignore metta/conftest.py
+markers = ["slow: marks tests as slow (deselect with '-m \"not slow\"')"]
 testpaths = ["tests", "benchmarks"]
 python_files = ["test_*.py"]
 python_functions = ["test_*"]
 python_classes = ["Test*"]
+addopts = ["--rootdir=.."]
```
berekuk, line 79: huh, why is this necessary?

mettagrid/src/metta/mettagrid/test_support/token_types.py
rjwalters: comes from mettagrid/src/metta/mettagrid/objects/constants.hpp -- we could try to export these via pybind maybe... I'll add a TODO for now

mettagrid/tests/test_curriculum.py
```diff
@@ -30,6 +30,97 @@
 from metta.mettagrid.curriculum.random import RandomCurriculum
 from metta.mettagrid.curriculum.sampling import SampledTaskCurriculum
 from metta.mettagrid.curriculum.util import curriculum_from_config_path
+from metta.mettagrid.room.random import Random
+
+
+@pytest.fixture(autouse=True)
+def mock_hydra_everywhere(monkeypatch):
```
berekuk, line 37: I'm not sure why this is needed and I don't like it just because it's long and complex (especially as a part of this PR); mettagrid tests already pass in `main`.

metta/map/__init__.py
```diff
@@ -0,0 +1 @@
+# metta/map/__init__.py
```
berekuk, line 1: unnecessary
(you have these useless comments in a few other files too, I assume from llms?)

.github/actions/disk-cleanup/README.md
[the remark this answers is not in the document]
berekuk: Upd: I see that you've optimized this while I was reviewing, ok.

mettagrid/tests/test_pufferlib_integration.py
```diff
@@ -15,41 +16,73 @@
...
+
+        # 2) Detect whether the modern training entrypoint exists
+        detect_cmd = [
+            "uvx",
+            "--from",
+            str(pufferlib_dir),
+            "python",
+            "-c",
+            "import importlib.util as i; print(i.find_spec('pufferlib.pufferl') is not None)",
+        ]
```
berekuk, line 63: This test takes ~40-50 seconds on my machine.

It also doesn't test anything from the current branch - pufferlib installs another separate copy of metta repo from `main` branch. (easy to verify by raising an exception in `MettaGridCore` constructor).

So, in its current version, what'll happen is:
- it'll continue to pass on all PRs and even in merge queue, no matter what
- until we merge something in `main` that will break pufferlib
- then it'll continue to fail on all PRs until we force-merge the fix to `main` or remove the test

Bottom line, I think you should disable this test until we figure out a better way to implement it.

berekuk (changes requested):
- I skipped most of agent/ tests because I'm not familiar with that code
- there are many mettagrid tests changes that look more like unrelated refactorings to me, I haven't reviewed those either, would prefer to review them separately but I know we're in a hurry
- new `checks.yml` is good!
- I know this PR is urgent but I think `sys.path` shenanigans should be cleaned up before the merge, and pufferlib test fixed (hard) or disabled

metta/__init__.py
```diff
@@ -1,22 +1 @@
...
-    sys.path.remove(metta_src_dir)
-sys.path.insert(0, metta_src_dir)
-
-if mettagrid_src_dir in sys.path:
-    sys.path.remove(mettagrid_src_dir)
-sys.path.append(mettagrid_src_dir)  # Add it to the end to deprioritize it
-
-
-# Now, extend the path for namespace packaging
-__path__ = __import__("pkgutil").extend_path(__path__, __name__)
```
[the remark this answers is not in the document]
rjwalters: I removed a bunch of path shenanigans and maybe we don't need it? I'll put it back for now since I don't think it will hurt anything

mettagrid/tests/test_curriculum.py
rjwalters: david is refactoring curriculum anyway and this was the solution to a bunch of test failures so I think it is good enough for now (TM)

metta/map/__init__.py
rjwalters: file path comments are a convenience for LLM context and don't hurt anything - if I had my way we would have it in a docstring on every file

metta/__init__.py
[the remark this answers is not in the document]
rjwalters: weird that all the tests pass then? I put it back anyway

mettagrid/tests/test_curriculum.py
berekuk: I checked out your branch and commented this function and the test is still passing.

PR #2040, opened 2025-08-11
Support using stats server to determine "top" policies. 

nishu-builder:
To determine policy scores when selecting "top" policy records, optionally use stats server scores, not just those in policy.metadata. This is useful when policy evals are done exclusively remotely: policy.metadata will not contain eval-related scores. It is also probably just the right model more broadly (shoudlnt expect policy.metadata to be a source of truth on policy performance).


app_backend/src/metta/app_backend/routes/scorecard_routes.py
```diff
@@ -249,19 +254,25 @@ async def get_available_metrics_for_selection(
 
 async def fetch_policy_scorecard_data(
     con: AsyncConnection,
-    training_run_ids: List[str],
-    run_free_policy_ids: List[str],
+    training_run_ids: list[str],
+    run_free_policy_ids: list[str],
+    policy_ids: list[str] | None,
```
Barraketh, line 259: I don't love this api - I think there should be just two parameters: training_run_ids, and 'policy_ids'

nishu-builder: sounds great -- will remove run_free_policy_ids; you're good with (a) allowing policy_ids to specify policies that do belong to a run (and so are not run-free), and (b) not grouping policies by "best"/"latest" within their associated training run (if it exists)?

Barraketh: Yeah. Basically, we should group/select when computing the data from training_run_ids, and we should NOT group or select when computing data for policy_ids.  In fact, those could be refactored into two different functions, just to make the logic more explicit, but that's up to you.

nishu-builder: also sorry i didnt refresh the page so i didnt see your comments and offer to take it over. thank you for the offer! but i'm happy to do this one

I agree on spliting out the grouped/ungrouped selection; done

Barraketh (commented):
Actually, this is subtly wrong for the case when the a policy is individually requested for a training run that's also requested.  We could fix it, but honestly, I'm not sure why we're doing all of this. We should just create a new route for this instead of reusing the 'scorecard' route. The query to get the score of a single policy on a single eval with a single metric is trivial - trying to get the much more complicated scorecard logic to conform isn't useful here

PR #2052, opened 2025-08-11
Create/load policy only on master and then distribute using NCCL

dominik3141:
This is a new version of the reverted PR #1614.

**Summary**
Make sure only the master rank loads/creates a new policy as DDP syncs the master policy to all other ranks at initialization time.

**Context**
From the [PyTorch docs](https://docs.pytorch.org/docs/stable/notes/ddp.html):
> Construction: The DDP constructor takes a reference to the local module, and broadcasts state_dict() from the process with rank 0 to all other processes in the group to make sure that all model replicas start from the exact same state.

**Testing**
Ran training on 2 nodes with 4 GPUs each and everything looks fine. 

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211029236032548)


metta/rl/checkpoint_manager.py
```diff
@@ -211,29 +225,32 @@ def load_or_create_policy(
...
         elif torch.distributed.is_initialized():
+            # Non-master ranks: do not wait for file. Create a fresh policy locally.
             logger.info(
-                f"No existing policy found. Rank {self.rank}: Waiting for master to create policy at {default_path}"
+                (f"No existing policy found. Rank {self.rank}: Creating local policy and syncing metadata via NCCL")
             )
+            policy_record = self.policy_store.create_empty_policy_record(
+                checkpoint_dir=trainer_cfg.checkpoint.checkpoint_dir, name=default_model_name
+            )
+            policy_record.policy = make_policy(metta_grid_env, env_cfg, agent_cfg)
```
mstormbull, line 236: If our models get big, could this be memory intensive to initialize models locally and sync? Fine for now, I think.

dominik3141: The syncing already happens whenever we wrap the agent to become a DDPAgent. This is standard PyTorch behavior, so I assume that their implementation is quite efficient. Also all of the model sharding for LLMs depends on PyTorch broadcasts being quick and efficient, so if there is a bottleneck, these guys will notice long before we do. 
Or do I maybe misunderstand your question?

PR #2057, opened 2025-08-12
Notebook hello world v1

sandover:
- draft of a hello world Jupyter notebook (WIP).  target audience: new softmax employee  (rebased & the notebook broke)
- port of the same notebook to marimo


uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
```
treeform, line 2: going down in version is suspicions

tools/renderer.py
```diff
@@ -125,13 +125,247 @@ def predict(self, obs: np.ndarray) -> np.ndarray:
...
+        # Normalize probabilities to sum to 1.0
+        total = pickup_prob + cardinal_prob + any_prob + rotate_prob
+        self.pickup_prob, self.cardinal_prob, self.any_prob, self.rotate_prob = [
+            p / total for p in [pickup_prob, cardinal_prob, any_prob, rotate_prob]
+        ]
+
+        # Get action indices from environment
+        self._initialize_action_indices()
+
+    # ---------------------------------------------------------------------
```
treeform, line 249: I don't think we use this style headers anywhere else, a smile comment should be enough

tools/renderer.py
```diff
@@ -125,13 +125,247 @@ def predict(self, obs: np.ndarray) -> np.ndarray:
...
+                self.pickup_idx = action_names.index("pickup")
+            else:
+                self.pickup_idx = 2
+        except (AttributeError, ValueError):
+            # Fallback defaults
+            self.move_idx = 0
+            self.rotate_idx = 1
+            self.pickup_idx = 2
+
+    # ------------------------------------------------------------------
```
treeform, line 271: same for this comment

tools/renderer.py
```diff
@@ -125,13 +125,247 @@ def predict(self, obs: np.ndarray) -> np.ndarray:
         return actions
 
 
+class PickupAlwaysPolicy(BasePolicy):
```
treeform, line 128: Why are these policies live in a file called renderer? Should it be hardcoded_policy.py or something?

mettagrid/src/metta/mettagrid/renderer/miniscope.py
```diff
@@ -146,3 +143,16 @@ def render(self, step: int, grid_objects: Dict[int, dict]) -> str:
         self._last_buffer = current_buffer
 
         return current_buffer
+
```
treeform, line 146: Are you using both miniscope and nethack?

uv.lock
treeform: you might have older UV installed?

tools/renderer.py
```diff
@@ -125,13 +125,247 @@ def predict(self, obs: np.ndarray) -> np.ndarray:
...
+        # Get action indices
+        self._initialize_action_indices()
+
+    def _initialize_action_indices(self) -> None:
+        """Initialize action type indices from environment."""
+        try:
+            action_names: List[str] = self.env.action_names
+            self.move_idx: int = action_names.index("move") if "move" in action_names else 0
+            self.rotate_idx: int = action_names.index("rotate") if "rotate" in action_names else 1
+            self.pickup_idx: int = action_names.index("pickup") if "pickup" in action_names else 2
```
treeform, line 148: Is pickup a valid action? I thought it was get/put_item?

experiments/notebooks/demo_minimal_approach.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env -S uv run
```
treeform, line 1: This is not a notebook but it sits as part of notebooks?

uv.lock
sandover: inadvertent, fixed.

tools/renderer.py
sandover: I was going with existing structure (which is admittedly a bit odd), rather than do a refactor.

PR #2062, opened 2025-08-12
feat: auto-generate thumbnails

golergka:


Barraketh (changes requested):
This broadly looks good, except that I don't want the map viewer to have to know about particular thumbnail naming conventions / file organization.  Can we just store the thumbnail url with the episode like replay url?

PR #2070, opened 2025-08-12
Debug skypilot objects

pygma-tau:
Currently skypilot jobs training on arena/random and arena/learning_progress curricula are broken. This is because _level was not being reset properly in MettaGridEnv. This is a patch for that so that curricula with different map configs can be trained on in the short-term, noting that this is not a clean solution.


teodorionita (approved):

mettagrid/src/metta/mettagrid/mettagrid_env.py
```diff
@@ -101,6 +101,13 @@ def _reset_trial(self) -> None:
         game_config_dict = OmegaConf.to_container(task_cfg.game)
         assert isinstance(game_config_dict, dict), "Game config must be a dictionary"
 
+        # explicit code to keep level and config in sync
```
rjwalters, line 104: consider moving to a method to DRY up `self.set_level()`

PR #2072, opened 2025-08-12
remove the ability to run "rm" from claude code

zfogg:
claude can currently run the rm command in our repo. not good practice. he is deleting files without permission. you can put that in ./.claude/settings.local.json if you want claude to be able to rm without permission

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -11,7 +11,6 @@
       "Bash(cat:*)",
       "Bash(pytest:*)",
       "Bash(./devops/test_uv_setup.sh:*)",
-      "Bash(rm:*)",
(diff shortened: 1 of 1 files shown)


.claude/settings.json
```diff
@@ -11,7 +11,6 @@
       "Bash(cat:*)",
       "Bash(pytest:*)",
       "Bash(./devops/test_uv_setup.sh:*)",
-      "Bash(rm:*)",
```
rjwalters, line 14: is this safe? can we sandbox to the repo?

rjwalters: flips this to an addition in my head -- far better to remove!

rjwalters (approved):

PR #2082, opened 2025-08-13
In context learning algorithmically genreated environments and curriculum

daphnedemekas:
Temporary until we have the new env generation merged next week


berekuk (commented):
Even without the upcoming env generation code, won't this be easier to do with a custom curriculum class?

- inherit from `MultiTaskCurriculum`
- copy-paste bits of `RandomCurriculum` and `LearningProgressCurriculum` there (~100 lines of code); reuse `BidirectionalLearningProgress` which does LP heavy lifting
-  generate curricula dynamically in that class

PR #2084, opened 2025-08-13
Major refactor of MettaGrid Env Code

daveey:
1. Clean up the Env hierarchy

MettaGridCore just manages the c_env, and keeps it private
PufferEnvBase handles auto-reset, buffer injection, and on_episode_done handling
MettaGridEnv does everything else, including stats, desync, replays

Envs now take EnvConfig, instead of Curriculum

Curriculum registers itself with the env, and updates env_config between resets.

Simplified map building in mettagrid. EnvConfig contains MapBuilderConfig which is then used to construct the map on every reset.

Cleaned up a lot of the mettagrid tests. We can now just construct EnvConfig to be a minimal config or arena()

mettagrid unittests pass, but I haven't tested the rest. this is WIP

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211041387965288)


mettagrid/src/metta/mettagrid/mettagrid_config.py
```diff
@@ -223,9 +224,7 @@ class GameConfig(Config):
     # and other parts of the template can read from there.
     params: Optional[Any] = None
 
-    # Level map - should be provided by CurriculumEnvWrapper or created directly
-    level_map: Optional[Level] = None
-    map_builder: Optional[Any] = None
+    map_builder: MapBuilderConfigUnion = RandomMapBuilderConfig(agents=24)
```
berekuk, line 227: The `Union` type makes this impossible to extend:
- not possible to pass mapgen configs to it without adding the dependency from mettagrid to metta
- even if you find some trick for mapgen, I think we want other mettagrid users to be able to add their own configs

Was there some problem with `MapBuilderConfig`, as you initially wanted to do?

mettagrid/src/metta/mettagrid/mettagrid_env.py
```diff
@@ -74,57 +70,26 @@ def __init__(
         self._stats_writer = stats_writer
         self._replay_writer = replay_writer
         self._episode_id: str | None = None
-        self._reset_at = datetime.datetime.now()
+        self._last_reset_ts = datetime.datetime.now()
         self._is_training = is_training
 
-        # Initialize with base PufferLib functionality
+        # DesyncEpisodes - when training we want to stagger experience. The first episode
+        # will end early so that the next episode can at a different time on each worker.
```
berekuk, line 77:
```suggestion
        # will end early so that the next episode can begin at a different time on each worker.
```

metta/map/mapgen.py
```diff
@@ -4,17 +4,16 @@
 from omegaconf import DictConfig, OmegaConf
 from pydantic import model_validator
 
-from metta.common.util.config import Config
 from metta.map.scene import load_class, make_scene, scene_cfg_to_dict
 from metta.map.scenes.copy_grid import CopyGrid
 from metta.map.scenes.room_grid import RoomGrid, RoomGridParams
 from metta.map.scenes.transplant_scene import TransplantScene
-from metta.mettagrid.level_builder import Level, LevelBuilder, create_grid
+from metta.mettagrid.game_map import GameMap, GameMapBuilder, GameMapBuilderConfig, create_grid
```
berekuk, line 11: imports and names are wrong here, as you probably already know

metta/rl/trainer_config.py
```diff
@@ -3,7 +3,10 @@
 from omegaconf import DictConfig, OmegaConf
 from pydantic import ConfigDict, Field, model_validator
 
+from metta.cogworks.curriculum import CurriculumConfig
+from metta.cogworks.curriculum import env as env_curriculum
```
berekuk, line 7: just name it `env_curriculum` in the module? otherwise we'll always have to do this

metta/rl/trainer_config.py
```diff
@@ -3,7 +3,10 @@
...
 from metta.common.util.config import Config
+from metta.mettagrid.config.builder import arena as arena_builder
```
berekuk, line 9: Consider naming it `make_arena` in `metta.mettagrid.config.builder`? (functions are verbs, vars are nouns, easier to avoid collisions)

mettagrid/demo.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env -S uv run
+
+"""
+Demo showing how to create a MettaGridEnv and run a simulation with random actions.
```
berekuk, line 4: doesn't do what the description says

metta/rl/vecenv.py
```diff
@@ -32,19 +32,20 @@ def make_env_func(
...
         buf=buf,
         stats_writer=stats_writer,
         replay_writer=replay_writer,
         is_training=is_training,
         **kwargs,
     )
-    # Ensure the environment is properly initialized
-    if hasattr(env, "_c_env") and env._c_env is None:
-        raise ValueError("MettaGridEnv._c_env is None after hydra instantiation")
+    env.set_on_episode_done(lambda obs, rew, term, trunc, info: curriculum.get_task().complete(rew.mean()))
```
berekuk, line 47: Should this call `env.set_episode_config`? I can't find any other code that's supposed to update the config

mettagrid/src/metta/mettagrid/core.py
```diff
@@ -8,14 +8,19 @@
...
 from gymnasium import spaces
 
-from metta.mettagrid.level_builder import Level
 from metta.mettagrid.mettagrid_c import MettaGrid as MettaGridCpp
 from metta.mettagrid.mettagrid_c_config import from_mettagrid_config
+from metta.mettagrid.mettagrid_config import EnvConfig
+
+if TYPE_CHECKING:
+    # Import EpisodeStats type for type checking only
+    # This avoids circular imports at runtime while still providing type hints
```
berekuk, line 22: very minor, but the comment is wrong, you already import from `mettagrid_c` above, and `mettagrid_c` doesn't import anything else

mettagrid/src/metta/mettagrid/puffer_base.py
```diff
@@ -114,6 +99,17 @@ def __init__(
...
+        # Auto-Reset
+        self._should_reset = False
+        self._on_episode_done: Optional[DoneHandlerType] = None
+
+        # Buffer Injection - we inject buffers into MettaGridCore after the first reset
+        self._should_inject_buffers = True
+
+    def set_on_episode_done(self, handler: DoneHandlerType) -> None:
+        """Set the callback function to be called when an episode completes."""
+        self._on_episode_done = handler
```
berekuk, line 111: In the future, I'd like us to go one step further and do something similar to DOM event handler APIs:
- have more kinds of events, wherever the performance is not prohibitive
- use lists of handlers for each event type: `add_on_episode_done_handler(handler: DoneHandlerType) -> HandlerId`, `remove_on_episode_done_handler(handler_id: HandlerId)`

metta/rl/trainer_config.py
```diff
@@ -190,6 +193,7 @@ class TrainerConfig(Config):
     env: str | None = None  # Environment config path
     # Default curriculum: Simple environment for initial experiments
     curriculum: str | None = "/env/mettagrid/curriculum/simple"
+    new_curriculum: CurriculumConfig = env_curriculum(arena_builder.arena(num_agents=24))
```
berekuk, line 196: I assume this is WIP and the final version will have `curriculum: CurriculumConfig`

metta/rl/trainer_config.py
teodorionita: fixed now

mettagrid/src/metta/mettagrid/mettagrid_config.py
```diff
@@ -214,7 +215,7 @@ class GameConfig(Config):
     groups: dict[str, GroupConfig] = Field(
         default_factory=lambda: {"agent": GroupConfig(id=0, sprite=0, props=AgentConfig())}, min_length=1
     )
-    actions: ActionsConfig = Field(default_factory=ActionsConfig)
+    actions: ActionsConfig = ActionsConfig(noop=ActionConfig())
```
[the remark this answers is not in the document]
teodorionita: yep should be correct and reverted here

mettagrid/src/metta/mettagrid/mettagrid_config.py
teodorionita: we can use `MapBuilderConfig` i'll add a new PR

PR #2088, opened 2025-08-13
Add `experiments` system for launching repeatable experiments

jacklionheart:
This is a WIP, but it creates a way to define experiments (sets of training jobs) in a way that you can repeatable launch them, as well as introduced wandb and skypilot "services" that are meant to be the backbone of notebooks/marimo/plotly analysis tools. 

This also introduces a new path for actually launching jobs where the flow is:

* locally, in python, we create config objects, and then serialize them to yaml
* we generate a fully specified yaml from the python configs and then pass that as the input to the skypilot job, rather than e.g. forwarding commandline configs

because agent and curricula are not currently actually config objects, there is a bunch of hackiness to handle that for now that we can clean up as those imminent changes come in.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211057828509828)


jacklionheart:
reassigning to daveey because he is concerned that this takes things in a different direction than his big outstanding changes

PR #2103, opened 2025-08-14
Better postgres connection pool handling

Barraketh:
Better connection pool error recovery and logging

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211051514430642)


berekuk (approved):
Unsure if this will fix the root cause, but looks like an improvement.

Have you considered exposing metrics from the server? We could collect pool stats there and chart them in datadog, then it'll be more clear what's going on.

PR #2111, opened 2025-08-14
Add AgentConfig (Agent Specification)

harshbhatt7585:
This PR introduces a new configuration class, AgentConfig, which serves as the agent specification (similar to TrainerConfig).

Key changes:

Added AgentConfig to define the complete agent architecture and related parameters in a structured, compact form.

Enabled AgentConfig to generate detailed component configurations for building the agent’s policy.

This improves code organization by separating agent structure/configuration from its runtime instantiation, making it easier to modify or reuse agent setups.

### DIFF
--- a/agent/src/metta/agent/agent_config.py
+++ b/agent/src/metta/agent/agent_config.py
@@ -1,0 +1,414 @@
+from typing import Any, ClassVar, Literal, Optional
+
+from omegaconf import DictConfig, OmegaConf
+from pydantic import ConfigDict, Field, model_validator
+
+from metta.common.util.config import Config
+
+
+class ObservationConfig(Config):
+    """Configuration for observation processing."""
+
+    obs_key: str = Field(default="grid_obs", description="Key for observation data")
+
+    # Observation encoder type
+    encoder_type: Literal["cnn", "attention", "latent_attention"] = Field(
+        default="cnn", description="Type of observation encoder"
+    )
+
+    # CNN-specific parameters
+    cnn_channels: list[int] = Field(
+        default=[64, 64], description="Number of channels for each CNN layer"
+    )
+    cnn_kernel_sizes: list[int] = Field(
+        default=[5, 3], description="Kernel sizes for CNN layers"
+    )
+    cnn_strides: list[int] = Field(
+        default=[3, 1], description="Strides for CNN layers"
+    )
+
+    # Attention-specific parameters
+    attention_dim: int = Field(
+        default=128, description="Dimension for attention layers"
+    )
+    attention_heads: int = Field(
+        default=4, description="Number of attention heads"
+    )
+    attention_layers: int = Field(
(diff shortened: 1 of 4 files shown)


jacklionheart:
Reassigning to Richard who will explain better than I can what we want this to look like

agent/src/metta/agent/agent_config.py
```diff
@@ -0,0 +1,414 @@
...
+    """Configuration for observation processing."""
+
+    obs_key: str = Field(default="grid_obs", description="Key for observation data")
+
+    # Observation encoder type
+    encoder_type: Literal["cnn", "attention", "latent_attention"] = Field(
+        default="cnn", description="Type of observation encoder"
+    )
+
+    # CNN-specific parameters
```
relh, line 19: this is more a part of the network and not a part of the observation.

agent/src/metta/agent/agent_config.py
```diff
@@ -0,0 +1,414 @@
...
+        default=[64, 64], description="Number of channels for each CNN layer"
+    )
+    cnn_kernel_sizes: list[int] = Field(
+        default=[5, 3], description="Kernel sizes for CNN layers"
+    )
+    cnn_strides: list[int] = Field(
+        default=[3, 1], description="Strides for CNN layers"
+    )
+
+    # Attention-specific parameters
```
relh, line 30: same as above

agent/src/metta/agent/agent_config.py
```diff
@@ -0,0 +1,414 @@
...
+        default=128, description="Dimension for attention layers"
+    )
+    attention_heads: int = Field(
+        default=4, description="Number of attention heads"
+    )
+    attention_layers: int = Field(
+        default=2, description="Number of attention layers"
+    )
+
+    # Latent attention parameters
```
relh, line 41: same as above

agent/src/metta/agent/agent_config.py
```diff
@@ -0,0 +1,414 @@
...
+    # Fourier feature parameters
+    fourier_freqs: int = Field(
+        default=8, description="Number of Fourier frequencies"
+    )
+    attr_embed_dim: int = Field(
+        default=8, description="Attribute embedding dimension"
+    )
+
+
+class CoreConfig(Config):
```
relh, line 58: doesn't need to be its own config if it wraps 2 things

agent/src/metta/agent/agent_config.py
```diff
@@ -0,0 +1,414 @@
...
+    )
+    effective_rank: bool = Field(
+        default=True, description="Use effective rank regularization"
+    )
+
+
+class ActorConfig(Config):
+    """Configuration for the actor network."""
+
+    hidden_size: int = Field(
```
relh, line 86: having a config that wraps one field is an antipattern

PR #2117, opened 2025-08-14
make both notebookwidgets use pnpm and turbo

zfogg:
also added jq to the brewfile because i use it with turbo to check cache status to rebuild a notebook widget or not. deleted package-lock.json from both projects and will just remember to use turbo and pnpm from now on :)


experiments/notebooks/utils/eval_finder_widget/turbo.jsonc
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "extends": ["//"],
+  "tasks": {
+    "build": {
+      "outputs": ["eval_finder_widget/static/**"]
+    },
+    // common linting and formatting tasks
+    "lint": {},
+    "lint:eslint": {}
```
berekuk, line 10: `lint` part is not necessary, these are inherited from the top-level config

PR #2121, opened 2025-08-14
[MettaScope] Font sprite atlas + rendering + kerning

monofuel:
- add font sprites to our atlas so we can render text in the game world
- add metadata for all the font stuff like the advance, bounds, ascent/decent, kerning mapping

Screenshot From 2025-08-14 18-08-41

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211060462041878)


mettascope/tools/gen_atlas.py
```diff
@@ -23,6 +25,35 @@
...
+def _get_font_config():
+    """Return the current font-related configuration that should trigger rebuilds when changed."""
+    cfg = {
+        "FONT_ID": FONT_ID,
+        "FONT_PATH": FONT_PATH,
+        "FONT_SIZE": FONT_SIZE,
+        "FONT_CHARSET": FONT_CHARSET,
+        "GLYPH_INNER_PADDING": GLYPH_INNER_PADDING,
+    }
+    if os.path.exists(FONT_PATH):
```
monofuel, line 44: these extra checks on the font file itself feel excessive, might be worth removing? or just keep?

treeform: We will want to have several fonts, please pass them as params to the create font function not a whole config.

mettascope/tools/gen_atlas.py
```diff
@@ -120,6 +163,100 @@ def put_image(img, name):
     return images[name]
 
 
+def char_label(cp: int) -> str:
+    """Convert a character code to a safe to use label string."""
+    return f"U+{cp:04X}"
+
+
+def generate_font_glyphs():
```
treeform, line 171: Just have the config for the font passed here.

mettascope/tools/gen_atlas.py
```diff
@@ -162,8 +301,15 @@ def main():
 
     # Write the atlas image and the atlas json file.
     try:
+        font_cfg = _get_font_config()
+        atlas_out = {
+            "images": dict(images),
+            "fonts": fonts_meta,
+            "fontConfig": font_cfg,
```
treeform, line 308: We don't want this just make it part of font

mettascope/tools/gen_atlas.py
```diff
@@ -120,6 +163,100 @@ def put_image(img, name):
...
+            adjust = float(typeface.get_kerning_adjustment(left, right)) * scale
+            if adjust != 0.0:
+                if row is None:
+                    row = {}
+                row[char_label(right)] = adjust
+        if row:
+            kerning[left_label] = row
+
+    fonts_meta[FONT_ID] = {
+        "ascent": float(ascent_px),
```
treeform, line 249: just add hash and other config param here, not in the main atlas.

mettascope/tools/gen_atlas.py
treeform: this will not work with multiple fonts

mettascope/src/context3d.ts
```diff
@@ -38,9 +38,40 @@ const FRAGMENT_SHADER_SOURCE = `
...
+interface FontMeta {
+  ascent: number
+  descent: number
+  lineHeight: number
+  glyphs: { [label: string]: FontGlyph }
+  kerning: { [leftLabel: string]: FontKerningRow }
+}
+
+interface FontConfig {
+  FONT_ID: string
```
treeform, line 62: the upper case here is really ugly

mettascope/src/context3d.ts
```diff
@@ -38,9 +38,40 @@ const FRAGMENT_SHADER_SOURCE = `
...
+  bearingX: number
+  bearingY: number
+}
+
+interface FontKerningRow {
+  [rightLabel: string]: number
+}
+
+interface FontMeta {
+  ascent: number
```
treeform, line 54: font config and font metta should be same thing? I would just call It Font

mettascope/src/context3d.ts
```diff
@@ -38,9 +38,40 @@ const FRAGMENT_SHADER_SOURCE = `
...
+  descent: number
+  lineHeight: number
+  glyphs: { [label: string]: FontGlyph }
+  kerning: { [leftLabel: string]: FontKerningRow }
```
treeform, line 58: The type for this should be simpler why do we need an object to store a value?

mettascope/src/context3d.ts
```diff
@@ -38,9 +38,40 @@ const FRAGMENT_SHADER_SOURCE = `
   }
 `
 
+// Font atlas metadata types
+interface FontGlyph {
```
treeform, line 42: Just call it Glyph

mettascope/src/context3d.ts
```diff
@@ -38,9 +38,39 @@ const FRAGMENT_SHADER_SOURCE = `
...
+}
+
+interface Font {
+  ascent: number
+  descent: number
+  lineHeight: number
+  glyphs: { [label: string]: Glyph }
+  kerning: { [leftLabel: string]: FontKerningRow }
+  // Inlined former FontConfig fields and hash
+  FontID: string
```
treeform, line 60: please just use normal case fontId

mettascope/src/context3d.ts
```diff
@@ -38,9 +38,39 @@ const FRAGMENT_SHADER_SOURCE = `
...
+  FontPathMTime: number | null
+  FontPathSize: number | null
+  FontConfigHash: string
+}
+
 /** Type definition for atlas data. */
 interface AtlasData {
-  [key: string]: [number, number, number, number] // [x, y, width, height]
+  images: { [key: string]: [number, number, number, number] }
+  fonts: { [fontId: string]: Font }
```
treeform, line 73: it should be font name not font id? Id kind of implies a number?

mettascope/src/context3d.ts
```diff
@@ -571,6 +604,119 @@ export class Context3d {
...
+      const u1 = (sx + sw + mBase) / this.textureSize.x()
+      const v1 = (sy + sh + mBase) / this.textureSize.y()
+
+      // Position the glyph image so that its baseline aligns at (penX, penY).
+      const drawX = penX + glyph.bearingX - glyphInnerPadding - m
+      const drawY = penY + glyph.bearingY - glyphInnerPadding - m
+      const drawW = sw + 2 * m
+      const drawH = sh + 2 * m
+
+      // Outline pass: draw the glyph tinted black around the main position for readability.
```
treeform, line 693: This feels bad to me. Maybe just generate them with a border already in pixie? We wound need infinite offsets to do it right this way?

mettascope/tools/gen_atlas.py
```diff
@@ -23,6 +25,39 @@
 images = {}
 heights = [0] * atlas_image.width
 padding = 64
+FONTS = [
+    {
+        "FontID": "plexSans",
```
treeform, line 30: can you use normal case here?

mettascope/tools/gen_atlas.py
```diff
@@ -23,6 +25,39 @@
...
+        "FontPath": "data/fonts/IBMPlexSans-Regular.ttf",
+        "FontSize": 64,
+        "FontCharset": " 0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz",
+        "GlyphInnerPadding": 2,
+    }
+]
+
+
+def _get_font_config(font: dict):
```
treeform, line 39: this feels so complex with the hashes...
maybe just check the hash of this file and the font file? Is that not enough? why store and read hash back?

treeform (changes requested):

mettascope/tools/gen_atlas.py
monofuel: we could hash this file, but we would still need a place to store the hash to check when it changes

mettascope/src/context3d.ts
monofuel: Yea, i was wondering that. I think this method in the UI would work better font scaling? if we do it with pixie then the borders may end up too big or too thin depending on the zoom level.

monofuel: Screenshot From 2025-08-15 16-01-47

did some testing with different zoom levels, doing it with pixie should be fine.

Updated to do the border with pixie

PR #2122, opened 2025-08-14
8way move restricted agents

gregsoftmax:
Two things:
- agent orientation was being updated after rejecting a movement for the target location having an object, making the movement not be able to allow for rotations. Now moving into an occupied target location updates your orientation in that direction. 
- adding agent_interference flag to move_8way


relh (approved):

relh:
Hang tight I'm correcting the test we do want to change orientation if we attempt to move but cant

PR #2123, opened 2025-08-14
compile-time optimized no_agent_interference

relh:


relh:
Probs not worth @gregsoftmax

PR #2144, opened 2025-08-15
Support replay format v2 in thumbnail generation tool

lcodes:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211068100180246)

### DIFF
--- a/mettascope/tools/gen_thumb.py
+++ b/mettascope/tools/gen_thumb.py
@@ -52,17 +52,44 @@ def get_position_component(object, step, component):
     return result
 
 
-def read_replay_map(input, step):
-    if input["version"] != 1:
-        raise ValueError("Unsupported replay version")
+def get_location(object, step):
+    x = object["location"]
+    if not isinstance(x[1], list):
+        return x
+    result = [0, 0, 0]
+    for [frame, value] in x:
+        if frame > step:
+            break
+        result = value
+    return result
+
+
+def read_replay_map(input, step, debug):
     if input["max_steps"] <= step:
         raise ValueError("Step is out of range")
+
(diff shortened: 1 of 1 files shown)


mettascope/tools/gen_thumb.py
```diff
@@ -52,17 +52,44 @@ def get_position_component(object, step, component):
...
+    return result
+
+
+def read_replay_map(input, step, debug):
     if input["max_steps"] <= step:
         raise ValueError("Step is out of range")
 
+    version = input["version"]
+    match version:
+        case 1:
```
treeform, line 73: I think it would make code simpler if we just don't support the old format. Can you just remove it?

mettascope/tools/gen_thumb.py
```diff
@@ -77,12 +104,17 @@ def read_replay_map(input, step):
...
 
-    objects = input["grid_objects"]
+    objects = input[objects_key]
     nodes = [0] * len(objects)
     for i, object in enumerate(objects):
-        x = get_position_component(object, step, "c")
-        y = get_position_component(object, step, "r")
-        nodes[i] = y | (x << 16) | (object["type"] << 32) | (object.get("agent_id", 0) << 48)
+        match version:
+            case 1:
```
treeform, line 111: no need for v1

treeform (changes requested):
I think the view ranges is now also a field.

mettascope/tools/gen_thumb.py
treeform: We are going to generate thumps only for new replay when we implement it.

PR #2149, opened 2025-08-15
Component Pytorch Policies

harshbhatt7585:
This PR implement component pytorch policies.

### DIFF
--- a/agent/src/metta/agent/component_policies/fast.py
+++ b/agent/src/metta/agent/component_policies/fast.py
@@ -1,0 +1,405 @@
+import logging
+from typing import Optional, Union
+
+import gymnasium as gym
+import torch
+from tensordict import TensorDict
+from torch import nn
+from omegaconf import DictConfig, OmegaConf
+from torch.nn.modules import padding
+
+from metta.agent.util.distribution_utils import evaluate_actions, sample_actions
+from metta.agent.util.safe_get import safe_get_from_obs_space
+from metta.agent.util.debug import assert_shape
+from metta.common.util.datastruct import duplicates
+
+from metta.agent.lib.nn_layer_library import Conv2d, Linear, Flatten
+from metta.agent.lib.lstm import LSTM
+from metta.agent.lib.action import ActionEmbedding
+from metta.agent.lib.actor import MettaActorSingleHead
+from metta.agent.lib.observation_normalizer import ObservationNormalizer
+from metta.agent.lib.obs_token_to_box_shaper import ObsTokenToBoxShaper
+
+logger = logging.getLogger(__name__)
+
+
+class Fast(nn.Module):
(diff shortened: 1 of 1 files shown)


agent/src/metta/agent/component_policies/component_policy.py
```diff
@@ -29,99 +28,104 @@ def __init__(
...
         # Initialize action conversion tensors (will be set by MettaAgent)
         self.cum_action_max_params = None
         self.action_index_tensor = None
 
+        logger.info(f"{self.__class__.__name__} policy initialized with components: {list(self.components.keys())}")
+
+    def _build_components(self) -> dict:
+        """Build the component dictionary. Override this in subclasses to define architecture."""
+        # Default implementation returns empty dict - subclasses should override
+        return {}
```
harshbhatt7585, line 104: I want this be an abstract class that's why I also inherited ABC, because this function needs to be implemented and should be defined as @abstract method.

relh: fix incoming

agent/src/metta/agent/component_policies/component_policy.py
```diff
@@ -29,99 +29,104 @@ def __init__(
...
-        # Validate and extract observation key
-        if not (hasattr(self.cfg.observations, "obs_key") and self.cfg.observations.obs_key is not None):
-            raise ValueError("Configuration missing required field 'observations.obs_key'")
+        # Device setup
+        self.device = torch.device(device) if device else torch.device("cuda" if torch.cuda.is_available() else "cpu")
 
-        obs_key = self.cfg.observations.obs_key
-        obs_shape = safe_get_from_obs_space(obs_space, obs_key, "shape")
+        # Validate and extract observation key
+        obs_key = "grid_obs"
```
relh, line 41: fixing this to not be hard-coded and/or to see if we can just remove it

agent/src/metta/agent/component_policies/component_policy.py
```diff
@@ -29,99 +29,104 @@ def __init__(
...
+        # Store config parameters
+        self.config = config or {}
+        self.clip_range = self.config.get("clip_range", 0)
+        self.analyze_weights_interval = self.config.get("analyze_weights_interval", 300)
 
-        # Validate and extract observation key
-        if not (hasattr(self.cfg.observations, "obs_key") and self.cfg.observations.obs_key is not None):
-            raise ValueError("Configuration missing required field 'observations.obs_key'")
+        # Device setup
+        self.device = torch.device(device) if device else torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
relh: removing this because metta agent sets it

relh (approved):

PR #2188, opened 2025-08-18
Give skypilot workers access to datadog

nishu-builder:
Because skypilot workers aren't managed by kubernetes, we need to separately provide datadog configuration env vars and install datadog agent. We could just use the python sdk without datadog agent, but then we don't get things like log aggregation and system perf monitoring out of the box, so I think on net this is simpler

./devops/skypilot/launch.py checks for DD_API_KEY in environment, and if present, installs datadog agent

metrics will appear in Datadog tagged with:
  - env:production
  - service:skypilot-worker
  - run_id: run-id
  - node_rank: (0,1,2...)

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211081598907943)


devops/skypilot/config/sk_setup.sh
```diff
@@ -10,3 +10,52 @@ echo "[SETUP] Checked out: $(git rev-parse HEAD)"
 
 # Note that different sets of skypilot environment variables are available in "run" vs "setup"
 # see https://docs.skypilot.co/en/latest/running-jobs/environment-variables.html
+
+# Install Datadog Agent if DD_API_KEY is provided
+if [ -n "${DD_API_KEY:-}" ]; then
+    echo "[SETUP] Installing Datadog Agent..."
+
+    # Install Datadog Agent using the official script
+    DD_AGENT_MAJOR_VERSION=7 DD_API_KEY="$DD_API_KEY" DD_SITE="${DD_SITE:-datadoghq.com}" bash -c "$(curl -L https://s3.amazonaws.com/dd-agent/scripts/install_script.sh)"
```
berekuk, line 19: chatgpt says that s3 link is outdated and the right one should point at https://install.datadoghq.com/..., see https://app.datadoghq.com/fleet/install-agent/latest?platform=linux

devops/skypilot/config/sk_setup.sh
```diff
@@ -10,3 +10,52 @@ echo "[SETUP] Checked out: $(git rev-parse HEAD)"
...
+
+    # Configure Datadog to collect Docker metrics
+    if [ -f /etc/datadog-agent/datadog.yaml ]; then
+        echo "[SETUP] Configuring Datadog for Docker monitoring..."
+
+        # Enable Docker integration
+        sudo sed -i 's/# process_config:/process_config:/' /etc/datadog-agent/datadog.yaml
+        sudo sed -i '/process_config:/a\  enabled: true' /etc/datadog-agent/datadog.yaml
```
berekuk, line 27: Another tip from chatgpt, I think it's correct: you can write these as env vars to `/etc/datadog-agent/environment` instead of patching with sed, see https://docs.datadoghq.com/agent/guide/environment-variables/#using-environment-variables-in-systemd-units

berekuk (approved):
lgtm

PR #2206, opened 2025-08-19
Upload eval logs to s3

Barraketh:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211086651855660)

### DIFF
--- a/app_backend/pyproject.toml
+++ b/app_backend/pyproject.toml
@@ -13,12 +13,15 @@ readme = "README.md"
 urls = { Homepage = "https://daveey.github.io", Repository = "https://github.com/Metta-AI/metta" }
 
 dependencies = [
+  "boto3>=1.38.32",
+  "ddtrace>=2.6.0",
   "fastapi>=0.115.5",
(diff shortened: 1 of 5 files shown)


Barraketh:
> Do we just want to use datadog log aggregation for this?

I don't mind also pushing these to datadog, but I also want these to be accessible from the UI so that we can easily debug what was going on in a particular eval run

app_backend/pyproject.toml
```diff
@@ -13,12 +13,15 @@ readme = "README.md"
 urls = { Homepage = "https://daveey.github.io", Repository = "https://github.com/Metta-AI/metta" }
 
 dependencies = [
+  "boto3>=1.38.32",
```
nishu-builder, line 16: I'm not entirely sure how this works but should metta-common deps be automatically adopted into app_backend deps?

nishu-builder (approved):
I am super happy doing this for now, and if later we find it's duplicative with what's in datadog, we could remove it

PR #2245, opened 2025-08-20
HRM Policy

harshbhatt7585:
Implements HRM based Policy

### DIFF
--- a/agent/src/metta/agent/pytorch/hrm.py
+++ b/agent/src/metta/agent/pytorch/hrm.py
@@ -1,0 +1,15 @@
+import torch
+from torch import nn
+
+
+class ReasoningBlock(nn.Module):
+    def __init__(self, layers: List[ReasoningAttnBlock]):
+        super().__init__()
+
+        self.layers = torch.nn.ModuleList(layers)
+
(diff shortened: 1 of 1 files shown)


relh (approved):
This looks great! Can you put the arena_hrm.py recipe into a experiments/recipes/arena/hrm.py file instead?

agent/src/metta/agent/policies/hrm.py
```diff
@@ -0,0 +1,163 @@
...
+            out_key="logits",
+            embed_dim=_embedding_dim,
+        ),
+    ]
+
+    action_probs_config: ActionProbsConfig = ActionProbsConfig(in_key="logits")
+
+
+# Default HRM configuration
+HRMDefaultConfig = HRMPolicyConfig
```
relh: shouldn't need this alias

harshbhatt7585:
> This looks great! Can you put the arena_hrm.py recipe into a experiments/recipes/arena/hrm.py file instead?

Right now we have arena.py having arena as a directory won't gonna work, it will directly lead to arena.py, we need to think any other directory name except arena.

PR #2273, opened 2025-08-21
cold: combine move/cardinal/8way -> move.. but then became hot :'(

relh:
ORIGINAL COLD:
JackE found that 8way and Cardinal learnt faster than tank so as cooling we're simplifying our movement to 8way.

8way is good, because if an agent learns about spatial positioning it can use the Pythagorean theorem to return to objects outside of view and that shows it knows where things are.

NEW HOT:
- moved the binding of the actions to the files that have the actions in them
- linted some of the .cpp files
- pyi updates

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211110995857959)


mettagrid/src/metta/mettagrid/config/envs.py
```diff
@@ -32,7 +32,7 @@ def make_arena(
 
     actions = ActionsConfig(
         noop=ActionConfig(),
-        move=ActionConfig(),
+        move_8way=ActionConfig(),
```
rjwalters, line 35: can we just call this "move"

mettagrid/src/metta/mettagrid/util/actions.py
```diff
@@ -137,29 +132,28 @@ def move(env: MettaGrid, orientation: Orientation, agent_idx: int = 0) -> Dict[s
...
+        result["error"] = "8-way movement not available"
         return result
 
-    move_action_idx = action_names.index("move")
-    rotate_action_idx = action_names.index("rotate")
+    move_8way_idx = action_names.index("move_8way")
+
+    # Map orientations to 8-way movement indices
+    # 0=N, 1=NE, 2=E, 3=SE, 4=S, 5=SW, 6=W, 7=NW
+    orientation_to_8way = {
```
rjwalters, line 144: I think it would be better to change the Orientation helper class to "Compass" and change to N/NE/E/SE/S/SW/W etc

mettagrid/tests/test_rewards.py
```diff
@@ -109,10 +107,8 @@ def collect_heart_from_altar(env):
         if not move_result["success"]:
             return False, 0.0
 
-    # Rotate to face right (towards altar at (1,3))
-    rotate_result = rotate(env, Orientation.RIGHT, agent_idx=0)
-    if not rotate_result["success"]:
-        return False, 0.0
+    # With move_8way, we don't need to rotate - just collect directly
+    # The get_items action will work from any adjacent position
```
rjwalters, line 111: this seems like a second change that deserves more conversation -- for example, do we still rotate for attack?

I think we should postpone changing the meaning of rotation for now

bluewin4:
To be clear, 8 way did better because it moved around faster

Cardinal movement learned better and had better memory in general than 8-way or the tank movement.

bluewin4:
So after our conversation we are on-board with the cardinal as our main movement with an option that add diagonal movement/facing as well, correct?

relh:
Types of Right Angle Triangle[750 × 429](https://www.google.com/url?sa=i&url=https%3A%2F%2Fbyjus.com%2Fmaths%2Fright-angle-triangle%2F&psig=AOvVaw3iJGMt03-wtmzq9DcqjLW4&ust=1756237982172000&source=images&cd=vfe&opi=89978449&ved=0CBYQjRxqFwoTCNiJmbrepo8DFQAAAAAdAAAAABAL)

I think if we see an agent go from X to Y to Z and then back to X along the diagonal I think it will be really interesting

relh:
image

I think if we see an agent go from X to Y to Z and then back to X along the diagonal I think it will be really interesting

rjwalters (approved):

PR #2281, opened 2025-08-21
library tf stack

berekuk:


berekuk:
* **#2281** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2281?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2281?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

devops/tf/library/outputs.tf
```diff
@@ -0,0 +1,7 @@
+output "postgres_endpoint" {
+  value = aws_db_instance.postgres.endpoint
+}
+
+output "postgres_url" {
+  value = nonsensitive(local.postgres_url)
```
berekuk, line 6: `nonsensitive` is intentional because I needed to obtain the postgres url for github secret somehow. "Nonsensitive" just means that spacelift shows the url in its web UI.
 
A better solution would be to create github secret _from_ spacelift, but then I'd need to configure github provider (copy github pat from github stack or configure a reusable spacelift context, etc), and we're mostly not terraforming github secrets yet.

devops/tf/library/db.tf
```diff
@@ -0,0 +1,47 @@
...
+  identifier     = "softmax-library-pg"
+  engine         = "postgres"
+  engine_version = var.db_postgres_version
+
+  instance_class    = var.db_instance_class
+  allocated_storage = var.db_allocated_storage
+  multi_az          = true
+
+  publicly_accessible    = true
+  vpc_security_group_ids = [aws_security_group.rds_public.id]
```
berekuk, line 35: Public, even though LLMs say it's not the best practice, for two reasons:
1) Amplify can't access RDS and there are no security groups that would allow Amplify -> RDS permission; LLMs recommend a lambda proxy but that's annoying to do
2) We need to access it from the github migration workflow

(in my previous experience I often used postgres endpoints that can be accessed from the internet so I think LLMs are overcautious about this)

amplify.yml
```diff
@@ -0,0 +1,33 @@
+version: 1
+applications:
+  - appRoot: library
+    frontend:
+      phases:
+        preBuild:
+          commands:
+            - nvm install 22
+            - nvm use 22
```
berekuk, line 9: Amplify doesn't support node v24 yet, I tried :(

morgancmartin (approved):
LGTM

PR #2282, opened 2025-08-21
[MettaScope] fix CI tests

monofuel:
- try to get the CI tests going
- so far I've just brute forced all the args? but we should probably have a proper ci config somewhere?

### DIFF
--- a/.github/workflows/test-mettascope.yml
+++ b/.github/workflows/test-mettascope.yml
@@ -8,7 +8,6 @@ on:
 
 jobs:
   check-container:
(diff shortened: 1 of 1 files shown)


common/src/metta/common/config/tool.py
```diff
@@ -20,6 +20,8 @@ class Tool(Config):
 
     system: SystemConfig = Field(default_factory=SystemConfig)
 
+    consumed_args: list[str] = Field(default_factory=list)
```
monofuel, line 23: This also fixes "play metta" and "generate replay" in vscode tasks

PR #2298, opened 2025-08-22
CI optimizations

berekuk:
- new runner names
- run tests without shards on large runners (4 cpus)
- run non-critical jobs on free ubuntu-latest runners
- removed `mettagrid_build.py`
- switched cmake to ninja to auto-detect cpus (we were building mettagrid C++ in a single thread previously)
- removed smoke tests which were broken and disabled for a long time now; after dehydration smoke tests are even more broken and I think it's enough to have unit tests (which we can now make for tools) and the larger skypilot test

This PR still runs each project in its own job, it's possible we could run everything in a single job but since we want to add `uv run --exact` eventually it might get tricky, so for now it's still a matrix.

Total `checks.yml` time: under 4 minutes. (previously: 6-7 minutes when we're lucky)

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211127048355668)


berekuk:
* **#2298** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2298?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2298?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/actions/disk-cleanup/README.md
berekuk, line 1: Removed again because we can (and do) just ask for larger runners now.

.github/workflows/build-metta-fe-docker.yml
```diff
@@ -17,7 +17,7 @@ env:
 
 jobs:
   build-and-push-image:
-    runs-on: default
+    runs-on: ubuntu-4core
```
berekuk, line 20: I used 4core machines for all docker builders, 16core for more cpu-heavy tests, and `ubuntu-latest` (free) for everything else that's not on the critical path.

berekuk: Hmm maybe I'll tune it down from 16core, I'm trying to estimate how expensive it'll be.

berekuk: I tried a few versions and settled on 4core for now (for unit tests and for C++). Still under 4 minutes, possibly because pytest didn't detect workers correctly, but I think 4 minutes is good enough.

PR #2302, opened 2025-08-23
MettaScope 2 in Nim

treeform:
This is just a start but many parts work already:
* Both native and web works (emscripten)
* Loading replay works
* Playing the replay works
* Moving and zooming works

image

I just want to commit this so that we have something to work from.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211165567334289)


mettascope2/src/mettascope/worldmap.nim
```diff
@@ -0,0 +1,347 @@
...
+              hasWall(x + 1, y - 1):
+            continue
+        bxy.drawImage(wallSprites[tile], vec2(x.float32, y.float32), angle = 0, scale = 1/200)
+
+  for fillPos in wallFills:
+    bxy.drawImage("objects/wall.fill", fillPos.vec2 + vec2(0.5, 0.3), angle = 0, scale = 1/200)
+
+proc drawObjects*() =
+  ## Draw the objects on the map.
+  for x in 0 ..< MapWidth:
```
lcodes, line 97: that's a large scan when w*h is much larger than numobjects

the switch on thing.kind is also a branch prediction trasher

if objects can be bucketed by kind permanently, that'd be ideal

mettascope2/src/mettascope/worldmap.nim
```diff
@@ -0,0 +1,347 @@
...
+          gridPos = (agent.pos + ivec2(i - ObservationWidth div 2, j - ObservationHeight div 2))
+
+        if gridPos.x >= 0 and gridPos.x < MapWidth and
+           gridPos.y >= 0 and gridPos.y < MapHeight:
+          visibility[gridPos.x][gridPos.y] = true
+
+  for x in 0 ..< MapWidth:
+    for y in 0 ..< MapHeight:
+      if not visibility[x][y]:
+        bxy.drawRect(
```
lcodes, line 183: shader idea;
- vertex quad, nothing fancy there (common vert for all effect shaders, which 2d ends up being)
- visibility specific fragment; upload vis as a bitmap (1bit/cell, xy indexing, point mip)

one draw for the whole visibility, shader determines which cell the pixel is in, looks up vis in the texture, and shades accordingly -> also need the right blend mode but thats a simple additive

mettascope2/src/mettascope/worldmap.nim
```diff
@@ -0,0 +1,347 @@
...
+proc drawAgentDecorations*() =
+  # Draw energy bars, shield and frozen status.
+  for agent in env.agents:
+    # if agent.shield:
+    #   bxy.drawImage(
+    #     "shield",
+    #     agent.pos.vec2 * 64,
+    #     angle = 0
+    #   )
+    if agent.frozen > 0:
```
lcodes, line 265: DoD helps here; separate frozen[] as a bitarray from agents, thats the only thing needed here (and a position[] too), everything else is extra cache line data being moved without using it

can even have the frozen events maintain a subview of just the agent-ids being frozen,
so a scan knows exactly what to pick; but that begins to ask for systems to automate this

mettascope2/src/mettascope/utils.nim
```diff
@@ -0,0 +1,59 @@
...
+  color: Color
+) =
+  ## Draw text on the screen.
+  var font = newFont(typeface)
+  font.size = size
+  font.paint = color
+  let
+    arrangement = typeset(@[newSpan(text, font)], bounds = vec2(1280, 800))
+    globalBounds = arrangement.computeBounds(transform).snapToPixels()
+    textImage = newImage(globalBounds.w.int, globalBounds.h.int)
```
lcodes, line 25: This seems costly, no?

mettascope2/src/mettascope/sim.nim
```diff
@@ -0,0 +1,817 @@
...
+    AltarHpLayer = 22
+    AltarReadyLayer = 23
+
+  Orientation* = enum
+    N # Up, Key W
+    S # Down, Key S
+    W # Right, Key D
+    E # Left, Key A
+
+  ThingKind* = enum
```
lcodes, line 92: How about archetypes?

Thing -> pos, id, layer
Owner -> input/output resource, energy, cooldown
Agent -> id, orientation, energy, ...

then a wall is "Thing" and a generator is "Thing+Owner" and so is a converter and agent is "Thing+Owner+Agent"

then we have a block that contains Thing[], a block that contains Thing[],Owner[], and a block that contains Thing[],Owner[],Agent[], and now we can start to make generic drawing code

goal is to optimize traversals during the frame's hot path; minimize branching, optimize caches

mettascope2/src/mettascope/sim.nim
```diff
@@ -0,0 +1,817 @@
...
+  if argument == 0:
+    newPos -= orientationToVec(agent.orientation)
+  elif argument == 1:
+    newPos += orientationToVec(agent.orientation)
+  else:
+    inc env.stats[id].actionInvalid
+    return
+  if env.isEmpty(newPos):
+    env.grid[agent.pos.x][agent.pos.y] = nil
+    env.updateObservations(AgentLayer, agent.pos, 0)
```
lcodes, line 371: Those are going to scan a lot of agents per updateObservations() call when only the layer change is relevant, could invert how to go at it and observe all layers at once

mettascope2/src/mettascope/sim.nim
```diff
@@ -0,0 +1,817 @@
...
+    inc env.stats[id].actionInvalid
+
+proc rotateAction(env: Environment, id: int, agent: Thing, argument: int) =
+  ## Rotate the agent
+  if argument < 0 or argument > 3:
+    inc env.stats[id].actionInvalid
+    return
+  agent.orientation = Orientation(argument)
+  env.updateObservations(AgentOrientationLayer, agent.pos, argument)
+  inc env.stats[id].actionRotate
```
lcodes, line 407: concrete stats sampling end up mixing concerns and gets nasty at scale

could collect actions as data first instead of directly applying their effects

two advantages; everyone decides on actions using a stable view of the frame (no changes between actors or anything order-dependent) meaning the world is considered immutable while everyone observes it and queues their actions; then traversing that queue at the end of frame yields all the stats to increment (count each action type, add to totals; adding new types adds new stat types automatically, whole manual segment of the codebase disappears)

lcodes: in ECS that'd be what entity command buffers are for; any changes which is either structural or not on the direct path of a system's main traversal is queued for the end of that system group

mettascope2/src/mettascope/sim.nim
```diff
@@ -0,0 +1,817 @@
...
+
+proc step*(env: Environment, actions: ptr array[MapAgents, array[2, uint8]]) =
+  ## Step the environment
+  inc env.currentStep
+  for id, action in actions[]:
+    let agent = env.agents[id]
+    if agent.frozen > 0:
+      continue
+
+    case action[0]:
```
lcodes, line 718: this could be data-driven too; moves the action's impl (gameplay code) outside the simulation (engine code)

here you track a few arrays like action_functions[] action_names[] action_descriptions[] action_icons for the ui and so on, and in the gameplay file you register these

mettascope2/src/mettascope/sim.nim
```diff
@@ -0,0 +1,817 @@
...
+    of 4: env.attackAction(id, agent, action[1].int)
+    of 6: env.shieldAction(id, agent, action[1].int)
+    of 7: env.giftAction(id, agent)
+    of 8: env.swapAction(id, agent, action[1].int)
+    #of: env.jumpAction(id, agent)
+    #of: env.transferAction(id, agent)
+    else: inc env.stats[id].actionInvalid
+
+  # Update objects
+  for thing in env.things:
```
lcodes, line 732: could decouple cooldowns from objects, they're different traversals and orderings

if going for object buckets, this simplifies a ton here; each bucket is assigned a tick function

lcodes: Important part is a stable update order, e.g. if agents and converters can interact on time-sensitive state (has it finished or not?) then whether the scene says "agent, converter" in its things or "converter, agent" has different behaviors (+-1 frame offset) from a deterministic "all converters, all agents" update order

agent<->agent is more tricky since that depends on creation order in a same bucket, but here too queueing all decisions before applying them removes much of the ordering constraints

mettascope2/src/mettascope/replays.nim
```diff
@@ -0,0 +1,59 @@
...
+    groupId: int
+    agentId: int
+    location: seq[IVec3]
+    orientation: int
+    inventory: seq[seq[ItemAmount]]
+    inventoryMax: int
+    color: int
+
+    # Agent specific keys.
+    actionId: seq[int]
```
lcodes, line 22: Could be worthwhile to investigate a different structure of this
instead of "all entities -> all timelines" its "all timelines -> all entities"

so at frame 0 all entities are present and contain all their params
each frame after is just a list of deltas to apply to the previous frame

if direct jumps to late frames are slow, can store intermediate snapshots too

but one direct advantage is removing *every* seq scan for properties during a frame, perf++

mettascope2/src/mettascope/panels.nim
```diff
@@ -0,0 +1,127 @@
...
+
+  for subarea in area.areas:
+    updatePanelsSizes(subarea)
+
+proc drawFrame*(area: Area) =
+  # Draw the frame of the area.
+
+  # Draw the header ribbon background.
+  bxy.saveTransform()
+  bxy.translate(vec2(area.rect.x.float32, area.rect.y.float32))
```
lcodes, line 83: transform/translate is kinda heavy with matrix stack ops when the pos is absolute

also want to delegate combining most of this to the gpu (viewProj once per frame per camera, model matrix is just the object's transform for that frame, both batch uploaded to the gpu)

lcodes: mat4 x mat4 on the gpu is free, ton of cycles on the cpu

mettascope2/src/mettascope/common.nim
```diff
@@ -0,0 +1,70 @@
...
+
+type
+  IRect* = object
+    x*: int
+    y*: int
+    w*: int
+    h*: int
+
+  PanelType* = enum
+    WorldMap
```
lcodes, line 12: Would have a data driven list of these; always starts at a fixed 5-6 and ends up with 50+ custom panels over 3 menus :D

mettascope2/src/mettascope/actions.nim
```diff
@@ -0,0 +1,104 @@
...
+  if selection != nil and selection.kind == Agent:
+    let agent = selection
+
+    # Rotate
+    if window.buttonPressed[KeyW] or window.buttonPressed[KeyUp]:
+      if agent.orientation == N:
+        actionsArray[agent.agentId] = [1, 1]
+      else:
+        actionsArray[agent.agentId] = [2, 0]
+      simStep()
```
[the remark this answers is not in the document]
lcodes: Netcode does the same kind of user input buffering, but for different reasons, so useful :)

mettascope2/src/mettascope.nim
```diff
@@ -0,0 +1,122 @@
...
+
+# Build the atlas.
+for path in walkDirRec("data/"):
+  if path.endsWith(".png"):
+    echo path
+    bxy.addImage(path.replace("data/", "").replace(".png", ""), readImage(path))
+
+
+while not window.closeRequested:
+  display()
```
lcodes, line 121: why not poll events first? makes it clear its fetching events for the current frame

(also never know what else goes on in that loop, want those events as fresh as possible before the simulation with no delays in between)

mettascope2/src/mettascope.nim
```diff
@@ -0,0 +1,122 @@
...
+
+
+
+  bxy.endFrame()
+  window.swapBuffers()
+  inc frame
+
+
+# Build the atlas.
+for path in walkDirRec("data/"):
```
lcodes, line 114: Would push that at build time, over wasm its going to be one net request per file

also heavy startup cost vs loading a premade texture

mettascope2/src/mettascope.nim
```diff
@@ -0,0 +1,122 @@
...
+    echo "showObservations: ", settings.showObservations
+  if window.buttonPressed[KeyM]:
+    inc settings.showObservations
+    echo "showObservations: ", settings.showObservations
+  settings.showObservations = clamp(settings.showObservations, -1, 23)
+
+  let now = epochTime()
+  while play and (lastSimTime + playSpeed < now):
+    lastSimTime += playSpeed
+    simStep()
```
lcodes, line 50: if replaying, do we need to step the sim? we already know the data at every frame

(unless going for a new replay format based on snapshots & delta, which needs to replay here)

mettascope2/src/mettascope.nim
```diff
@@ -0,0 +1,122 @@
...
+  bxy.beginFrame(window.size)
+  const RibbonHeight = 64
+  rootArea.rect = IRect(x: 0, y: RibbonHeight, w: window.size.x, h: window.size.y - RibbonHeight*3)
+  rootArea.updatePanelsSizes()
+  globalHeaderPanel.rect = IRect(x: 0, y: 0, w: window.size.x, h: RibbonHeight)
+  globalFooterPanel.rect = IRect(x: 0, y: window.size.y - RibbonHeight, w: window.size.x, h: RibbonHeight)
+  globalTimelinePanel.rect = IRect(x: 0, y: window.size.y - RibbonHeight*2, w: window.size.x, h: RibbonHeight)
+
+
+  rootArea.drawFrame()
```
lcodes, line 64: with this many draws in a script i'd start thinking about data driving this too

a list of layers to begin with; if using object buckets thats an extension of that

mettascope2/src/mettascope/utils.nim
```diff
@@ -0,0 +1,59 @@
...
+  bxy: Boxy,
+  imageKey: string,
+  transform: Mat3,
+  typeface: Typeface,
+  text: string,
+  size: float32,
+  color: Color
+) =
+  ## Draw text on the screen.
+  var font = newFont(typeface)
```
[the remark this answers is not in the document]
treeform: font is a really simple object ment to be created a lot

lcodes (approved):

mettascope2/mettascope.nimble
```diff
@@ -0,0 +1,12 @@
+version     = "0.0.1"
+author      = "Softmax"
+description = "Visualization of the MettaGrid environment."
+license     = "MIT"
+
+srcDir = "src"
+
+requires "nim >= 2.2.4"
+requires "boxy >= 0.1.4"
+requires "puppy >= 2.1.2"
```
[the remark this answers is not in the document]
monofuel: no I mean it's a submodule, like a git submodule. why is it not like the other nim packages?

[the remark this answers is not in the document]
treeform: removed!

PR #2307, opened 2025-08-24
policy creation refactor

badjacent:


metta/rl/checkpoint_manager.py
```diff
@@ -43,23 +42,102 @@ def __init__(
...
+            checkpoint_config: Checkpoint configuration
             device: Training device
             is_master: Whether this is the master process
             rank: Process rank for distributed training
             run_name: Name of the current run
         """
-        self.policy_store = policy_store
+        self.policy_loader = policy_loader
         self.checkpoint_cfg = checkpoint_config
         self.device = device
```
harshbhatt7585, line 54: device is not used anywhere :)
let that sit anyway, maybe will be useful in future.

agent/src/metta/agent/policy_finder.py
```diff
@@ -0,0 +1,337 @@
...
+if TYPE_CHECKING:
+    from typing import Literal
+
+    PolicySelectorType = Literal["all", "top", "latest", "rand"]
+
+logger = logging.getLogger("policy_finder")
+
+
+class PolicyMissingError(ValueError):
+    pass
```
harshbhatt7585, line 30: Not sure why we need this? can be just used as a ValueError is much simpler and cleaner.

agent/src/metta/agent/policy_loader.py
```diff
@@ -0,0 +1,324 @@
...
+
+        Returns:
+            PolicyHandle with appropriate factory function
+        """
+        if uri.startswith("wandb://"):
+            return self._load_from_wandb_uri(uri)
+        elif uri.startswith("file://"):
+            return self.load_from_file(uri[len("file://") :])
+        else:
+            return self.load_from_file(uri)
```
harshbhatt7585, line 145: Let's simplify this by using `from urllib.parse import urlparse` . It will automatically parse `wandb` and `file`

agent/src/metta/agent/policy_finder.py
```diff
@@ -0,0 +1,337 @@
...
+            return self._prs_from_wandb(uri)
+
+        elif uri.startswith("file://"):
+            return self._prs_from_path(uri[len("file://") :])
+
+        elif uri.startswith("pytorch://"):
+            return self._prs_from_pytorch(uri)
+
+        else:
+            return self._prs_from_path(uri)
```
harshbhatt7585, line 173: this function is duplication of what has already been defined in PolicyLoader.

metta/rl/checkpoint_manager.py
relh: done

agent/src/metta/agent/policy_finder.py
relh: done

PR #2323, opened 2025-08-26
Run codebot tests as part of ci

jacklionheart:
# Factor out `gitta` library and reorganize codebot package and create gitlib utility package

Goals:
* Make sure `codebot` and `codeclip` tests run as part of CI
* Make it easy to publish codebot as an independent repo
* Shrink metta.common to be really things specific to training on mettagrid

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211154682136161)


codebot/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Codebot tools for AI development."""
```
nishu-builder, line 1: if we want codebot to be a [namespace package](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) (which the others are), i think we dont want the __init__.py

gitlib/src/gitlib/__init__.py
```diff
@@ -0,0 +1,77 @@
+"""Git utilities library for Metta projects."""
```
nishu-builder, line 1: i dont love the patttern of adding __init__ files from which other things import; adds more things to maintain

nishu-builder (changes requested):
Do we need `gitlib` to be outside of `metta.common`? I think it's not so specialized that we'd publish it as a separate package, so putting it in `metta.common` feels right to me

I do like moving codebot to its own namespace package!

metta/setup/components/codebot.py
```diff
@@ -0,0 +1,82 @@
...
+
+@register_module
+class CodebotSetup(SetupModule):
+    install_once = True  # Only install once, not every time
+
+    @property
+    def description(self) -> str:
+        return "Codebot tools for AI development (includes codeclip)"
+
+    def _is_applicable(self) -> bool:
```
nishu-builder, line 19: because I think we don't want to stop other profiles from installing this if they ask to explicitly, I think it's better to put this PROFILE_DEFINITIONS

metta/setup/components/codebot.py
```diff
@@ -0,0 +1,82 @@
...
+                self.run_command(
+                    ["uv", "tool", "install", "--force", str(codebot_dir), "--with", f"gitta @ file://{gitta_dir}"]
+                )
+            except subprocess.CalledProcessError as e:
+                warning(f"Failed to install codebot: {e}")
+                warning("You can manually install it with:")
+                warning(f"  uv tool install --force {codebot_dir} --with 'gitta @ file://{gitta_dir}'")
+                return
+
+        # Fix the Python path for the install
```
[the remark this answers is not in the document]
jacklionheart: yikes, the dangers of using claude code at night when youre too tired to review

app_backend/src/metta/app_backend/eval_task_worker.py
```diff
@@ -112,7 +112,7 @@ def _setup_versioned_checkout(self, git_hash: str) -> None:
         os.makedirs(os.path.dirname(self._versioned_path), exist_ok=True)
 
         result = subprocess.run(
-            ["git", "clone", METTA_API_REPO_URL, self._versioned_path],
+            ["git", "clone", git.METTA_API_REPO_URL, self._versioned_path],
```
nishu-builder, line 115: if gittai is supposed to be separately publishable/non-metta specific, does it make sense to hvae METTA_AI_REPO_URL? or is gittai understood to be "git helpers for metta repo usage"?

gitta/src/gitta/__init__.py
```diff
@@ -0,0 +1,757 @@
+"""Git utilities library for Metta projects."""
+
```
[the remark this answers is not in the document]
nishu-builder: some things in here, like `memoize`, would be broadly used; could be in metta-common

nishu-builder (approved):
approving assuming you'll handle the comments :)

would you mind double checking that `uv sync` from the main repo by default doesn't (a) install codebot, or (b) install the codebot tool? since we dont wnat to do so for e.g. docker system environments (like our skypilot jobs)

rjwalters (changes requested):
I think you should split this pr up into separate concerns. I'm not convinced we need a separate git utilities package vs a module in common

rjwalters (approved):
post discussion; let's see how things evolve with multiple subpackages for awhile -- we can move things around later as needed

PR #2338, opened 2025-08-27
Create a new slim dockerfile for training

nathants:
It is ~13GB instead of ~33GB.

It is tested with `test_dockerfile.sh` on a g6 instance.

It duplicates and slightly modifies the existing docker build and push github action, so likely it will just work.

https://app.asana.com/1/1209016784099267/project/1211112167711514/task/1211112167711529

### DIFF
--- a/devops/docker/Dockerfile.slim
+++ b/devops/docker/Dockerfile.slim
@@ -1,0 +1,83 @@
+FROM nvidia/cuda:12.8.1-cudnn-devel-ubuntu24.04 AS builder
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV TORCH_CUDA_ARCH_LIST=7.5;8.0;8.6;8.9;9.0
+ENV READTHEDOCS=True
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    software-properties-common \
+    && add-apt-repository ppa:deadsnakes/ppa \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+    build-essential \
+    curl \
+    git \
+    python3.11 \
+    python3.11-dev \
+    python3.11-venv \
+    ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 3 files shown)


devops/docker/test_dockerfile.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+# Test script for quick training verification with minimal steps
+set -eou pipefail
+
+CONTAINER="${1}"
+echo "Testing container: $CONTAINER"
+
+docker kill $(docker ps -q) || true
```
berekuk, line 8: This could kill more than it should, users might be running other unrelated docker images.
Either remove this line or grep for metta images specifically.

devops/docker/Dockerfile.slim
```diff
@@ -0,0 +1,83 @@
...
+    software-properties-common \
+    && add-apt-repository ppa:deadsnakes/ppa \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+    build-essential \
+    curl \
+    git \
+    python3.11 \
+    python3.11-dev \
+    python3.11-venv \
```
berekuk, line 17: uv can manage its own python (`uv python install`, that would match our `.python-version` precisely), so we should use that instead of system install.

devops/docker/Dockerfile.slim
```diff
@@ -0,0 +1,83 @@
...
+COPY experiments/pyproject.toml experiments/
+
+RUN uv sync --no-install-workspace --frozen
+
+COPY . .
+
+RUN uv sync --locked
+
+RUN rm -rf /root/.cache \
+    && find /workspace/metta -type d -name __pycache__ -exec rm -rf {} + 2>/dev/null || true
```
berekuk, line 44: `.dockerignore` should take care of this

nathants:
Made those changes, and tested.

berekuk:
Oops, I missed that you didn't have permissions to merge.

nathants:
@berekuk cool!

PR #2340, opened 2025-08-27
move run.py code to common, provide "tool" script alias, minimal tests

berekuk:
Reasoning: I'm moving mapgen to mettagrid in another PR, and mapgen includes some tools.

So I need tool-running script to be available in mettagrid without doing `../tools/run.py`.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211164250028273)


nishu-builder (approved):
makes sense!

PR #2345, opened 2025-08-27
fix: avoid possible underflow in move action

rjwalters:
### Problem
The `GridCoord` type is `uint16_t` (unsigned), which can cause underflow when agents attempt to move beyond map boundaries. For example, if an agent at position (0, 0) tries to move up (dr = -1), the direct arithmetic `target_location.r += dr` would underflow, wrapping around to 65535 instead of -1.

### Solution
This PR adds proper bounds checking by:
1. Performing position calculations using signed integers
2. Validating that the resulting coordinates are non-negative and within `uint16_t` range
3. Returning early if the movement would result in invalid coordinates

### Changes
- Calculate new position using signed integer arithmetic
- Add bounds check before casting back to `GridCoord`
- Return `false` if movement would cause underflow or exceed valid range

This prevents edge-case bugs where agents could teleport to the opposite side of the map due to unsigned integer wraparound.

### DIFF
--- a/mettagrid/src/metta/mettagrid/actions/move.hpp
+++ b/mettagrid/src/metta/mettagrid/actions/move.hpp
@@ -39,8 +39,16 @@ protected:
     getOrientationDelta(move_direction, dc, dr);
 
     // Calculate target location
-    target_location.r += dr;
-    target_location.c += dc;
+    int new_r = static_cast<int>(target_location.r) + dr;
+    int new_c = static_cast<int>(target_location.c) + dc;
+
+    if (new_r < 0 || new_r > std::numeric_limits<GridCoord>::max() || new_c < 0 ||
+        new_c > std::numeric_limits<GridCoord>::max()) {
+      return false;
+    }
+
+    target_location.r = static_cast<GridCoord>(new_r);
+    target_location.c = static_cast<GridCoord>(new_c);
 
     // Update orientation to face the movement direction (even if movement fails)
     actor->orientation = move_direction;


mettagrid/src/metta/mettagrid/actions/move.hpp
```diff
@@ -39,8 +39,16 @@ class Move : public ActionHandler {
     getOrientationDelta(move_direction, dc, dr);
 
     // Calculate target location
-    target_location.r += dr;
-    target_location.c += dc;
+    int new_r = static_cast<int>(target_location.r) + dr;
```
sasmith, line 42: Right now this is effectively handled in a couple of semi-overlapping ways:
1) We consistently have boundaries on the map, so agents should never be at r=0 or c=0. Thus an agent should never be in a position where they can underflow.
2) Our maps are universally (for now!) smaller than `std::numeric_limits<GridCoord>::max()` in height / width. This means that if agents did manage to underflow, they would still end up in an invalid square, and thus would fail `_is_valid_square` on line 57.

How about we add an assertion that the grid dimensions are always less than `std::numeric_limits<GridCoord>::max() - small_buffer`, and then we can just comment here about why we're not worried about over / underflow?

sasmith (changes requested):
As per inline comment.

Or better, can we compile with in a way to abort if we overflow? Something like
https://stackoverflow.com/questions/5005379/c-avoiding-overflows-when-working-with-big-numbers/5005792#5005792

or using `-fsanitize=unsigned-integer-overflow`

relh (approved):

PR #2358, opened 2025-08-27
Remove stale-seeming test outputs

sasmith:


rjwalters (approved):
these should have been gitignored

PR #2369, opened 2025-08-28
Stub out the panels.

treeform:


mettascope2/src/mettascope.nim
```diff
@@ -86,13 +59,32 @@ proc display() =
   drawTimeline(globalTimelinePanel)
   globalTimelinePanel.endDraw()
 
+  worldMapPanel.beginDraw()
```
lcodes, line 62: Lots of begin/draw/end here, starting to look like there's a pattern to make general :)

panel could get its draw function, and that wraps the whole sequence inside it

but counterpoint; begin/end draw is rendering concern, panel is just the data in a layout,
meaning object-level drawing loses potential to batch across panels - or complexifies it

ideally, want panels as just data, a PanelRenderer thing taking *all* panels each frame, maintaining its own cache of which are dirty/clean (can be added later, a system makes this sort of thing easy to do over individual objects since it always sees the world of objects at once), handling redraw regions across panels uniformly, handling layout across panels just like panels have layout inside them, the system data is 1:1 (or close) the binary blob to save user layouts, and it decouples concerns by lifting all drawing code away from assets (unless they are rendering assets, but those end up lower level like textures and materials and meshes and shaders)

hm kinda long comment, tl;dr: many panel instances for data, one panel renderer for code

lcodes: which further generalizes to a pass renderer, of which ui ends up movable in the pipeline :)

meaning if you want world-space UI (say health bars that move with entities, or hp bars/text) its a trivial change from screen-space UI, or even moving the whole UI to a render texture and that to a 3D mesh (which yeah is beyond scope, but to show how flexible it gets :D)

so the frame is cycling through an array of render pass definitions, at least for the gfx part

mettascope2/src/mettascope/agenttable.nim
```diff
@@ -0,0 +1,12 @@
+import
+  boxy, chroma,
+  common, panels
+
+proc drawAgentTable*(panel: Panel) =
+
+  let box = IRect(x: 0, y: 0, w: panel.rect.w, h: panel.rect.h)
```
lcodes, line 7: at the ui level it would be nice to switch from boxy to controls and common styles

background = color
layout = hbox(tools, viewport, status)

even if it ends up immediate mode under the hood, meaning those translate to boxy calls
(or batch up an IR that flushes to boxy at the end of frame)

have you seen Dear ImGui? kinda same idea, but much simpler (or can just use it directly, got it working in wasm in the past and it pushes 50k vertices per frame on old hardware without touching the gpu/cpu perf much)

lcodes: retained mode is the one that's tricky, adds state, and is closer to the web DOM; too big

mettascope2/src/mettascope/envconfig.nim
```diff
@@ -0,0 +1,12 @@
+import
+  boxy, chroma,
+  common, panels
+
+proc drawEnvConfig*(panel: Panel) =
+
+  let box = IRect(x: 0, y: 0, w: panel.rect.w, h: panel.rect.h)
```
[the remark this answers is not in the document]
lcodes: simplicity to me is "when I work in X, do I need to also think about Y, to which degree?"

and also "when there's many X's, can I batch shared behavior across a SystemX?"

mettascope2/src/mettascope/worldmap.nim
```diff
@@ -296,3 +296,30 @@ World
     16,
     color(1, 1, 1, 1)
   )
+
+proc drawWorldMap*(panel: Panel) =
```
lcodes, line 300: nitpick, at this point we could rename this to drawWorld; "map" implies there's a bigger version

:)

mettascope2/src/mettascope.nim
monofuel: "render texture to a 3D mesh"


obviously the next iteration needs to include VR. imagine floating panels that can be positioned and scaled. or maybe we could enter the simulation ourselves in VR from the perspective of an agent. why use a regular gamepad when we could have motion controls to pick up ore.

mettascope2/src/mettascope.nim
```diff
@@ -111,8 +103,12 @@ else:
   proc cmd(replay: string = "") =
     if replay != "":
       if replay.startsWith("http"):
-        let data = puppy.fetch(replay)
-        common.replay = loadReplay(data, replay)
+        let req = startHttpRequest(replay)
+        req.onError = proc(msg: string) =
+          echo "onError: " & msg
```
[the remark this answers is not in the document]
monofuel: I've used something like this on some of my projects: https://github.com/monofuel/Regen/blob/master/src/logs.nim (logs time, uses colors, info/debug/warn/error) however I feel like that pattern might be overkill for this app?

I feel like mettascope should not log very much, logging things should mean either debugging or there's been an error. otherwise it's a lot of noise.

mettascope2/src/mettascope.nim
lcodes: kinda did that a long time ago, before webgl2 existed; custom webgl renderer and a forked pixijs to get its sprites onto textures projected back to surfaces in VR where they'd animate when looking at them hehe

lcodes: ugh that was also before webxr and webvr was kinda jank :p

treeform: I don't want this behind more abstractions, i don't want to hide this before we know how it will shape out in the end.

mettascope2/src/mettascope/agenttable.nim
treeform: I am thinking about common controls. Its a difficult problem. They mostly should live in ui.nim and in this new thing.

lcodes (approved):

PR #2384, opened 2025-08-28
feat: change skypilot to use python

rjwalters:
wip


devops/skypilot/launch.py
```diff
@@ -274,15 +275,12 @@ def main():
...
-            copy_task = copy_task.update_envs({"METTA_RUN_ID": run_id})
-            copy_task.name = run_id
-            copy_task.validate_name()
-            launch_task(copy_task)
+    request_ids = [launch_task(copy.deepcopy(task)) for _ in range(args.copies)]
+
+    # auto launch log if we have only one task
+    if len(request_ids) == 1:
+        time.sleep(1)
+        subprocess.run(["sky", "api", "logs", request_ids[0]])
```
berekuk, line 283: Is this reliable (1 second timeout)? Random failures when api server is slow would be annoying.

I remember David was opposed to launch waiting for the request to finish by default, though I'm not sure if it's still important (and I kinda disagree, I think waiting for job id would be good for usability, we could print the link to web dashboard).

gitta/src/gitta/__init__.py
```diff
@@ -691,6 +694,89 @@ def post_commit_status(
...
+        wandb_run_id: Optional WandB run ID for target URL
+        max_retries: Maximum number of retry attempts
+
+    Returns:
+        True if status was posted successfully, False otherwise
+    """
+    logger = logging.getLogger(__name__)
+
+    if not all([state, description, commit_sha]):
+        logger.warning("GitHub status requires state, description, and commit_sha")
```
berekuk, line 728: Seems like this warning will always happen for non-CI jobs? Remove if I'm right.

berekuk (changes requested):
I'm marking "request changes" because I left some minor comments, the most significant (but still minor) is about heartbeat code duplication.

I also wonder if it makes sense to fold `run.sh` into `skypilot_run.py` code, they're too tightly coupled to make sense separately.

devops/skypilot/launch.py
```diff
@@ -277,15 +277,12 @@ def main():
...
-        for _ in range(1, args.copies + 1):
-            copy_task = copy.deepcopy(task)
-            copy_task = copy_task.update_envs({"METTA_RUN_ID": run_id})
-            copy_task.name = run_id
-            copy_task.validate_name()
-            launch_task(copy_task)
+    request_ids = [launch_task(copy.deepcopy(task)) for _ in range(args.copies)]
+
+    # auto launch log if we have only one task
+    if len(request_ids) == 1:
```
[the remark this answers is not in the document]
rjwalters: collecting the job id is the most frequent path for me. We don't have much real world data on how people are using the launch command but happy to optimize for your work path if that is the design goal

devops/skypilot/utils/nccl_tests.py
```diff
@@ -8,6 +8,7 @@
 import subprocess
 import sys
 import time
+from contextlib import redirect_stderr, redirect_stdout
```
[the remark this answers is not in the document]
rjwalters: these are used to redirect output to logs -- but I think you are asking if we need the nccl tests at all. this is a fair question -- it was useful for debugging shared memory and I don't think we are finished with optimizing nccl communication yet. My preference would be to leave this in until the next cooling cycle and then delete it at that point but we can also remove now and add it again later if/when we want to optimize nccl again

devops/skypilot/utils/configure_environment.py
```diff
@@ -0,0 +1,243 @@
...
+    else:
+        logger.info_master("restart count file not found, setting restart_count to 0")
+        restart_count = 0
+
+    # Only update restart count on master node
+    if is_master:
+        restart_count_file.write_text(str(restart_count))
+        logger.info_master(f"Updated restart count file to {restart_count}")
+    else:
+        logger.info("Skipping RESTART_COUNT_FILE updates on non-master node")
```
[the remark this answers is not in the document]
rjwalters: this is useful for detecting crashing in loops by noting that many restarts have occurred without accumulating training time.

devops/skypilot/utils/configure_environment.py
```diff
@@ -0,0 +1,243 @@
...
+        "accumulated_runtime_sec": accumulated_runtime_sec,
+        "accumulated_runtime_file": str(accumulated_runtime_file),
+        "heartbeat_file": str(heartbeat_file),
+    }
+
+
+def write_environment_variables(metta_env_file, metadata=None):
+    """Write environment variables to METTA_ENV_FILE."""
+    env_vars = """export PYTHONUNBUFFERED=1
+export PYTHONPATH="${PYTHONPATH:+$PYTHONPATH:}$(pwd)"
```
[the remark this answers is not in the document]
rjwalters: this function write dynamic variables to a file. maybe I don't understand the concern...

```
# Job metadata exports
export RESTART_COUNT="{metadata["restart_count"]}"
export ACCUMULATED_RUNTIME_SEC="{metadata["accumulated_runtime_sec"]}"

# File path exports for monitors
export ACCUMULATED_RUNTIME_FILE="{metadata["accumulated_runtime_file"]}"
export HEARTBEAT_FILE="{metadata["heartbeat_file"]}"
```

devops/skypilot/utils/configure_environment.py
```diff
@@ -0,0 +1,239 @@
...
+    if args.sandbox:
+        logger.info("Running in SANDBOX mode - skipping job metadata setup")
+    else:
+        logger.info("Running in MANAGED JOB mode")
+
+    # Print initial environment info only on master
+    logger.info_master(f"VIRTUAL_ENV: {os.environ.get('VIRTUAL_ENV', 'Not set')}")
+    logger.info_master(f"Which python: {run_command('which python')}")
+    python_exec = run_command([sys.executable, "-c", "import sys; print(sys.executable)"])
+    logger.info_master(f"Python executable: {python_exec}")
```
berekuk, line 195: These seem more like remnants from debugging than like something generally useful.

If I had an issue related to these that I needed to debug, I'd probably do something like `kubectl run my-test --image 751442549699.dkr.ecr.us-east-1.amazonaws.com/metta:latest-slim --command -- sleep infinity`, then attach through `kubectl exec -it my-test -- bash`, and investigate from there.

(or just launch a simplified sandbox with skypilot, with the minimal setup and run)

devops/skypilot/utils/configure_environment.py
```diff
@@ -0,0 +1,239 @@
...
+export NCCL_P2P_DISABLE="${NCCL_P2P_DISABLE:-0}"
+export NCCL_SHM_DISABLE="${NCCL_SHM_DISABLE:-0}"
+export NCCL_IB_DISABLE="${NCCL_IB_DISABLE:-1}"
+"""
+
+    # Append to file
+    with open(metta_env_file, "a") as f:
+        f.write(env_vars)
+
+    logger.info(f"Environment variables written to: {metta_env_file}")
```
berekuk, line 143: I feel pretty skeptical about the entire pattern of writing env vars to the file, only to read them again immediately in `skypilot_run.py`. It seems like it could be simplified; `skypilot_run.py` could do this through a module.

At the same time, I realize that:
1) this pattern was there before so it might be out of scope of this PR
2) we need some of these vars in sandbox mode (though I'm not sure how many we really need, and would question some of these, e.g. `PYTHONOPTIMIZE`, and node rank-related stuff, and possibly datadog; I don't know anything about NCCL)

devops/skypilot/utils/configure_environment.py
```diff
@@ -0,0 +1,239 @@
...
+        if observatory_path.exists():
+            logger.info("~/.metta/observatory_tokens.yaml already exists")
+        else:
+            observatory_path.parent.mkdir(exist_ok=True)
+            observatory_content = (
+                f"{PROD_STATS_SERVER_URI}: {observatory_token}\n"
+                + f"{PROD_OBSERVATORY_FRONTEND_URL}/api: {observatory_token}\n"
+            )
+            observatory_path.write_text(observatory_content)
+            logger.info("~/.metta/observatory_tokens.yaml created")
```
berekuk, line 175: This looks like it should be another `metta` cli command, or at least a function that's closer to `metta.setup` that knows where to write tokens and in which format.

Otherwise it might get out of sync.

devops/skypilot/utils/job_config.py
```diff
@@ -68,7 +68,7 @@ def to_filtered_dict(self, filtered_field_names: list[str] | None = None) -> dic
 def log_job_config(jc: JobConfig):
     """Log job configuration with sensitive values redacted."""
     logger.info("Run Configuration:")
-    for field_name, value in jc.to_filtered_dict().items():  # Need .items() here
+    for field_name, value in jc.to_filtered_dict().items():
```
berekuk, line 71: I see that `to_filtered_dict()`tries to filter out github pat and discord webhook.

It's possible that skypilot logs are more important to protect than skypilot UI (maybe because we could publish the logs by accident somewhere?), but we still leak these secrets in skypilot entrypoint in dashboard: https://skypilot-api.softmax-research.net/dashboard/jobs/3255

devops/skypilot/launch/skypilot_run.py
```diff
@@ -0,0 +1,227 @@
...
+            }
+
+            if job_config.wandb_project and job_config.wandb_entity:
+                wandb_cfg = WandbConfig(
+                    enabled=True,
+                    project=job_config.wandb_project,
+                    entity=job_config.wandb_entity,
+                    run_id=job_config.metta_run_id,
+                )
+                with WandbContext(wandb_cfg, extra_cfg=job_config.to_filtered_dict()):
```
berekuk, line 178: Wandb merges the config keys if you initialize it twice, and after you pass `job_config` here, it will later be merged with the train-specific config.

Because of this, I think it'd be safer and more orderly to pass `extra_cfg={"skypilot_job_config": job_config.to_filtered_dict()}` here.

devops/skypilot/launch/skypilot_run.py
```diff
@@ -0,0 +1,227 @@
...
+            average_runtime_minutes = (job_config.accumulated_runtime_sec / job_config.restart_count) / 60
+            if average_runtime_minutes < 3:
+                termination_reason = "rapid_restarts"
+
+        if not termination_reason:
+            subprocess = run_job_in_background()
+            termination_reason = monitor_until_termination(job_config, subprocess)
+
+    except BaseException as e:
+        exit_code = getattr(e, "code", 1) if isinstance(e, SystemExit) else 1
```
berekuk, line 198: No need for `getattr` here? If `isinstance` returns true then you can do `e.code` directly.

devops/skypilot/launch/skypilot_run.py
```diff
@@ -0,0 +1,227 @@
...
+        if isinstance(e, SystemExit) and e.code is None:
+            exit_code = 1
+
+        termination_reason = f"job_failed_{exit_code}"
```
berekuk, line 202: Hmm, why is `termination_reason` a string?

I know it was convenient when you did github workflows to test it, but now we don't have that code, and all this serialization and parsing makes the implementation more fragile and less type safe. Could be a tiny dataclass with enum field and an optional exit code...

devops/skypilot/utils/runtime_monitors.py
```diff
@@ -0,0 +1,173 @@
...
+import time
+from pathlib import Path
+from typing import Optional
+
+from metta.common.util.log_config import getRankAwareLogger
+
+logger = getRankAwareLogger(__name__)
+
+
+class HeartbeatMonitor:
```
berekuk, line 16: Can you add an abstract base class with the typed abstract method for `check_condition` for all these classes? Then `skypilot_run` code could be a bit more typesafe.

devops/skypilot/utils/runtime_monitors.py
```diff
@@ -0,0 +1,173 @@
...
+        self.heartbeat_file = Path(heartbeat_file_path)
+
+        try:
+            self.heartbeat_file.parent.mkdir(parents=True, exist_ok=True)
+            self.heartbeat_file.touch()  # Updates mtime on restart
+            logger.info(f"Updated heartbeat file at {self.heartbeat_file}")
+        except Exception as e:
+            logger.error(f"Failed to update heartbeat file: {e}")
+
+    def check_condition(self) -> tuple[bool, Optional[str]]:
```
berekuk, line 43: Why not just `Optional[str]` (here and in other monitors)? You always return either `True, "..."` or `False, None`.

devops/skypilot/launch/skypilot_run.py
```diff
@@ -0,0 +1,227 @@
...
+            logger.info(f"Subprocess exited with code {exit_code}")
+            if exit_code == 0:
+                return "job_completed"
+            else:
+                return f"job_failed_{exit_code}"
+
+        for monitor in monitors:
+            should_terminate, reason = monitor.check_condition()
+            if should_terminate:
+                logger.info(f"{monitor.name} triggered: {reason}")
```
berekuk, line 139: `monitor.__class__.__name__`, and then you won't have to define `name` in each class implementation.

berekuk (approved):
Overall thoughts:
- monitors pattern and `skypilot_run.py` are good, well-organized, etc.
- I'm not sure if `JobConfig` and the amount of env vars you have to pass around is optimal; probably could be simplified, but I understand this evolved from shell scripts, so I'm fine with it
- I haven't paid much attention to NCCL test since it's a test and I don't know enough about NCCL to judge
- I left a bunch of requests for improvements but I recommend merging this in soon and then continuing from there in other branches (assuming this passes your test runs)

Kudos on getting this to completion!

devops/skypilot/launch/skypilot_run.yaml
```diff
@@ -43,15 +43,26 @@ file_mounts:
...
+  set -e
+  cd /workspace/metta
 
-envs:
-  TEST_NCCL: "false"
+  # Note that the docker image may start with its own venv - switch to metta venv
+  if [ -n "$VIRTUAL_ENV" ]; then
+      deactivate 2>/dev/null || true
+  fi
+  . .venv/bin/activate
```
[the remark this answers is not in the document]
rjwalters: Do we want to leave this in as a defensive measure for future images or remove it?

I could also replace with an assertion if you prefer:
```
[ -z "${VIRTUAL_ENV:-}" ] || { echo "[ERROR] Already inside a virtualenv" >&2; exit 1; }
. .venv/bin/activate

```

rjwalters:
Thanks for the migration! A few things I noticed while reading through the branch:

1. devops/skypilot/utils/termination_reason.py:31 –  currently interpolates the enum member itself () so the formatted strings come out as . The notification pipeline expects strings that start with , so we should use the enum value instead (e.g. ).
2. . devops/skypilot/connect.py:37 – the new lookup points at , but after the refactor the recipe lives at . As written the command raises .
3. . devops/skypilot/launch.py:318 – we now append a random suffix to every Sky cluster name. That helps when launching multiple copies, but it also pushes otherwise valid 63-char names over Sky’s limit even for a single copy. Maybe only append when , or trim the base name before suffixing.
 
Let me know what you think!

rjwalters:
Follow-up from the earlier feedback: the branch now applies the three fixes we discussed (enum value, connect path, cluster name suffix). Let me know if you spot anything else we should tighten up.

devops/skypilot/recipes/skypilot_run.yaml
```diff
@@ -43,15 +43,26 @@ file_mounts:
...
 
-# `set -u`: Exit on undefined variables to catch missing env vars early
-setup: set -u && bash <(curl -fsSL "https://raw.githubusercontent.com/${GITHUB_REPOSITORY}/${METTA_GIT_REF}/devops/skypilot/config/lifecycle/setup.sh")
+setup: set -u && bash <(curl -fsSL "https://raw.githubusercontent.com/${GITHUB_REPOSITORY}/${METTA_GIT_REF}/devops/skypilot/launch/setup.sh")
 
-# `set -e`: Ensure SkyPilot propagates exit codes from skypilot_run.sh (otherwise restart test fails)
-run: set -e && /workspace/metta/devops/skypilot/config/skypilot_run.sh
+run: |
+  set -e
+  cd /workspace/metta
```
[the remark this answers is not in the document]
daveey: please fix

devops/skypilot/utils/job_helpers.py
```diff
@@ -370,3 +371,63 @@ def tail_job_log(job_id: str, lines: int = 100) -> str | None:
...
+            return str(job_id), status_msg
+
+        except KeyboardInterrupt:
+            print("\n" + yellow("Stopped tailing logs"))
+            return str(job_id), "Log tailing interrupted by user"
+
+    except sky.exceptions.ClusterNotUpError:
+        error_msg = "Error: Jobs controller is not up"
+        print(red(error_msg))
+        return None, error_msg
```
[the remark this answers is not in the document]
daveey: please fix

rjwalters:
## Cluster Test Results - Full Validation Complete ✅

Successfully completed comprehensive cluster validation testing with **18/18 tests passing** across 1, 2, and 4 node configurations.

### Test Summary
- **1-node tests**: 6/6 ✅
- **2-node tests**: 6/6 ✅  
- **4-node tests**: 6/6 ✅
- **Total**: 18/18 SUCCEEDED (100% success rate)

### Critical Fixes Validated

1. **Init Script Execution** (commit 9325372500)
   - Fixed: Job initialization now properly pipes `init_run.py` through python3
   - Previously: `bash: import: command not found` errors
   - Status: ✅ All jobs initialize correctly

2. **WandbContext Parameter** (commit 70709eb36d)
   - Fixed: Corrected parameter name from `extra_cfg` to `run_config`
   - Previously: `TypeError: WandbContext.__init__() got an unexpected keyword argument 'extra_cfg'`
   - Status: ✅ W&B integration working, notifications sent successfully

### Test Coverage Validated

**Exit Conditions** (all working correctly):
- ✅ Normal completion: Jobs complete successfully with `job_completed`
- ✅ Heartbeat timeout: Jobs detect missing heartbeats and exit with `heartbeat_timeout`
- ✅ Runtime timeout: Jobs respect max runtime limits
- ✅ Invalid parameters: Jobs handle parameter validation errors

**Infrastructure Tests**:
- ✅ NCCL communication: Multi-node GPU communication validated across 1, 2, and 4 nodes
- ✅ Job restart functionality: Restart recovery tested successfully
- ✅ Torchrun wrapper: Multi-GPU coordination working
- ✅ Monitoring loops: Heartbeat and runtime monitors operational

**Multi-Node Coordination**:
- ✅ 4-node NCCL test: All 4 nodes (head + worker1,2,3) successfully coordinated
- ✅ 4-node restart test: All nodes reported proper termination
- ✅ Cross-node communication: Working correctly

### Job Execution Details

Example job durations:
- 1-node tests: ~6-7 minutes (setup + execution)
- 2-node tests: ~6-7 minutes  
- 4-node tests: ~6-7 minutes

All jobs:
- Provisioned clusters successfully
- Installed dependencies via uv
- Executed test workloads
- Reported proper termination reasons
- Sent W&B notifications
- Cleaned up resources

### Infrastructure Validation

✅ SkyPilot job orchestration working end-to-end  
✅ AWS cluster provisioning functional  
✅ Docker container setup operational  
✅ S3 bucket mounting working  
✅ Secret management functional  
✅ Cost tracking reporting correctly  

**Conclusion**: The Python refactor of SkyPilot infrastructure is production-ready. All core functionality validated across multiple cluster configurations.

PR #2399, opened 2025-08-29
feat: rank-aware logging

rjwalters:
This PR enhances the logging utilities to support distributed computing contexts like PyTorch DDP, SkyPilot, and multi-node deployments.

**Key changes:**
- Add `show_rank` parameter to `init_logging()` to display `[rank]` prefix in log messages
- Add `get_node_rank()` to detect rank from common distributed environment variables (SKYPILOT_NODE_RANK, RANK, OMPI_COMM_WORLD_RANK)
- Add `log_master()` utility for logging only on rank 0 (master node)
- Update file logging to create rank-specific log files (e.g., `script_1.log`, `script_2.log`)
- Add comprehensive documentation with examples for PyTorch DDP, SkyPilot, and data pipelines

**Benefits:**
- Easier debugging of distributed training by seeing which worker produced each log
- Cleaner logs with master-only messages for coordination events
- Better log organization with separate files per rank

Example:
```python
init_logging(show_rank=True)
logger.info("Worker ready")
# Output: [12:34:56.789] [0] INFO     Worker ready
# Output: [12:34:56.790] [1] INFO     Worker ready

log_master("Starting training")  # Only rank 0 logs this
```

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211186822765357)


nishu-builder (changes requested):
Love the changes. This may be scope creep, and if you like these but feel this PR is not the right time for them, I would personally enjoy following up with these:
- `log(message, level=INFO, show_rank=True, master_only=False)` that also idempotently does init_logging at the start, and serves as the only thing through which we log
- get_node_rank() to read off constants.py
- i think this is sufficiently self-documenting and doesnt need the README, but I'm also particularly README-allergic by disposition so consider this a biased suggestion :P

PR #2422, opened 2025-09-02
Remove the remaining non-mettagrid imports

berekuk:
I had to keep the conditional PlayTool/mettascope import in mapgen show_map, for now (unsure if it'll be useful post-dehydration, will revisit later).

This PR also includes the test that'll catch all future violations.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211213067779041)

### DIFF
--- a/metta/sim/map_preview.py
+++ b/metta/sim/map_preview.py
@@ -9,6 +9,7 @@ import wandb
 from wandb.sdk import wandb_run
 
 from metta.common.util.constants import METTASCOPE_REPLAY_URL
+from metta.common.util.fs import get_repo_root
 from metta.mettagrid.mettagrid_config import MettaGridConfig
 from metta.mettagrid.mettagrid_env import MettaGridEnv
 from metta.mettagrid.util.file import write_file
@@ -40,7 +41,8 @@ def write_map_preview_file(preview_path: str, env: MettaGridEnv, gzipped: bool):
 
 
 def write_local_map_preview(env: MettaGridEnv):
-    maps_dir = "./outputs/maps"
+    repo_root = get_repo_root()
+    maps_dir = repo_root / "outputs" / "maps"
     os.makedirs(maps_dir, exist_ok=True)
 
     with tempfile.NamedTemporaryFile(delete=False, dir=maps_dir, suffix=".json") as temp_file:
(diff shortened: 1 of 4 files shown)


metta/sim/map_preview.py
```diff
@@ -40,7 +41,8 @@ def write_map_preview_file(preview_path: str, env: MettaGridEnv, gzipped: bool):
 
 
 def write_local_map_preview(env: MettaGridEnv):
-    maps_dir = "./outputs/maps"
+    repo_root = get_repo_root()
+    maps_dir = repo_root / "outputs" / "maps"
```
jacklionheart, line 45: Why is this writing to the lcoal repo anyway?

mettagrid/tests/test_forbidden_imports.py
```diff
@@ -0,0 +1,72 @@
+from __future__ import annotations
```
jacklionheart, line 1: Thanks!

mettagrid/src/metta/mettagrid/mapgen/utils/show.py
```diff
@@ -18,6 +14,14 @@ def show_map(storable_map: StorableMap, mode: ShowMode | None):
...
 
     if mode == "mettascope":
+        try:
+            import mettascope.server
+            from metta.sim.map_preview import write_local_map_preview
+            from metta.sim.simulation_config import SimulationConfig
+            from metta.tools.play import PlayTool
+        except ImportError:
+            raise ImportError("Mettascope mode is not available outside of metta monorepo.") from None
+
```
jacklionheart, line 24: I feel like this is not quite the right answer, though I understand feeling like you need more outside input to know the right answer.

What I think is:

* If/when we had a public mettagrid repo, we would probably want to bundle mettascope with it in some fashion.
* It could be ok to rewrite this so that show_map doesnt allow mettascope and you need to invoke a different function to show it
* It could also be ok to move PlayTool into mettagrid/common if necessary
* We can/should be able to rewrite PlayTool to take a MettaGridConfig rather than an env cfg, though I think its also fair to say that simulations should be considered part of Mettagrid -- inference should be baked into the library

metta/sim/map_preview.py
berekuk: Because the only way to preview maps with mettascope is to produce a local replay. I mean, I could produce an s3 replay, but that would be even more clunky. That outputs dir is gitignored, btw.

jacklionheart (approved):

PR #2431, opened 2025-09-02
fixfriction: dependabot -> renovate migration

relh:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211213878573606)


rjwalters (changes requested):
let's move the script out of the workflow to its own python file

PR #2437, opened 2025-09-02
fix resource name rechange

jacklionheart:


sasmith:
I see no changed files.

sasmith:
Maybe this is just a merge, in a way where it doesn't show up?

PR #2444, opened 2025-09-02
feat: add intelligent uv store caching to speed up CI

zfogg:
## Summary

Adds intelligent caching for uv store and virtual environments to dramatically speed up CI by avoiding repeated downloads and installations when dependencies haven't changed.

### Smart Caching Strategy
- **Cache .venv and ~/.cache/uv** based on comprehensive dependency tracking
- **Cache key includes**: `uv.lock`, `pyproject.toml`, `**/pyproject.toml` (all workspace members)
- **Per-install-mode caching**: Separate caches for `testing`, `full`, `linting`, and `minimal` modes
- **Fallback cache keys**: Partial hits when lockfile unchanged but git SHA differs

### Performance Optimization
- ✅ **Cache Hit**: Skip `uv sync` entirely (~90% faster)
- ❌ **Cache Miss**: Run full `uv sync` as before
- **Clear feedback**: Logs show cache hit/miss status

### When Cache Invalidates (New Download/Install):
- Any `pyproject.toml` in workspace changes (root or any member)
- `uv.lock` file changes
- Different install mode (`testing` vs `full` etc.)
- Different OS (`ubuntu-latest` vs `macos-latest`)

### When Cache Hits (Skip Download/Install):
- Same lockfile + same workspace configs + same mode + same OS
- Just activate existing `.venv` and continue

### Cache Hierarchy:
```
Primary:   uv-store-{mode}-{os}-{hash(uv.lock,**/pyproject.toml)}-{git-sha}
Fallback:  uv-store-{mode}-{os}-{hash(uv.lock,**/pyproject.toml)}-
Generic:   uv-store-{mode}-{os}-
```

### Workspace Member Tracking:
The cache key includes ALL workspace member `pyproject.toml` files:
- `mettagrid/pyproject.toml`
- `app_backend/pyproject.toml`
- `common/pyproject.toml`
- `agent/pyproject.toml`
- `experiments/pyproject.toml`
- `gitta/pyproject.toml`
- `codebot/pyproject.toml`

This ensures cache invalidation when ANY workspace member's dependencies change.

## Expected Impact

**Current CI times** (rough estimates):
- Setup dependencies: ~2-3 minutes per job
- Multiple jobs per PR: 4-8 jobs
- Total dependency setup time: 8-24 minutes per PR

**With caching** (after first run):
- Cache hit: ~10-20 seconds per job
- Multiple jobs per PR: 4-8 jobs  
- Total dependency setup time: 40-160 seconds per PR

**Expected speedup**: ~90% faster CI runs when dependencies unchanged (most PRs)

## Testing

This PR itself will test the caching mechanism:
1. **First workflow run**: Cache miss → normal timing
2. **Second workflow run**: Cache hit → should be much faster
3. **Check logs**: Should show "✅ Cache hit - skipping uv sync (much faster!)"

Let's see how much faster the second run is! 🚀

### DIFF
--- a/.github/actions/setup-uv/action.yml
+++ b/.github/actions/setup-uv/action.yml
@@ -20,35 +20,58 @@ runs:
       env:
         UV_CACHE_DIR: ${{ github.workspace }}/.uv-cache
 
+    - name: Cache uv store
+      uses: actions/cache@v4
+      id: cache-uv-store
+      with:
+        path: |
+          .venv
+          ~/.cache/uv
+        key: uv-store-${{ inputs.install-mode }}-${{ runner.os }}-${{ hashFiles('uv.lock', 'pyproject.toml', '**/pyproject.toml') }}-${{ github.sha }}
+        restore-keys: |
+          uv-store-${{ inputs.install-mode }}-${{ runner.os }}-${{ hashFiles('uv.lock', 'pyproject.toml', '**/pyproject.toml') }}-
+          uv-store-${{ inputs.install-mode }}-${{ runner.os }}-
+
     - name: Create virtual environment with uv
       shell: bash
       env:
         UV_CACHE_DIR: ${{ github.workspace }}/.uv-cache
       run: |
-        uv venv
-
-        # Export environment variables
+        # Export environment variables first
         echo "VIRTUAL_ENV=$(pwd)/.venv" >> $GITHUB_ENV
         echo "$(pwd)/.venv/bin" >> $GITHUB_PATH
 
-        # Handle different installation modes
-        case "${{ inputs.install-mode }}" in
+        # Check if we hit the cache
+        if [ "${{ steps.cache-uv-store.outputs.cache-hit }}" = "true" ]; then
+          echo "✅ Cache hit - skipping uv sync (much faster!)"
+          # Just ensure venv exists and is activated
+          if [ ! -d ".venv" ]; then
+            echo "Creating venv (cache had no .venv)"
+            uv venv
+          fi
+        else
+          echo "❌ Cache miss - running full uv sync"
+          uv venv
 
-          "minimal")
-            # don't install any packages
-            ;;
+          # Handle different installation modes
+          case "${{ inputs.install-mode }}" in
 
-          "testing")
-            uv sync
-            uv pip install pytest pytest-cov pytest-xdist pytest-benchmark gcovr
-            ;;
+            "minimal")
+              # don't install any packages
+              ;;
 
-          "linting")
-            uv pip install ruff cpplint
-            ;;
+            "testing")
+              uv sync
+              uv pip install pytest pytest-cov pytest-xdist pytest-benchmark gcovr
+              ;;
 
-          "full"|*)
-            uv sync
-            ;;
+            "linting")
+              uv pip install ruff cpplint
+              ;;
 
-        esac
+            "full"|*)
+              uv sync
+              ;;
+
+          esac
+        fi


rjwalters (changes requested):
I remember spending a lot of time on this - the cache suffix was important since we call the action from jobs that use different sets of packages. if you want to change this, please make sure you test everything super well and that it makes a meaningful difference to CI times

rjwalters:
> I remember spending a lot of time on this - the cache suffix was important since we call the action from jobs that use different sets of packages. if you want to change this, please make sure you test everything super well and that it makes a meaningful difference to CI times

also -- it really looks like there are only two substantive changes being suggested here - changing "prune cache" and dropping the hash from the suffix. I think you can remove basically all the other proposed changes

berekuk (approved):

.github/actions/asana/pr_description_asana_to_gh/action.yml
```diff
@@ -18,11 +18,11 @@ inputs:
...
   steps:
-    - name: Install uv
-      uses: astral-sh/setup-uv@v6
+    - name: Setup uv
+      uses: ./.github/actions/setup-uv
       with:
-        version: "0.7.3"
-        enable-cache: true
+        install-mode: minimal
+        dependencies-files: dependencies.txt
```
rjwalters, line 25: none of these have giant lists of dependencies -- I think we can avoid creating new files by using your other input:

```
extra-dependencies: |
          requests>=2.31.0
 ```

File: .github/actions/setup-uv/action.yml [hunk 2]
```diff
@@ -5,33 +5,67 @@ inputs:
     description: "Installation mode: 'testing', 'runtime', 'linting', or 'minimal'"
     required: false
     default: "testing"
+  extra-dependencies:
+    description: "Extra pip dependencies to install (newline-separated list)"
+    required: false
+    default: ""
+  dependencies-files:
```
rjwalters, line 12: i really don't think we'll need this but if you want to leave it in it is okay with me

.github/actions/setup-uv/action.yml
```diff
@@ -5,33 +5,67 @@ inputs:
...
       env:
-        UV_CACHE_DIR: ${{ github.workspace }}/.uv-cache
+        UV_NO_SYNC: "1"
       run: |
-        uv venv
-
         # Export environment variables
         echo "VIRTUAL_ENV=$(pwd)/.venv" >> $GITHUB_ENV
         echo "$(pwd)/.venv/bin" >> $GITHUB_PATH
+        echo "UV_NO_SYNC=1" >> $GITHUB_ENV
```
rjwalters, line 62: nice solution!

.github/actions/setup-uv/action.yml
```diff
@@ -5,33 +5,67 @@ inputs:
...
-        cache-local-path: "${{ github.workspace }}/.uv-cache"
         # disabling cache saves 30 sec on testing but costs 40 sec in other modes
-        prune-cache: ${{ inputs.install-mode == 'linting' || inputs.install-mode == 'minimal'}}
-        # we hash in the action script in case the install-mode content has changed
-        cache-suffix: "${{ inputs.install-mode }}-${{ hashFiles('.github/actions/setup-uv/action.yml') }}"
-      env:
-        UV_CACHE_DIR: ${{ github.workspace }}/.uv-cache
+        prune-cache: false
+        cache-suffix: "${{ inputs.install-mode }}-${{ inputs.extra-dependencies }}-${{ inputs.dependencies-files }}"
+        cache-dependency-glob: |
```
rjwalters, line 31: this is kind of ugly... I think we can invalidate caches if/when people change uv.lock. I think this would be better as:

```
cache-dependency-glob: |
    uv.lock
    **/pyproject.toml
    .github/actions/setup-uv/dependencies-*
```

I don't thin you need the action.yml or the complexity here

.github/workflows/checks.yml
```diff
@@ -203,19 +203,25 @@ jobs:
...
+    name: "Unit Tests - ${{ matrix.package }}"
     needs: [graphite-ci-optimizer, setup-checks]
     if: |
       (needs.graphite-ci-optimizer.outputs.should_skip == 'false') &&
       (needs.setup-checks.outputs.run_test == 'true')
     runs-on: ubuntu-8core
     timeout-minutes: 15
+    strategy:
+      fail-fast: false
+      matrix:
```
rjwalters, line 215: we specifically do not want to use a matrix here; please revert!

tests/test_no_xcxc.py
```diff
@@ -11,61 +11,49 @@ def test_no_xcxc_debug():
...
-        "--exclude-dir=build",
-        "--exclude-dir=.git",
-        "--exclude-dir=__pycache__",
-        "--exclude-dir=.pytest_cache",
-        "--exclude-dir=metta.egg-info",
-        "--exclude-dir=node_modules",
-        "--exclude-dir=.tox",
-        str(project_root),
-    ]
+    # Use git to get all files (tracked + untracked), respecting .gitignore
```
rjwalters, line 14: what's broken with the grep version?

rjwalters (approved):
lots of discussion in person -- if you are convinced we want a matrix of ubuntu-latest rather than an xdist of parallel background jobs on a ubuntu-8core that is okay with me

be sure to check that the codecov uploader doesn't break

also if you can keep the report of slowest tests over all all packages that i something I find useful

.github/actions/setup-uv/action.yml
zfogg: we don't actually need uv.lock for modes modes of this action. like half the uses of this action don't need the full 4GB install and they don't need uv.lock or pyproject.tomls.

the asana task shouldn't need to rebuild its cache if uv.lock changes. it has its three dependencies in its cache and they're not gonna change too much.

this solution only hashes the actual dependencies that are needed for your specific call to setup-uv

berekuk (changes requested):
- there's a lot of changes in this PR that are not related to uv caches (Robb please learn how to use graphite!)
- again, I don't think `uv pip install` is a good practice; we should be using `dependency-groups` in pyproject.toml, instead of adding more `requirements.txt`-like files (which probably don't even use `uv.lock`, so they might accidentally break in the future)
- something is broken in coverage tests - https://github.com/Metta-AI/metta/actions/runs/17477348879/job/49640367656
- I agree that `UV_NO_SYNC=1` is a good practice in CI

But my more important overall opinion is... I'm not convinced caches speed up anything.

The last run, https://github.com/Metta-AI/metta/actions/runs/17477348879/job/49640367656, fully cached, took 1m20s to do UV setup.

The last run on `main`, https://github.com/Metta-AI/metta/actions/runs/17476894461/job/49638918318, took 59s.

The previous ~10 runs on `main` took anywhere between 1m and 2m. So maybe this reduces variance? If so, I'd like to see the proof before we merge - for now my priors are that github caches might have variance too, and other failure modes as well (caches not fitting in 10G global limit and pushing each other out would be bad and can easily happen if we start caching something else too).

If you do a benchmark that will run UV setup ~20 times and it'll be clear that timings are always stable, then I'll hesitantly get on board with this change.

PR #2447, opened 2025-09-02
Reorganize navigation and navigation sequence evals, and add nav sequence training recipe

georgejwdeane:
migrating evals and recipes into dehydrated codebase, removed redundant maps (we now use navigation sequence maps and modify them to change mines and generators to altars). 

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211215462772828)


daphnedemekas (approved):
Let's merge this once we have successful aws runs

experiments/evals/navigation.py
```diff
@@ -11,14 +15,36 @@ def make_nav_eval_env(env: MettaGridConfig) -> MettaGridConfig:
...
+    atexit.register(lambda p=ascii_map_nav: os.path.exists(p) and os.remove(p))
+
+    return ascii_map_nav
+
+
 def make_nav_ascii_env(
-    name: str, max_steps: int, border_width: int = 1, num_agents=4
+    name: str,
+    max_steps: int,
+    border_width: int = 1,
```
sasmith, line 37: consider having a larger default here to ensure agents can't see each other by default

mettagrid/src/metta/mettagrid/config/envs.py
```diff
@@ -103,9 +103,43 @@ def make_arena(
 
 
 def make_navigation(num_agents: int) -> MettaGridConfig:
-    altar = building.altar.model_copy()
+    altar = empty_converters.altar.model_copy()
     altar.cooldown = 255  # Maximum cooldown
```
sasmith, line 107: I think instead we want to have the number of conversions set to zero.

sasmith (changes requested):
nav_sequence seems misconfigured to me

experiments/evals/navigation.py
```diff
@@ -91,4 +106,84 @@ def make_navigation_eval_suite() -> list[SimulationConfig]:
         ),
         SimulationConfig(name="labyrinth", env=make_nav_ascii_env("labyrinth", 250)),
         SimulationConfig(name="emptyspace_sparse", env=make_emptyspace_sparse_env()),
+        SimulationConfig(
```
sasmith, line 109: Consider: This map list looks basically the same for navigation and for nav_sequence. Are the verbatim the same? If so, can we pull the list out?

There are a few ways to do this, but my guess is that the easiest would be to have some sort of `common` file with
```
NAVIGATION_EVAL_MAPS = {
  "boxout": {"max_steps": 150},
  ...
  "easy_memory": {"map_name": "easy_sequence", "max_steps": 42, "num_agents": 2, "num_instances": 2},
  ...
}
```

And then this (and nav_sequence) can look like
```
...
return [
  SimulationConfig(
    name=name,
    env=make_nav_ascii_env(**params)
  )
  for name, params in common.NAVIGATION_EVAL_MAPS
]
```

Mostly, this would save us from having to declare this list twice, and would reduce having to deal with drift between nav and nav_sequence.

(This is non-blocking for this PR, but I think it would be good in a fast-follow.)

sasmith (approved):

PR #2448, opened 2025-09-03
refactor(sweep): refactor of new sweep infrastructure for improved clarity.

akrbc9:
refactor(sweep): refactor of new sweep infrastructure for improved clarity.

Cooling on new code


experiments/recipes/arena_basic_easy_shaped.py
```diff
@@ -112,3 +112,40 @@ def evaluate(
         simulations=simulations,
         policy_uris=[policy_uri],
     )
+
+
+def evaluate_in_sweep(
```
rjwalters, line 117: we discussed the distinction between "evaluations that happen during a sweep training run" and evaluations that happen as part of the sweep assigning a final score to some set of hyperparams... is this "evaluate_hyperparams()" or "evaluate_policy()"

experiments/sweeps/standard.py
```diff
@@ -1,19 +1,18 @@
...
-from typing import Optional
-
 from metta.sweep.protein_config import ParameterConfig, ProteinConfig, ProteinSettings
-from metta.tools.sweep_orchestrator import SweepOrchestratorTool
+from metta.tools.sweep import SweepTool
 
 
 def ppo(
-    sweep_name: Optional[str] = None,
+    run: str | None = None,  # Accept run parameter from dispatcher (unused)
```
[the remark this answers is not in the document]
akrbc9: This is just a workaround so that I can run the controller on skypilot. There's no real "run" associated with the controller. The real solution is to not require run on skypilot.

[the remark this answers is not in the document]
rjwalters: I think it would be better to generate a name for your run like "sweep_controller_{datetime}_uuid"

metta/sweep/schedulers/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""Scheduler implementations for sweep orchestration."""
+
+from metta.sweep.schedulers.optimizing import OptimizingScheduler, OptimizingSchedulerConfig
```
rjwalters, line 3: do we need this extra layer of naming? how about just "Scheduler" and "SchedulerConfig"

tests/sweep/dispatcher/test_skypilot.py
```diff
@@ -1,13 +1,33 @@
...
 
 import pytest
 
+from metta.sweep import Dispatcher, JobDefinition, JobTypes
 from metta.sweep.dispatcher.skypilot import SkypilotDispatcher
-from metta.sweep.sweep_orchestrator import Dispatcher, JobDefinition, JobTypes
+
+
+# Helper to get the expected launch script path
+def get_launch_script_path():
```
rjwalters, line 15: this could go in devops/skypilot/utils

experiments/recipes/arena_basic_easy_shaped.py
akrbc9: missed this comment on first readthrough. The way I'm intending this to be used is that we have set of hypers to sweep over, and then the sweep tool takes a train command and an eval command. 
For example `./tools/run.py experiments.recipes.sweeps.standard.ppo --args recipe=experiments.recipes.arena_basic_easy_shaped train_entrypoint=train eval_entrypoint=evaluate_in_sweep`
This way we don't have to create a new file everytimes we want to do another sweep.

PR #2461, opened 2025-09-03
Re-write trainer as a bunch of callback hooks

relh:
Trainer should just train. Everything else should not be mentioned in the trainer code.

Trainer(config: TrainerConfig, curriculum: Curriculum, policy: Policy)

Everything else should happen externally. (wandb, checkpointing, evals, stats, heartbeats [ideally perf])

### DIFF
--- a/docs/design/trainer_hooks_refactor.md
+++ b/docs/design/trainer_hooks_refactor.md
@@ -1,0 +1,372 @@
+# Trainer Hook Architecture Design Document
+
+## Executive Summary
+
+This document proposes a hook-based architecture to extract non-training concerns from the Trainer class, achieving true separation of concerns while maintaining simplicity. The design consolidates related functionality into 3 coarse-grained hooks that respect natural system boundaries.
+
+## Current State Analysis
+
+### Tightly Coupled Operations
+
+After analyzing the current trainer implementation, several operations are naturally coupled:
+
+1. **Stats + Wandb + Monitoring**
+   - Stats collection feeds directly into wandb logging
+   - Memory/system monitoring data goes through stats processing
+   - Gradient stats are computed and logged together
+   - These three are always used together when wandb is enabled
+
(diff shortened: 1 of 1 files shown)


metta/rl/hooks/base.py
```diff
@@ -0,0 +1,140 @@
...
+    __slots__ = (
+        "trainer_cfg",
+        "device",
+        "instance_name",
+        "critical",
+    )
+
+    def __init__(
+        self,
+        trainer_cfg,
```
daveey, line 25: : TrainerConfig ?

metta/rl/loss/base_loss.py
```diff
@@ -119,6 +119,14 @@ def on_mb_end(self, trainer_state: TrainerState) -> None:
     def on_train_phase_end(self, trainer_state: TrainerState) -> None:
         """We've completed the train phase and will be transitioning to the next rollout phase."""
 
+    def on_rollout_end(self, trainer_state: TrainerState) -> None:
```
daveey, line 122: Should this just subclass Hook and not overload most noop methods?

metta/rl/trainer.py
```diff
@@ -209,6 +209,9 @@ def train(
     # Instantiate configured composable losses dynamically
     loss_instances = trainer_cfg.losses.init_losses(policy, trainer_cfg, vecenv, device, checkpoint_manager)
```
daveey, line 210: _losses: list[BaseLoss]

The losses should be passed in or we should have a self._create_losses()

relh:
This is still a draft, let me make it one

metta/rl/trainer.py
daveey, line 210: losses: List[Loss]

daveey (approved):
This would be easier to review if you included at least one hook. 

I'm still not convinced we want TrainerState, train() instead of Trainer.train()

PR #2485, opened 2025-09-04
Better trainer separation of concerns

Barraketh:


relh:
Kicking back to @Barraketh

PR #2516, opened 2025-09-04
object use eval porting

bluewin4:
Adding object use maps and evals


experiments/evals/object_use.py
```diff
@@ -0,0 +1,389 @@
...
+def make_armory_use_env() -> MettaGridConfig:
+    """Test armory: ore_red -> armor"""
+    mine = building.mine_red.model_copy()
+    mine.initial_resource_count = 3
+    mine.output_resources = {"ore_red": 3}
+    mine.cooldown = 255
+
+    armory = building.armory.model_copy()
+    armory.initial_resource_count = 0
+    armory.input_resources = {"ore_red": 3}
```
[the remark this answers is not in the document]
daphnedemekas: same with the line below, basically if the recipe in the eval is the same as in buildings.py, no need to specify it

experiments/evals/object_use.py
```diff
@@ -0,0 +1,389 @@
...
+    generator = building.generator_red.model_copy()
+    generator.initial_resource_count = 2
+    generator.input_resources = {}
+    generator.output_resources = {"battery_red": 2}
+    generator.cooldown = 255
+
+    lasery = building.lasery.model_copy()
+    lasery.initial_resource_count = 0
+    lasery.input_resources = {"ore_red": 1, "battery_red": 2}
+    lasery.output_resources = {"laser": 1}
```
daphnedemekas, line 199: same thing here

experiments/evals/object_use.py
```diff
@@ -0,0 +1,389 @@
...
+        objects={"wall": building.wall, "generator_red": generator},
+        map_objects={"generator_red": 1},
+        rewards={"battery_red": 1},
+    )
+
+
+def make_lasery_use_env() -> MettaGridConfig:
+    """Test lasery: ore_red + battery_red -> laser"""
+    mine = building.mine_red.model_copy()
+    mine.initial_resource_count = 2
```
daphnedemekas, line 186: don't think initial resource count should be two here

experiments/evals/object_use.py
```diff
@@ -0,0 +1,389 @@
...
+    mine.output_resources = {"ore_red": 1}
+    mine.cooldown = 255
+
+    generator = building.generator_red.model_copy()
+    generator.initial_resource_count = 2
```
daphnedemekas, line 191: i see this in the old config, i don't think it makes sense, though, probably a leftover from issues with the recipe. if input resources is empty, and output is two batteries, that should work with no initial resources

experiments/evals/object_use.py
```diff
@@ -0,0 +1,389 @@
...
+    ascii_map: str,
+    max_steps: int,
+    objects: dict[str, ConverterConfig | WallConfig | BoxConfig],
+    rewards: dict[str, float],
+    num_agents: int = 1,
+    num_instances: int = 4,
+) -> MettaGridConfig:
+    """Create an object use evaluation environment from ASCII map."""
+
+    env = MettaGridConfig(
```
daphnedemekas, line 77: could potentially do [setattr(obj, 'cooldown', 255) for obj in objects if obj != "wall"]

 since that seems globally true, and then can remove lines of code below

experiments/evals/object_use.py
```diff
@@ -0,0 +1,389 @@
...
+        rewards={"laser": 1, "ore_red": 0, "battery_red": 0},
+    )
+
+
+def make_lasery_use_free_env() -> MettaGridConfig:
+    """Test lasery with free resources"""
+    lasery = building.lasery.model_copy()
+    lasery.initial_resource_count = 1
+    lasery.input_resources = {}
+    lasery.output_resources = {"laser": 1}
```
daphnedemekas, line 221: same here, you get the point

experiments/evals/object_use.py
```diff
@@ -0,0 +1,389 @@
...
+        objects={"wall": building.wall, "lasery": lasery},
+        map_objects={"lasery": 1},
+        rewards={"laser": 1},
+    )
+
+
+def make_mine_use_env() -> MettaGridConfig:
+    """Test mine: produces ore_red"""
+    mine = building.mine_red.model_copy()
+    mine.initial_resource_count = 3
```
daphnedemekas, line 236: don't want initial resources

experiments/evals/object_use.py
```diff
@@ -0,0 +1,389 @@
...
+        objects={"wall": building.wall, "altar": altar},
+        map_objects={"altar": 1},
+        rewards={"heart": 1},
+    )
+
+
+def make_shoot_out_env() -> MettaGridConfig:
+    """Test shooting mechanics with ASCII map"""
+    armory = building.armory.model_copy()
+    armory.initial_resource_count = 3
```
daphnedemekas, line 285: I don't think shoot out should have armory? Just lasery and walls, lasery is initialized with 20 lasers
https://github.com/Metta-AI/metta/blob/daphne-operant-conditioning-multiday/configs/env/mettagrid/object_use/evals/shoot_out.yaml

experiments/evals/object_use.py
```diff
@@ -0,0 +1,389 @@
...
+            "altar": building.altar,
+        },
+        rewards={"heart": 1},
+    )
+
+
+def make_full_sequence_env() -> MettaGridConfig:
+    """Full sequence test: mine -> generator -> altar"""
+    mine = building.mine_red.model_copy()
+    mine.initial_resource_count = 1
```
daphnedemekas, line 343: don't need initial resource count

experiments/recipes/object_use.py
```diff
@@ -0,0 +1,254 @@
...
+        # Fallback: use timestamp
+        return f"object_use.{user}.{timestamp}"
+
+
+def make_mettagrid(num_agents: int = 1, num_instances: int = 4) -> MettaGridConfig:
+    """Create a base object use environment for training."""
+
+    # Configure objects with their resource chains
+    mine = building.mine_red.model_copy()
+    mine.cooldown = 50
```
daphnedemekas, line 60: why?

experiments/recipes/object_use.py
```diff
@@ -0,0 +1,254 @@
...
+
+    # Vary initial resource counts
+    tasks.add_bucket("game.objects.altar.initial_resource_count", [0, 1])
+    tasks.add_bucket("game.objects.generator_red.initial_resource_count", [0, 1])
+
+    # Add more complex environments with all objects
+    complex_env = object_use_env.model_copy()
+
+    # Add additional object types for complex scenarios
+    temple = empty_converters.temple.model_copy()
```
daphnedemekas, line 168: I'm confused about this, why inherit temple from empty converters? Also why do we have object_use_env and complex_env? If we just add the other objects as buckets to object_use_env and span from 0 to 2, then we should have one big set of tasks

bluewin4: Honestly I'm thinking about just purging the temples entirely

experiments/evals/object_use.py
bluewin4: Lasery takes 1 battery and 2 ore

experiments/evals/object_use.py
```diff
@@ -0,0 +1,379 @@
...
+    altar.cooldown = 255
+
+    generator = building.generator_red.model_copy()
+    generator.initial_resource_count = 0
+    generator.input_resources = {"ore_red": 1}
+    generator.output_resources = {"battery_red": 1}
+    generator.cooldown = 1
+
+    mine = building.mine_red.model_copy()
+    mine.initial_resource_count = altar.input_resources.get("battery_red", 0)
```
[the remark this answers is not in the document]
bluewin4: The purpose is to start with enough resources for the final step 👎

PR #2522, opened 2025-09-05
hotfix: update uv.lock and add bazel.lock to gitignore

rjwalters:
still working on those flaky tests

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211253474238915)


sasmith:
@rjwalters -- this PR is surprising on 3 levels:
* This PR was not reviewed.
* This bulk of this PR is not what it says in the title. Most of this PR is fudging test numbers, not updating gitignore.
* Generally speaking, I would expect for lock files to be checked in. I'm only 70% sure of this, but it's consistent with wanting to have reproducible builds (e.g., for bisecting).

PR #2547, opened 2025-09-05
cooling: simplified docker slim

nishu-builder:


.github/workflows/build-image-slim-simplified.yml
berekuk, line 1: Is this necessary?

The process I've been using to test new images is:
1. make a PR
2. run build action manually
3. update local `skypilot_run.yaml` with the image tag from PR branch, run with `--skip-git-check`

devops/docker/Dockerfile.slim-simplified
```diff
@@ -0,0 +1,40 @@
+################# Base builder image: install system and non-workspace dependencies #################
+FROM nvidia/cuda:12.8.1-cudnn-devel-ubuntu24.04 AS builder
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV TORCH_CUDA_ARCH_LIST=7.5;8.0;8.6;8.9;9.0
+ENV READTHEDOCS=True
+ENV UV_LINK_MODE=copy
+ENV METTA_SYSTEM_INSTALL_DIR=/bin
```
berekuk, line 8: This is non-standard (`/bin` in posix is supposed to be for core tools only), why not `/usr/local/bin/`?

devops/docker/Dockerfile.slim-simplified
```diff
@@ -0,0 +1,40 @@
...
+
+RUN mkdir -p /workspace/metta
+WORKDIR /workspace/metta
+
+COPY .python-version pyproject.toml uv.lock ./
+COPY */pyproject.toml */
+COPY devops/tools/install-system.sh devops/tools/install-system.sh
+
+RUN bash ./devops/tools/install-system.sh
+RUN uv sync --no-install-workspace --frozen
```
berekuk, line 18: `git remote remove && git remote add && git fetch` are still useful for skypilot and I don't see them here

berekuk (changes requested):
This overlaps with my #2546, but I still haven't tested that PR and we might as well go with your version

.github/workflows/build-image-slim-simplified.yml
nishu-builder: fair enough! will remove

devops/docker/Dockerfile.slim-simplified
nishu-builder: ah was trying to copy behavior from the `COPY --from=ghcr.io/astral-sh/uv:0.8.13 /uv /uvx /bin/` that this replaces, but agree /usr/local/bin is fine; updating

devops/docker/Dockerfile.slim-simplified
nishu-builder: ah true and i think we want it after the `COPY . .` because that's when we're copying .git

PR #2551, opened 2025-09-05
fix bazel cache

rjwalters:
The CI was failing to restore Bazel cache with errors:
- "Cannot open: File exists" when extracting cached files
- Future timestamp warnings (files dated 2035)

### Solution
- Configure Bazel to use dedicated cache directories (`~/.bazel-disk-cache` and `~/.bazel-repo-cache`) instead of the default `~/.cache/bazel`
- Add these paths to `.bazelrc` to ensure Bazel uses them consistently
- Clean Bazel state when there's no cache hit to avoid conflicts

### Changes
- Added "Setup Bazel caching" step to configure cache directories
- Updated cache paths in `actions/cache@v4` to use the new locations
- Added `bazel clean --expunge` when cache miss occurs

This prevents conflicts with Bazel's internal directory structure and resolves the extraction errors.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -380,15 +380,30 @@ jobs:
         with:
           install-mode: "testing"
 
+      - name: Setup Bazel caching
+        run: |
+          # Configure Bazel to use specific cache directories
+          echo "build --disk_cache=$HOME/.bazel-disk-cache" >> .bazelrc
+          echo "build --repository_cache=$HOME/.bazel-repo-cache" >> .bazelrc
+
+          # Ensure directories exist
+          mkdir -p $HOME/.bazel-disk-cache
+          mkdir -p $HOME/.bazel-repo-cache
+
       - name: Cache Bazel build artifacts
         uses: actions/cache@v4
+        id: bazel-cache
         with:
           path: |
-            ~/.cache/bazel
-            ~/.cache/bazel-repo
+            ~/.bazel-disk-cache
+            ~/.bazel-repo-cache
           key: bazel-${{ runner.os }}-${{ hashFiles('mettagrid/MODULE.bazel', 'mettagrid/.bazelrc', '.bazelversion') }}
           restore-keys: |
             bazel-${{ runner.os }}-
+
+      - name: Clean Bazel output if no cache hit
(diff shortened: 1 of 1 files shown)


rjwalters:
> I don't understand, why does using custom directories help? And what's up with files dated 2035?
> 
> Is there a bug in actions/cache? Or in bazel?
> 
> I can approve, and I don't know much about bazel, but patching .bazelrc looks sketchy, and this deserves a comment explaining why the workaround works in the workflow.

Here's a draft response to the review comment:

---

I added a comment. I'm also not a bazel expert so this is coming from ~30 min of LLM research. I don't think it is hallucinated. Here is the summary explanation after a lot of back and forth:

## The Issue

The problem isn't with `actions/cache` or Bazel itself, but rather with how they interact:

1. **The "File exists" errors**: Bazel's default `~/.cache/bazel` directory contains its entire internal workspace state, including symlinks, lock files, and special directory structures. When `actions/cache` uses `tar` to extract this, it fails on existing symlinks and special files with "Cannot open: File exists" errors.

2. **The 2035 timestamps**: These are intentional - Bazel sets some files to future dates as a mechanism to track file freshness and invalidation. It ensures certain files are always considered "newer" than others.

## Why Custom Directories Work

By using `--disk_cache` and `--repository_cache`, we're telling Bazel to store only the actual build artifacts and downloaded dependencies in separate, clean directories. This avoids caching Bazel's complex internal state that doesn't extract well.

## This is a Standard Approach

The popular [`[bazel-contrib/setup-bazel](https://github.com/bazel-contrib/setup-bazel)`](https://github.com/bazel-contrib/setup-bazel) action uses this exact same strategy:

```yaml
- uses: bazel-contrib/setup-bazel@0.15.0
  with:
    # Store build cache per workflow
    disk-cache: ${{ github.workflow }}
    # Share repository cache between workflows
    repository-cache: true
```

They configure these as separate cache directories precisely to avoid the tar extraction issues.

I've updated the workflow comment to include links to the Bazel documentation explaining these flags. This isn't a workaround - it's the recommended approach for CI environments where you need to cache and restore Bazel artifacts reliably.

rjwalters:
we could also explore using https://github.com/bazel-contrib/setup-bazel

PR #2556, opened 2025-09-07
Evaluation Fix

harshbhatt7585:
This PR fixes the evaluation pipeline which always gives 0, 
By identifying, I found that intermediate rewards aren't initalized which was causing no final rewards also.

This adds intermediate rewards and fixes pipelines.

### DIFF
--- a/experiments/recipes/arena.py
+++ b/experiments/recipes/arena.py
@@ -146,7 +146,6 @@ def evaluate(
     policy_uri: str, simulations: Optional[Sequence[SimulationConfig]] = None
 ) -> SimTool:
     simulations = simulations or make_evals()
-    print("simulations::::", simulations)
     return SimTool(
         simulations=simulations,
         policy_uris=[policy_uri],
--- a/experiments/recipes/arena.py
+++ b/experiments/recipes/arena.py
@@ -49,16 +49,35 @@ def make_curriculum(arena_env: Optional[MettaGridConfig] = None) -> CurriculumCo
     return CurriculumConfig(task_generator=arena_tasks)
 
 
-def make_evals(env: Optional[MettaGridConfig] = None) -> List[SimulationConfig]:
+def make_evals(
+    env: Optional[MettaGridConfig] = None, num_episodes: int = 5
+) -> List[SimulationConfig]:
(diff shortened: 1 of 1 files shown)


experiments/recipes/arena.py
```diff
@@ -49,16 +49,35 @@ def make_curriculum(arena_env: Optional[MettaGridConfig] = None) -> CurriculumCo
...
 
 
-def make_evals(env: Optional[MettaGridConfig] = None) -> List[SimulationConfig]:
+def make_evals(
+    env: Optional[MettaGridConfig] = None, num_episodes: int = 5
+) -> List[SimulationConfig]:
     basic_env = env or make_mettagrid()
     basic_env.game.actions.attack.consumed_resources["laser"] = 100
 
+    # Add intermediate rewards to help the agent succeed
```
rjwalters, line 58: I don't understand why this matters in evaluation... i can see how these breadcrumbs would help learning in training but once the policy is trained we ought to be able to evaluate it and score however we want. My concern is that this might point at some deeper problem that this is hiding.

experiments/recipes/arena.py
```diff
@@ -113,7 +125,7 @@ def train_shaped(rewards: bool = True, converters: bool = True) -> TrainTool:
         losses=LossConfig(),
         curriculum=cc.env_curriculum(env_cfg),
         evaluation=EvaluationConfig(
-            simulations=make_evals(env_cfg),
+            simulations=make_evals(env_cfg, num_episodes=5),
```
rjwalters, line 128: why are we changing to 5 episodes? (just curious)

rjwalters (changes requested):
why do we need intermediate rewards in evaluation?

experiments/recipes/arena.py
harshbhatt7585: I was testing whether the score will get better by having multiple episode, I will change it to 1.

experiments/recipes/arena.py
harshbhatt7585: checkout eng-help channel, I have posted about this.

harshbhatt7585:
figured out, Evals are working! I am closing this.

PR #2561, opened 2025-09-08
Feature: Run/Evalaute Pufferlib Checkpoint in Metta

harshbhatt7585:
This PR adds pufferlib maps state dict to compatible metta's agent, help us to run puffer checkpoint and evalaute in metta. 

It changes/adds following:
1. An policy architecture which is identical to puffer policy
2. Autodetection of puffer checkpoint
3. Support to evalaute puffer checkpoint in metta


metta/tools/train.py
```diff
@@ -7,7 +7,7 @@
 import torch
 from pydantic import Field
 
-from metta.agent.policies.fast import FastConfig
+from metta.agent.policies.puffer import PufferPolicyConfig
```
relh, line 10: we don't want to change metta/tools/train's default policy, this is the core of the repo. We want to just pass in the correct policy_architecture, or put it in an experiments/recipes/puffer.py

metta/rl/puffer_policy.py
```diff
@@ -1,63 +1,134 @@
...
+        if key in policy_state:
+            target_param = policy_state[key]
+            if target_param.shape == value.shape:
+                compatible_state[key] = value
+                keys_matched += 1
+            else:
+                shape_mismatches.append(f"{key}: checkpoint {value.shape} vs policy {target_param.shape}")
+                logger.debug(f"Shape mismatch for {key}: checkpoint {value.shape} vs policy {target_param.shape}")
+        else:
+            logger.debug(f"Skipping unmatched parameter: {key}")
```
relh, line 105: we don't want this. If any of the keys fail to load we want to crash. We don't want to have a neural network that's only partially built.

metta/rl/puffer_policy.py
```diff
@@ -1,63 +1,134 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def _is_puffer_state_dict(loaded_obj: Any) -> TypeGuard[Dict[str, torch.Tensor]]:
+    """Return True if the object appears to be a PufferLib state_dict."""
+    return isinstance(loaded_obj, dict) and bool(loaded_obj) and any(key.startswith("policy.") for key in loaded_obj)
+
+
+def _preprocess_state_dict(state_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
```
relh, line 24: why do we need this function? We should ideally be able to just name the weights in the metta policy to be the same as in the puffer. 

Pytorch does this by default depending on what the name of the class is. Delete this function and see if you can still get things to work

relh (changes requested):
Much better! We're almost there

metta/rl/puffer_policy.py
harshbhatt7585: most of the keys can be detected automatically but for lstm we need to do mapping as in metta we are have different strcture of LSTM so it can't map automatically.

relh (approved):

PR #2565, opened 2025-09-09
stochastic resource loss

badjacent:


rjwalters:
this seems like a feature we should be removing entirely

sasmith (commented):
Thanks for putting this together. I think this is heavier weight than we want for this feature, so I'm going to put this on ice for now.

badjacent:
> Thanks for putting this together. I think this is heavier weight than we want for this feature, so I'm going to put this on ice for now.

Just want to point out the reason for the complexity. From the main asana task: "We want to avoid this being O(objects)." The implementation if we allow O(objects) is straightforward and simple. Otherwise indexing of some sort is necessary and it's more complicated. More details on the Asana task if you are interested.

https://app.asana.com/0/home/1210755368917341/1211216072547179

PR #2568, opened 2025-09-10
mettagrid tags

nathants:
https://app.asana.com/1/1209016784099267/project/1209041170403474/task/1209208922660223

### DIFF
--- a/mettagrid/src/metta/mettagrid/char_encoder.py
+++ b/mettagrid/src/metta/mettagrid/char_encoder.py
@@ -55,19 +55,19 @@ def char_to_grid_object(char: str) -> str:
     """
     if not char:
         raise ValueError("Object name cannot be empty")
-    
+
     # Check for known single-character aliases first
     if char in CHAR_TO_NAME:
         return CHAR_TO_NAME[char]
 
     if char in NAME_TO_CHAR:
         return char
-    
+
     # Now check for invalid patterns in multi-character names
     # Reject trailing dots that would create empty tag segments
     if char.endswith("."):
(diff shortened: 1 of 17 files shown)


sasmith (changes requested):
I think this is not what we're looking for. Right now you're parsing tags in cpp based on the name of objects in the map. Instead, any interesting tag related logic should happen in Python, and cpp should get a pretty-darn-straightforward configuration of tags.

So working backwards:
The configs to go to cpp should have an integer list (vector) of tags. Cpp may need to figure out an offset for these in order to avoid collisions in feature space. We should probably also send a map of <tag_id, tag_name> to cpp so we can expose it for rendering / debugging, but these strings shouldn't impact agent observations, and they should be pretty black-box from the environment's perspective.

At the Python level, people should be specifying tags as strings. The string -> id should happen in mettagrid_c_config.py.

People should specify tags explicitly. In the task, we talked about automatically parsing tags from names, but let's not do that. For at-least-this-version, let's make people be explicit about tags. This would mean updating mettagrid_config pydantic classes to make this happen.

The general philosophy here is that the cpp config should be "pre digested" -- it should be very easy for cpp to use, and we don't care that much how easy it is for humans to use (e.g., it can be super verbose). In contrast, we care about usability within Python. And mettagrid_c_config handles the translation from "easy to write and understand by humans" to "easy to digest in cpp". This lets us keep various parsing logic out of cpp.

It would be great to bring these together. Some of the reason they're structured how they are is that Python dicts can (for example) have mixed value types, whereas cpp maps can't. So we built a bridging layer, and then kept leaning on it. Similarly for strings vs numeric ids.

Anyways, please lean on the bridge layer, and keep the cpp code clean.

PR #2574, opened 2025-09-10
rm .metta in clean

rjwalters:
Recently @daphnedemekas encountered a configuration issue on her laptop where an older file at "~/.metta/observatory_tokens.yaml" seems to have broken training. This PR makes the `metta clean` cmd more aggressive.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211319462606027)


nishu-builder (changes requested):
Right now metta clean runs before every install. Could we have removing observatory tokens gated by something else? Maybe: --reset, or only happens on install --force, etc.

rjwalters:
> Somehow we want clean and reset as two different things, where clean is attempting to be nondestructive (removes cached files that are trivially regenerated without human intervention). I'm open to however we do that

I think I'd like to make this safe enough to run every time... I am keeping the config.yaml file -- what else goes in `~/.metta`? Is there a downside to renewing the observatory token?

nishu-builder (changes requested):
i think this was the issue https://github.com/Metta-AI/metta/pull/2575

also explains why the failure was silent: `metta status` succeeded for her because her key did auth successsfully with observatory; it wasn't that it was a token in an old format. it was that it specifically didn't work with a skypilot-only command. so i think we dont have a reason to clean out ~/.metta anymore

rjwalters:
> i think this was the issue #2575
> 
> also explains why the failure was silent: `metta status` succeeded for her because her key did auth successsfully with observatory; it wasn't that it was a token in an old format. it was that it specifically didn't work with a skypilot-only command. so i think we dont have a reason to clean out ~/.metta anymore

ok - I will close this for now then!

PR #2575, opened 2025-09-10
fix passing of observatory tokens to create_job_secrets in skypilot jobs

nishu-builder:
Daphne ran into a very confusing issue where her observatory token existed locally, `metta status` said it was fine, and she kept getting issues in skypilot jobs indicating the arg was missing or malformed

It was because her token had a hyphen prefixing it, and our previous way of passing that to create_job_secrets within skypilot runs didn't escape that

This change should fix the passing of such tokens

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211320319176404)


rjwalters:
I think it is slightly better to fix the cli parsing?

PR #2598, opened 2025-09-12
friction: Extract configs from demo files 

teodorionita:
Extracted the configurations from the demo files in a single config file to maintain DRY principles 

Task: https://app.asana.com/1/1209016784099267/project/1209805539287945/task/1210863996775022?focus=true

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211341654539908)


daveey (changes requested):
Sorry, this task was mostly for when configs were giant json blobs. Now, having them in the demo file is totally fine. That said, it is nice to define all the parameters in one place. But we should not be using a DemoConfig and having pettingzoo_.... gym_... etc. Those should be separate configs:

GymDemoConfig() etc...

teodorionita:
@daveey awesome then I can close the PR if they're ok, I've rewritten them when we did the refactoring and didn't know if I can close this or not :D will close and keep them as they are!

PR #2618, opened 2025-09-15
add metric score flag + default to env_agent/heart.get + remove cost warning 

akrbc9:
## What

  - Replaced hardcoded observation.score dependency with
  configurable --score-metric flag
  - Added --score-metric CLI argument with default value
  "env_agent/heart.get"
  - Updated score display to use run.summary[score_metric]
  instead of run.observation.score
  - Modified banner to show which metric is being used for
  scoring
  - Removed observation dependency from test mode mock data
  - Cleaned up unused Observation import

## Why

  The live monitor was previously dependent on the observation
  field in RunInfo.summary, which typically only exists for sweep
   runs. This made the monitor less useful for regular training
  runs where scores are stored directly in the run summary under
  various metric keys.

  By making the score metric configurable, the monitor becomes
  more flexible and can display scores from any metric stored in
  run.summary, making it useful for both sweep and non-sweep
  runs.

 ## Testing

  - Updated test mode to generate mock summary data with the
  configurable score metric
  - Verified linting passes with metta lint --fix
  - Banner now displays the score metric being used: "Score:
  env_agent/heart.get"
  - Monitor gracefully handles missing metrics by showing "N/A"

## Breaking Changes

  None. The monitor maintains backward compatibility and will
  show "N/A" for scores when the specified metric key doesn't
  exist in the run summary.

### DIFF
--- a/metta/utils/live_run_monitor.py
+++ b/metta/utils/live_run_monitor.py
@@ -63,7 +63,7 @@ def _get_status_color(status: JobStatus) -> str:
         return "white"
 
 
-def make_rich_monitor_table(runs: List[RunInfo]) -> Table:
+def make_rich_monitor_table(runs: List[RunInfo], score_metric: str = "env_agent/heart.get") -> Table:
     """Create rich table for run monitoring."""
 
     # Create table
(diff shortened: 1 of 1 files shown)


nishu-builder (approved):
nice!

PR #2622, opened 2025-09-16
add notifications system

jacklionheart:
add notifications system

add test workflow

add job config

add test

fix branch and main sha collection

add python path

try runtime install

install metta.common separately

allow notifications on new runs

add rich dependency

update uv.lock

install gitta

add wandb to common

Update uv.lock

use full commit sha

switch to manual

add enum

Update .github/scripts/skypilot_notifications_test.py

Co-authored-by: graphite-app[bot] <96075541+graphite-app[bot]@users.noreply.github.com>

rm comment


.github/scripts/skypilot_notifications_test.py
```diff
@@ -0,0 +1,189 @@
...
+    with open("/tmp/notification_test_results.md", "w") as f:
+        f.write(summary_text)
+
+    # Also print to console
+    print(summary_text)
+
+
+def main():
+    """Run notification tests."""
+    requested_scenario = os.environ.get("TEST_SCENARIO", "all")
```
berekuk, line 152: This supports `"all"` but there's no way to set it in workflow. Also, the list of `scenarios` below doesn't match the list in the workflow.

.github/scripts/skypilot_notifications_test.py
```diff
@@ -0,0 +1,189 @@
...
+    # Override defaults if provided
+    if repo := os.environ.get("GITHUB_REPOSITORY"):
+        config.github_repository = repo
+    config.github_status_context = f"test/notifications/{scenario}"
+
+    # W&B configuration
+    if entity := os.environ.get("WANDB_ENTITY"):
+        config.wandb_entity = entity
+    if project := os.environ.get("WANDB_PROJECT"):
+        config.wandb_project = project
```
berekuk, line 43: Minor, but I don't think `wandb_entity`, `wandb_project` and `github_repository` are useful, unless we expect that people will do full forks of metta (maybe external contributors? but they don't have skypilot or credentials and they rarely will bother installing it just to test it in CI).

Also note that `skypilot/notifications/github.py` is inconsisent and uses common constants instead of these vars for wandb.

devops/skypilot/notifications/notification.py
```diff
@@ -0,0 +1,111 @@
...
+    130: "Script terminated by Ctrl-C (SIGINT)",
+    131: "Core dumped (SIGQUIT)",
+    134: "Assertion failed (SIGABRT)",
+    135: "Process terminated (SIGTRAP)",
+    136: "Floating point exception (SIGFPE)",
+    137: "Process killed (SIGKILL) - likely OOM",
+    139: "Segmentation fault (SIGSEGV)",
+    140: "Bad system call (SIGSYS)",
+    141: "Broken pipe (SIGPIPE)",
+    143: "Process terminated (SIGTERM)",
```
berekuk, line 66: Looks like some of these are different on macos vs linux, e.g. code 140.

Linux:
```
>>> signal.strsignal(140-128)
'User defined signal 2'
```

macOS:
```
>>> signal.strsignal(140-128)
'Bad system call: 12'
```

Safer to rely on `signal.strsignal(code-128)` and `signal.Signals(code-128)` for codes >128.

berekuk (approved):
Cool, this addresses my main issue with #2594.

You probably should fix status codes before the merge.

PR #2624, opened 2025-09-16
Make axel's live run monitor available as `metta run-monitor`

nishu-builder:
Also:
- move it to being a typer app instead of argparse
- remove unneeded carefulness around wandb import
- move around some imports so that gymnasium isn't imported unless metta run-monitor is actually running; this slows down the metta cli otherwise

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211368595727364)

### DIFF
--- a/metta/utils/live_run_monitor.py
+++ b/metta/utils/live_run_monitor.py
@@ -30,7 +30,6 @@ from datetime import datetime
 from typing import Annotated, List, Optional
 
 import typer
-import wandb
 from rich.console import Console, Group
 from rich.live import Live
 from rich.table import Table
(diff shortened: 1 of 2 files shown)


metta/utils/live_run_monitor.py
```diff
@@ -23,28 +23,42 @@
...
 from rich.console import Console, Group
 from rich.live import Live
 from rich.table import Table
 from rich.text import Text
 
-from metta.sweep.models import JobStatus, RunInfo
+from metta.common.util.constants import METTA_WANDB_ENTITY, METTA_WANDB_PROJECT
+
+if TYPE_CHECKING:
+    from metta.sweep.models import JobStatus, RunInfo
```
akrbc9, line 42: As these data models are making their way into the codebase, we should be thinking about moving them out of sweeps.

PR #2628, opened 2025-09-16
implement contrastive loss

tashapais:
Using Contrastive Loss on Any Recipe

  1. Enable with default parameters:
  uv run ./tools/run.py train arena run=my_experiment trainer.losses.enable_contrastive=true

  2. Enable with custom hyperparameters:
  uv run ./tools/run.py train arena run=my_experiment \
    trainer.losses.enable_contrastive=true \
    trainer.losses.contrastive_temperature=0.1 \
    trainer.losses.contrastive_coef=0.2

  3. Full control over all contrastive parameters:
  uv run ./tools/run.py train arena run=my_experiment \
    trainer.losses.enable_contrastive=true \
    trainer.losses.contrastive_temperature=0.1 \
    trainer.losses.contrastive_coef=0.2 \
    trainer.losses.contrastive_embedding_dim=256 \
    trainer.losses.contrastive_use_projection_head=false

  4. Works with any recipe:
  Arena basic easy shaped: uv run ./tools/run.py train arena_basic_easy_shaped run=test trainer.losses.enable_contrastive=true
  Navigation: uv run ./tools/run.py train navigation run=test trainer.losses.enable_contrastive=true
  Any other recipe: uv run ./tools/run.py train <recipe_name> run=test trainer.losses.enable_contrastive=true

The key insight is that all configuration overrides go through trainer.losses.* regardless of which recipe you're using.

Hypers updated are temperature and contrastive coefficient in contrastive_config.py. Hypers that I didn't add but should (they will mess with ppo core): entropy=0.02891525626182556, learning rate=0.0006680303486064076

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211376368229881)


tashapais:
Critical Bugs Fixed

  1. Fixed Negative Similarity Computing Identical Tensors

  - Problem: neg_sim = F.cosine_similarity(embeddings[:-1],
  neg_embeddings, dim=-1) where neg_embeddings = embeddings[:-1]
  (identical tensors)
  - Result: Negative similarity was always 1.0, breaking contrastive
  learning
  - Fix: Proper negative sampling using torch.randperm() with
  collision avoidance

  2. Fixed Improper Positive/Negative Pair Creation

  - Problem: Both positive and negative pairs were consecutive
  timesteps, not true contrastive pairs
  - Fix:
    - Positive pairs: consecutive timesteps (temporal continuity)
    - Negative pairs: randomly shuffled embeddings (different
  contexts)
    - Added collision detection to ensure negatives ≠ positives

  3. Fixed Value Expansion Memory/Learning Issues

  - Problem: expand(-1, embedding_dim) created identical values
  across embedding dimensions
  - Result: Memory bloat + defeats contrastive learning purpose
  - Fix: Created learnable Linear(1, embedding_dim) projection with
  Xavier initialization

  4. Fixed Config Inconsistency

  - Problem: ContrastiveConfig() always in loss_configs even when
  enable_contrastive=False
  - Fix:
    - Changed default enable_contrastive=False
    - Dynamic config addition via __post_init__()
    - Explicit warnings for inconsistent configurations

  5. Fixed Matrix Dimension Mismatch

  - Original Issue: Dynamic projection head creation based on actual
  embedding dimensions
  - Result: No more RuntimeError: mat1 and mat2 shapes cannot be
  multiplied

tashapais:
Training Progress: You can see the contrastive
  learning working properly:
    - Positive similarities staying high (~0.89-0.90)
    - Negative similarities decreasing over time (0.92 →
  0.26 → 0.14)
    - Contrastive loss decreasing (0.89 → 0.36)
    - Number of pairs: 16,383 per minibatch

experiments/recipes/arena_with_contrastive_sparse_rewards.py
```diff
@@ -0,0 +1,178 @@
...
+    enable_detailed_slice_logging: bool = False,
+    enable_contrastive: bool = False,
+) -> TrainTool:
+    """Train with sparse rewards and optional contrastive loss."""
+    curriculum = curriculum or make_curriculum(
+        enable_detailed_slice_logging=enable_detailed_slice_logging
+    )
+
+    # Create trainer config with contrastive loss enabled + PPO for actions
+    from metta.rl.loss.contrastive_config import ContrastiveConfig
```
relh, line 109: imports at the top of files please

metta/rl/loss/contrastive.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # Track additional metrics for wandb logging
+        for key, value in metrics.items():
+            if key not in self.loss_tracker:
+                self.loss_tracker[key] = []
+            self.loss_tracker[key].append(value)
+
+        return contrastive_loss, shared_loss_data, False
+
+    def _get_embeddings(self, policy_td: TensorDict) -> Tensor:
```
relh, line 93: Would it make sense for someone initializing the contrastive loss to specify what key from their policy_td the contrastive loss should be run with?

metta/rl/loss/contrastive.py
```diff
@@ -0,0 +1,258 @@
...
+    def _compute_contrastive_loss(self, embeddings: Tensor, minibatch: TensorDict) -> tuple[Tensor, dict]:
+        """Compute InfoNCE contrastive loss with geometric future positives and shuffled negatives."""
+
+        batch_shape = minibatch.batch_size
+        if len(batch_shape) != 2:
+            raise ValueError("Contrastive loss expects minibatch with 2D batch size (segments, horizon).")
+
+        segments, horizon = batch_shape
+
+        if embeddings.dim() == 2 and embeddings.shape[0] == segments * horizon:
```
relh, line 134: There's some ugly shape checking and resizing here. Which of these is typically the case for the embeddings dimensionality and can we just always use the embeddings dimensionality that always is? I guess if we're checking 4-5 keys for one to choose to make the contrastive loss then we need this

metta/rl/loss/loss_config.py
```diff
@@ -29,14 +30,26 @@ def model_post_init(self, __context: Any) -> None:
 
             self.loss_configs = {"ppo": PPOConfig()}
 
+        # Add contrastive config only if enabled to avoid inconsistent behavior
+        if self.enable_contrastive and "contrastive" not in self.loss_configs:
+            # Import here to avoid circular dependency
+            from metta.rl.loss.contrastive_config import ContrastiveConfig
```
relh, line 36: imports to top if possible. I think there's a `.pyi` file that can help resolve circular imports. Not sure how the other losses avoid this issue

relh (changes requested):
Few small changes

metta/rl/loss/contrastive.py
tashapais: Most policies in this repo emit embeddings aligned with the minibatch layout coming from replay—[segments, horizon, embedding_dim]. The only time we see [segments * horizon, embedding_dim] is when something flattens trajectories (e.g., a legacy data loader). So the 3-D shape is the common case; the 2-D variant exists just to stay compatible with older paths. I changed it to do embedding reshaping inline like the other loss files.

metta/rl/loss/contrastive.py
tashapais: ContrastiveLoss now reads an optional embedding_key from its config; when set we pull that tensor from policy_td and skip the heuristic fallbacks.

metta/rl/loss/loss_config.py
tashapais: Used the package's lazy loader to avoid circular dependencies.

relh (commented):
Looks good! Just imports should be good enough

rjwalters:
@claude review einops

experiments/recipes/arena_with_sparse_rewards.py
```diff
@@ -0,0 +1,178 @@
...
+    enable_detailed_slice_logging: bool = False,
+    algorithm_config: Optional[CurriculumAlgorithmConfig] = None,
+) -> CurriculumConfig:
+    """Create curriculum with sparse reward environment."""
+    arena_env = arena_env or make_mettagrid()
+
+    arena_tasks = cc.bucketed(arena_env)
+
+    # Only vary heart rewards (final objective) in curriculum
+    arena_tasks.add_bucket("game.agent.rewards.inventory.heart", [0.5, 1.0, 2.0])
```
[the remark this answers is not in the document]
tashapais: Varying game.agent.rewards.inventory.heart introduces curriculum slices where the final reward signal’s magnitude changes. That lets the learning-progress scheduler probe which reward scale best drives exploration and stabilizes learning under sparse feedback: lower values emphasize survival skills; higher values push agents to focus on the heart objective. Keeping it in the buckets gives the curriculum levers to measure progress and adjust training difficulty automatically. @mstormbull can you verify?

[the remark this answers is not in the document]
tashapais: inventory.heart is the per-agent reward weight: how much return we give for each heart in the inventory. inventory_max.heart is the capacity cap—the maximum number of hearts an agent can hold (and thus get credit for) before the count stops increasing.

metta/rl/loss/__init__.py
```diff
@@ -10,6 +10,7 @@
 
 # Type imports only happen during type checking, not at runtime
 if TYPE_CHECKING:
```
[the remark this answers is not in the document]
tashapais: ok will do a followup

relh (approved):

PR #2640, opened 2025-09-16
Relax python version requirement for mettagrid

dominik3141:
Now just requiring Python version 3.11.x instead of 3.11.7.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211378286130559)

### DIFF
--- a/mettagrid/pyproject.toml
+++ b/mettagrid/pyproject.toml
@@ -5,10 +5,10 @@ backend-path = ["."]
 
 [project]
 name = "mettagrid" # Do not change! Has to be 'mettagrid' for our PyPi package
-version = "0.2.0.1"
+version = "0.2.0.2"
 description = "A fast grid-based open-ended MARL environment"
 authors = [{ name = "David Bloomin", email = "daveey@gmail.com" }]
-requires-python = "==3.11.7"
+requires-python = ">=3.11,<3.12"
 license = "MIT"
 readme = "README.md"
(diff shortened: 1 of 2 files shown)


mettagrid/pyproject.toml
```diff
@@ -5,10 +5,10 @@ backend-path = ["."]
 
 [project]
 name = "mettagrid" # Do not change! Has to be 'mettagrid' for our PyPi package
-version = "0.2.0.1"
+version = "0.2.0.2"
 description = "A fast grid-based open-ended MARL environment"
 authors = [{ name = "David Bloomin", email = "daveey@gmail.com" }]
-requires-python = "==3.11.7"
+requires-python = ">=3.11,<3.12"
```
sasmith, line 11: Certainly a reasonable improvement.

Is this a system-level requirement, or is this something uv handles? If it's something uv is handling, then I like the semi-tight requirement we have here. If this is system-level python, then I'd want us to be looser.

sasmith (approved):

mettagrid/pyproject.toml
dominik3141: UV handles it for us in metta, but external people that just want to install mettagrid from PyPi might not use UV. So long term I would like to relax this even further.
However, stuff starts breaking as soon as we allow 3.12, so there is some more work to do.

PR #2644, opened 2025-09-16
Port thought bubbles from TS to Nim

lcodes:
Also removed string concats in draw path while at it :)


treeform:
Looks great!

image

treeform (approved):

PR #2655, opened 2025-09-17
move testing deps used in ci into dependency group

nishu-builder:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211381511388764)


rjwalters:
can we instead do #2658

PR #2667, opened 2025-09-17
drawing agent actions in nim

lcodes:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211389356415479)


mettascope2/src/mettascope/worldmap.nim
```diff
@@ -22,10 +22,10 @@ proc useSelections*(panel: Panel) =
       gridPos = (mousePos + vec2(0.5, 0.5)).ivec2
     if gridPos.x >= 0 and gridPos.x < replay.mapSize[0] and
       gridPos.y >= 0 and gridPos.y < replay.mapSize[1]:
-        for obj in replay.objects:
```
lcodes, line 25: hrm i think we have different "format on save" rules heh, i didnt intent to change these

treeform: Its fine.

PR #2669, opened 2025-09-17
Adds an assembler object

sasmith:
This adds Recipes, and Assemblers which can expose different recipes depending on how their surrounded by agents. Assemblers keep a vector of 256 recipes, which may be repeats. This keeps the cpp logic of which recipe to use simple.

Right now resources are consumed in a consistent but arbitrary order.

This includes a translation of Arena to use Assemblers instead of Converters, and Assemblers appear to work in this context. I haven't produced more thorough unit tests of this, but will do so reactively as it becomes more clear what sort of testing we need.

https://app.asana.com/1/1209016784099267/project/1211340635624763/task/1211355235230371

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211390250279726)


sasmith:
* **#2669** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2669?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2669?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mettagrid/src/metta/mettagrid/mettagrid_c_config.py
```diff
@@ -18,6 +24,44 @@ def recursive_update(d, u):
     return d
 
 
+def expand_position_patterns(positions: list[str]) -> list[int]:
+    """Expand position patterns into corresponding byte patterns.
```
daveey, line 28: convert from string positions to corresponding bitmasks

mettagrid/src/metta/mettagrid/objects/assembler.hpp
```diff
@@ -0,0 +1,221 @@
+#ifndef OBJECTS_ASSEMBLER_HPP_
+#define OBJECTS_ASSEMBLER_HPP_
+
+#include <map>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include "../event.hpp"
```
daveey, line 9: can we make them absolute?

mettagrid/src/metta/mettagrid/objects/assembler.hpp
```diff
@@ -0,0 +1,221 @@
...
+// Forward declaration
+class Agent;
+
+class Assembler : public Usable {
+private:
+  // Surrounding positions in deterministic order: NW, N, NE, W, E, SW, S, SE
+  std::vector<std::pair<GridCoord, GridCoord>> get_surrounding_positions() const {
+    GridCoord r = location.r;
+    GridCoord c = location.c;
+    return {
```
daveey, line 28: consider loop

daveey (approved):

mettagrid/src/metta/mettagrid/objects/assembler.hpp
sasmith: I think, sort of, no. Or at least, not without doing more work to change paths.

`event.hpp` is at the root of `mettagrid`, so right how our options for this are:
* `#include "../event.hpp"`
* `#include "event.hpp"`
* `#include "/Users/sasmith/code/metta/mettagrid/event.hpp"`

The last one is obviously useless. Additionally, I think `#include "event.hpp"` is crummy since it looks like a local include.

With a bigger refactor, we could do one of
* make it so no files are in mettagrid's root. Like, they could all be in a `source` directory. Then "absolute" imports would be obvious.
* make everything use "absolute" imports, so here we'd have `#include "event.hpp"`, but the local imports would become `#include "objects/agent.hpp"`. Then local and absolute wouldn't conflict.

PR #2687, opened 2025-09-18
fix: simplify mettagrid release ci

rjwalters:
This PR simplifies the mettagrid release workflow by using `cibuildwheel` to handle the complexity of building Python wheels across multiple platforms.

- **Replace manual wheel building with `cibuildwheel`**: Removes 150+ lines of manual Docker commands and wheel building logic
- **Consolidate build jobs**: Reduces from 5 jobs to 3 (sdist, wheels, publish)
- **Simplify artifact management**: Uses `merge-multiple` for cleaner artifact handling
- **Improve workflow inputs**: Clearer descriptions and proper choice types for the dispatch inputs
- **Remove Windows job**: The Windows build was already disabled, so it's been removed entirely
- Still builds only CPython 3.11 wheels as required
- Maintains manylinux_2_28 compatibility 
- Preserves Bazel installation for both Linux and macOS
- Keeps the same publishing flow to TestPyPI/PyPI

### ✅ **Preserved functionality**
- Python 3.11.7 specific builds
- Source distribution (sdist) build
- Linux x86_64 wheels with manylinux_2_28
- macOS x86_64 (macos-13) and arm64 (macos-14) wheels  
- Bazel installation on all platforms
- CI environment variable for Bazel
- Twine check before publishing
- Trusted publishing to TestPyPI/PyPI
- Skip musllinux builds
- Skip-existing for TestPyPI

### ⚠️ **Minor differences**
1. **Wheel verification removed**: The original checked for unwanted universal wheels (`*py3-none-any.whl`) and plain linux wheels (`*-linux_*.whl`). This is likely safe to remove as `cibuildwheel` won't produce these, but you could add it back if desired:
   ```yaml
   - name: Verify wheel tags
     run: |
       if ls dist/*py3-none-any.whl 2>/dev/null; then
         echo "Error: Found universal wheels"; exit 1
       fi
   ```

2. **Windows builds removed entirely**: The original had Windows disabled with `if: false`, so removing it is cleaner

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211399824899351)


dominik3141 (changes requested):
Doesn't work currently. 

Image

dominik3141:
From what I understand this PR mostly moves our hand-written logic to some external Github action, which seems like a good idea to me. But the whole logic to relax the python version during the build has become obsolete due to #2640. @rjwalters Are you fine with me just "intruding" your branch and removing those parts?

dominik3141 (approved):

PR #2688, opened 2025-09-18
mettagrid tags simpler

nathants:
Abandoning a previous [PR](https://github.com/Metta-AI/metta/pull/2568#pullrequestreview-3236567754) and reworking it, starting with a simpler foundation of tags that do nothing except be observed. 

Once we get this right, we can think about tags modifying object properties, and other tagging behaviors. 

@sasmith let me know what you think.

https://app.asana.com/1/1209016784099267/project/1209041170403474/task/1209208922660951

### DIFF
--- a/packages/mettagrid/cpp/include/mettagrid/core/grid_object.hpp
+++ b/packages/mettagrid/cpp/include/mettagrid/core/grid_object.hpp
@@ -51,8 +51,10 @@ public:
 struct GridObjectConfig {
   TypeId type_id;
   std::string type_name;
+  std::vector<int> tag_ids;
 
-  GridObjectConfig(TypeId type_id, const std::string& type_name) : type_id(type_id), type_name(type_name) {}
+  GridObjectConfig(TypeId type_id, const std::string& type_name, const std::vector<int>& tag_ids)
+      : type_id(type_id), type_name(type_name), tag_ids(tag_ids) {}
 
   virtual ~GridObjectConfig() = default;
 };
@@ -63,13 +65,15 @@ public:
   GridLocation location{};
   TypeId type_id{};
   std::string type_name;
+  std::vector<int> tag_ids;
 
   virtual ~GridObject() = default;
 
-  void init(TypeId object_type_id, const std::string& object_type_name, const GridLocation& object_location) {
+  void init(TypeId object_type_id, const std::string& object_type_name, const GridLocation& object_location, const std::vector<int>& tags) {
     this->type_id = object_type_id;
     this->type_name = object_type_name;
     this->location = object_location;
+    this->tag_ids = tags;
   }
 
   virtual bool swappable() const {
--- a/packages/mettagrid/cpp/include/mettagrid/objects/agent.hpp
+++ b/packages/mettagrid/cpp/include/mettagrid/objects/agent.hpp
@@ -66,7 +66,7 @@ public:
         prev_action_name(""),
         steps_without_motion(0) {
     populate_initial_inventory(config.initial_inventory);
-    GridObject::init(config.type_id, config.type_name, GridLocation(r, c, GridLayer::AgentLayer));
+    GridObject::init(config.type_id, config.type_name, GridLocation(r, c, GridLayer::AgentLayer), config.tag_ids);
   }
 
   void init(RewardType* reward_ptr) {
@@ -192,7 +192,7 @@ public:
   }
 
   std::vector<PartialObservationToken> obs_features() const override {
-    const size_t num_tokens = this->inventory.size() + 5 + (glyph > 0 ? 1 : 0);
+    const size_t num_tokens = this->inventory.size() + 5 + (glyph > 0 ? 1 : 0) + this->tag_ids.size();
 
     std::vector<PartialObservationToken> features;
     features.reserve(num_tokens);
@@ -209,6 +209,11 @@ public:
       assert(amount > 0);
       auto item_observation_feature = static_cast<ObservationType>(InventoryFeatureOffset + item);
       features.push_back({item_observation_feature, static_cast<ObservationType>(amount)});
+    }
+
+    // Emit tag features
+    for (int tag_id : tag_ids) {
+      features.push_back({ObservationFeature::Tag, static_cast<ObservationType>(tag_id)});
     }
 
     return features;
--- a/packages/mettagrid/cpp/include/mettagrid/objects/constants.hpp
+++ b/packages/mettagrid/cpp/include/mettagrid/objects/constants.hpp
@@ -47,7 +47,8 @@ constexpr ObservationType LastReward = 10;
 constexpr ObservationType Glyph = 11;
 constexpr ObservationType ResourceRewards = 12;
 constexpr ObservationType VisitationCounts = 13;
-constexpr ObservationType ObservationFeatureCount = 14;
+constexpr ObservationType Tag = 14;
+constexpr ObservationType ObservationFeatureCount = 15;
 }  // namespace ObservationFeature
 
 const ObservationType InventoryFeatureOffset = ObservationFeature::ObservationFeatureCount;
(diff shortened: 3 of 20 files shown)


packages/mettagrid/cpp/include/mettagrid/core/grid_object.hpp
```diff
@@ -72,6 +75,10 @@ class GridObject {
     this->location = object_location;
   }
 
+  void set_tags(const std::vector<int>& tags) {
```
sasmith, line 78: probably don't need this? I'd expect tags to be constant.

sasmith: To clarify: tags should be constant. Try to set them in the constructor if you can. If that ends up being messy, then setting them here is fine.

packages/mettagrid/cpp/include/mettagrid/objects/agent.hpp
```diff
@@ -211,6 +212,11 @@ class Agent : public GridObject, public HasInventory {
       features.push_back({item_observation_feature, static_cast<ObservationType>(amount)});
     }
 
+    // Emit tag features
+    for (int tag_id : tag_ids) {
+      features.push_back({ObservationFeature::Tag, static_cast<ObservationType>(tag_id)});
```
sasmith, line 217: This is more like Option 2 in https://app.asana.com/1/1209016784099267/project/1209041170403474/task/1209208922660223, not Option 1. But let's stick with it. It's a lot simpler to implement, and models should be able to figure it out. It's also more efficient than 1-hot encoding.

packages/mettagrid/cpp/include/mettagrid/objects/constants.hpp
```diff
@@ -92,6 +94,7 @@ inline const std::map<ObservationType, float>& GetFeatureNormalizations() {
       {ObservationFeature::Glyph, 255.0},
       {ObservationFeature::ResourceRewards, 255.0},
       {ObservationFeature::VisitationCounts, 1000.0},
+      {ObservationFeature::Tag, 255.0},
```
sasmith, line 97: Small potatoes, but lets make this 10.

packages/mettagrid/python/src/mettagrid/config/mettagrid_c_config.py
```diff
@@ -87,6 +87,27 @@ def convert_to_cpp_game_config(mettagrid_config: dict | GameConfig):
     resource_names = list(game_config.resource_names)
     resource_name_to_id = {name: i for i, name in enumerate(resource_names)}
 
+    # Build tag mappings - collect all unique tags from all objects
+    all_tags = set()
+    for obj_config in game_config.objects.values():
+        if hasattr(obj_config, "tags"):
```
sasmith, line 93: The config should have a default value (empty), which should let us skip this check.

packages/mettagrid/tests/test_observation_normalization.py
sasmith, line 1: These seem like reasonable tests (at a skim), but can you say why they're part of this PR?

packages/mettagrid/tests/test_tags.py
```diff
@@ -0,0 +1,490 @@
...
+    MettaGridConfig,
+    WallConfig,
+)
+from mettagrid.core import MettaGridCore
+from mettagrid.map_builder.ascii import AsciiMapBuilder
+from mettagrid.test_support import TokenTypes
+
+NUM_OBS_TOKENS = 200  # Increased to accommodate tag IDs starting at 100
+TAG_ID_OFFSET = 100  # Tags start at ID 100 to avoid conflicts
+TAG_FEATURE_ID = 15  # ObservationFeature::Tag value from constants.hpp
```
sasmith, line 19: Prefer to read this from the environment, to make the test less fragile. See (e.g.) test_global_rewards_observations.py where `env.feature_spec()` is called.

packages/mettagrid/tests/test_tags.py
```diff
@@ -0,0 +1,490 @@
...
+    GameConfig,
+    MettaGridConfig,
+    WallConfig,
+)
+from mettagrid.core import MettaGridCore
+from mettagrid.map_builder.ascii import AsciiMapBuilder
+from mettagrid.test_support import TokenTypes
+
+NUM_OBS_TOKENS = 200  # Increased to accommodate tag IDs starting at 100
+TAG_ID_OFFSET = 100  # Tags start at ID 100 to avoid conflicts
```
sasmith, line 18: I don't see where this comes from. Can you say how this is being used? (e.g., it looks like the tag offset in mettagrid_c_config is 0)

sasmith (changes requested):
Looks good, pending the minor comments.

One thing to note is around stability of ids between runs. You're sorting the tags, which certainly helps, but the set of tags might be different between runs. This is a problem for features as well. In the case of features, we solve this by exposing the string -> id mapping via feature_spec(), and we expect models to handle the remapping.

I.e., we expect for the strings to have stable semantic meanings. We expect the corresponding tag_ids to hold meaning only for the duration of the environment / training run / eval / etc (though in practice it's likely they'll be stable for a lot longer). When policies are initialized, we give them the (current) string -> id map, and they can set up a permutation to match the ids they were trained on to the ids that are currently being used.

We only currently do this for feature_ids, not feature_values (which the different tag_ids are). In the short term, we should just be able to rely on the set of tags not changing, and so tag_ids won't change either. In the medium term, we should add mapping information for tags to feature_spec.

packages/mettagrid/cpp/include/mettagrid/objects/agent.hpp
nathants: Ah. My misunderstandings are expanding faster than the universe. Is this worth changing? My apologies, sir.

packages/mettagrid/tests/test_tags.py
nathants: This is an error. I misunderstood how observations work, and thought I was sharing a byte with the 15 feature types. Fixed.

packages/mettagrid/tests/test_tags.py
nathants: Done.

nathants:
> Looks good, pending the minor comments.
> 
> One thing to note is around stability of ids between runs. You're sorting the tags, which certainly helps, but the set of tags might be different between runs. This is a problem for features as well. In the case of features, we solve this by exposing the string -> id mapping via feature_spec(), and we expect models to handle the remapping.
> 
> I.e., we expect for the strings to have stable semantic meanings. We expect the corresponding tag_ids to hold meaning only for the duration of the environment / training run / eval / etc (though in practice it's likely they'll be stable for a lot longer). When policies are initialized, we give them the (current) string -> id map, and they can set up a permutation to match the ids they were trained on to the ids that are currently being used.
> 
> We only currently do this for feature_ids, not feature_values (which the different tag_ids are). In the short term, we should just be able to rely on the set of tags not changing, and so tag_ids won't change either. In the medium term, we should add mapping information for tags to feature_spec.

Good idea. Added to Asana: https://app.asana.com/1/1209016784099267/project/1209041170403474/task/1211416595158309

packages/mettagrid/tests/test_tags.py
sasmith: I see you did it in some places but not others.

packages/mettagrid/cpp/include/mettagrid/objects/constants.hpp
sasmith: Yeah. This is me saying "we're more likely to have around 10 tags than we are 255 tags".

packages/mettagrid/cpp/include/mettagrid/objects/agent.hpp
sasmith: No. We might end up wanting to make that switch, but let's merge this first and then we can evaluate.

PR #2704, opened 2025-09-19
Stop using eval_name in observatory - use env_name and eval_category

Barraketh:
Fixes the currently broken scorecard display.  It's a little hacky, but maintains the current apis with observatory frontend.  I will probably refactor it later, but I think it's important to get the fix first

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211403130288287)

### DIFF
--- a/app_backend/src/metta/app_backend/eval_task_worker.py
+++ b/app_backend/src/metta/app_backend/eval_task_worker.py
@@ -186,7 +186,9 @@ class SimTaskExecutor(AbstractTaskExecutor):
             f"stats_server_uri={self._backend_url}",
             "push_metrics_to_wandb=true",
         ]
-        logger.info(f"Running command: {' '.join(cmd)}")
+        # exclude simulation_json_base64 from logging, since it's too large and undescriptive
+        logged_cmd = [arg for arg in cmd if not arg.startswith("simulations_json_base64")]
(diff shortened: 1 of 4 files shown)


app_backend/src/metta/app_backend/eval_task_worker.py
```diff
@@ -186,7 +186,9 @@ async def execute_task(
             f"stats_server_uri={self._backend_url}",
             "push_metrics_to_wandb=true",
         ]
-        logger.info(f"Running command: {' '.join(cmd)}")
+        # exclude simulation_json_base64 from logging, since it's too large and undescriptive
```
rjwalters, line 189: it might be nice to log if this command is present.. I think what you probably want is a "format_cmd" helper that truncates any long args.. like 
```python
def format_cmd(cmd, max_len=20):
    """Format command for logging, truncating long args."""
    def trunc(s):
        return s if len(s) <= max_len else s[:max_len-3] + '...'
    
    return ' '.join(trunc(k) + '=' + trunc(v) if '=' in arg and (k := arg.split('=', 1)[0]) and (v := arg.split('=', 1)[1]) else trunc(arg) for arg in cmd)
 ```

rjwalters (approved):
consider graphites concerns about null safe sql but maybe we know those values are all present?

PR #2715, opened 2025-09-19
Add PyPI publishing script with automatic version bumping

daveey:
# Add PyPI Publishing Script and Update Package Versions

### TL;DR

Added a new script for automated PyPI publishing and updated package versions for cogames and mettagrid.

### What changed?

- Added `devops/publish_to_pypi.py` - a new script that automates the process of publishing packages to PyPI with version bumping
- Updated cogames version from 0.2.0.1 to 0.3.1
- Updated mettagrid version from 0.2.0.3 to 0.2.0.6
- Updated mettagrid dependency in cogames from 0.2.0.3 to 0.2.0.6
- Updated corresponding version references in uv.lock

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211412856133315)

### DIFF
--- a/devops/publish_to_pypi.py
+++ b/devops/publish_to_pypi.py
@@ -1,0 +1,276 @@
+#!/usr/bin/env -S uv run
+"""
+Publish a package to PyPI with automatic version bumping.
+
+Usage:
+    python devops/publish_to_pypi.py <package_name>
+"""
+
+import argparse
+import glob
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+
+def bump_last_version(version_str: str) -> str:
+    """
+    Bump the last version number (patch version).
+
+    Examples:
+        "0.1.0" -> "0.1.1"
+        "1.2.3" -> "1.2.4"
+        "0.2.0.2" -> "0.2.0.3"
+        "0.2.1" -> "0.2.2"
+    """
+    parts = version_str.split(".")
+    if len(parts) < 2:
+        raise ValueError(f"Invalid version format: {version_str}")
+
+    # Increment the last part (patch or extra)
+    parts[-1] = str(int(parts[-1]) + 1)
+    return ".".join(parts)
+
+
+def update_pyproject_version(pyproject_path: Path) -> tuple[str, str]:
+    """
+    Update the version in pyproject.toml file.
+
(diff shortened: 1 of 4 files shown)


devops/publish_to_pypi.py
```diff
@@ -0,0 +1,276 @@
...
+        else:
+            print(f"(dry-run: would run 'uv build' in {package_dir})")
+
+        # Step 4: Publish to PyPI
+        print(f"\n🚀 Publishing {args.package_name} v{new_version} to PyPI...")
+        if not args.dry_run:
+            # The dist files are created in the project root's dist/ directory
+            dist_dir = project_root / "dist"
+            dist_files = [
+                dist_dir / f"{args.package_name}-{new_version}-py3-none-any.whl",
```
dominik3141, line 223: Filenames are important in pypi. Naming the files '*-py3-none-any.whl' indicates to pip that it contains a universal build. That is fine for cogames which is pure python, but for mettagrid this means we would ship a binary compiled for M series MacOS as if it would work on any system. On linux for example pip will still install these binaries which don't run on x86.

dominik3141 (changes requested):
The script pretends to build universal binaries which it doesn't do (see inline comment).

I would like to offer the following solution for the publishing thing:
We introduce version tags of the forms `mettagrid-v0.2.06` and `cogames-v0.xx`, you would just have to create and push a new tag and I configure our build CI to run on new tags and publish the packages to pypi.

If you prefer the script approach, we at least have to fix the universal build problem.

daveey:
Your plan sounds great, as long as we have a way to trigger it quickly for testing. As we iterate over the next month, we want to quickly build and push, not wait 15m

How about we merge this as a temporary tool, and then you can do it correctly

dominik3141:
> Your plan sounds great, as long as we have a way to trigger it quickly for testing. As we iterate over the next month, we want to quickly build and push, not wait 15m
> 
> How about we merge this as a temporary tool, and then you can do it correctly

Great. I'll make sure that it will still be possible for the action to be triggered manually (in which case it will publish to testpypi for testing).
Merging as a temporary tool is fine with me.

dominik3141 (approved):

PR #2716, opened 2025-09-19
fix: include c++ test results in branch protection

rjwalters:
### Problem
C++ test failures are not blocking merges because the `unit-tests-summary` job (which branch protection watches) only checks Python test results, not C++ test results.

### Root Cause
When we refactored the workflow to split C++ tests into a separate `mettagrid-cpp-build` job, we didn't update the summary job to include it as a dependency.

### Solution
Add `mettagrid-cpp-build` to the `needs` array of `unit-tests-summary` so it waits for and checks both Python and C++ test results.

```yaml
needs: [unit-tests, mettagrid-cpp-build]
```

This ensures our branch protection rule "Tests" correctly blocks merges when C++ tests fail.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211412965521271)

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -217,7 +217,7 @@ jobs:
           pnpm run lint
 
   unit-tests:
-    name: "Unit Tests - All Packages"
+    name: "Unit Tests - Python"
     needs: [graphite-ci-optimizer, setup-checks]
     if: |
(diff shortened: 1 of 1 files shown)


rjwalters:
let's get #2714 in first

PR #2718, opened 2025-09-19
resource mod action simpler

nathants:
Now that fractional consumption has [merged](https://github.com/Metta-AI/metta/pull/2661) we can get back to implementing [resource mod action](https://github.com/Metta-AI/metta/pull/2555).

https://app.asana.com/0/home/1211094599731339/1209208922660951


packages/mettagrid/MODULE.bazel
```diff
@@ -28,3 +28,11 @@ bazel_dep(name = "google_benchmark", version = "1.9.4")
...
 # Abseil (dependency of some Google libraries)
 bazel_dep(name = "abseil-cpp", version = "20250814.0")
+
+# make compile_commands.json with: bazel run @hedron_compile_commands//:refresh_all
+bazel_dep(name = "hedron_compile_commands", version = "")
+git_override(
+    module_name = "hedron_compile_commands",
+    remote = "https://github.com/hedronvision/bazel-compile-commands-extractor.git",
+    commit = "abb61a688167623088f8768cc9264798df6a9d10",
+)
```
nathants, line 38: need this for my IDE, don't want to bother you with a separate PR

packages/mettagrid/cpp/bindings/mettagrid_c.cpp
```diff
@@ -451,7 +455,7 @@ void MettaGrid::_step(Actions actions) {
       }
 
       // Tolerate invalid action arguments
-      if (arg > _max_action_args[action_idx]) {
+      if (arg < 0 || arg > _max_action_args[action_idx]) {
```
sasmith, line 458: arg should be unsigned, so we shouldn't need a `< 0` check. Am I wrong about that?

packages/mettagrid/tests/test_inventory_bounds.py
```diff
@@ -0,0 +1,122 @@
+"""Test that inventory properly handles overflow and underflow."""
```
sasmith, line 1: This feels like something I expect we have decent coverage of. If this seems like a gap to you, I don't mind the extra coverage, but I think a test of inventory boundaries would be much more natural as a cpp test. Then we could just call `inventory.update` directly and ensure that boundaries are respected.

packages/mettagrid/tests/test_resource_mod.py
```diff
@@ -0,0 +1,779 @@
...
+    for oid in converter_energies_before:
+        obj = objects[oid]
+        energy_after = obj.get("inventory", {}).get(0, 0)
+        # Each converter should get 12/3 = 4 energy
+        # But converters auto-convert when they have resources, consuming 1 energy
+        # So they should have 3 left
+        assert energy_after == converter_energies_before[oid] + 3
+
+
+def test_resource_mod_self_only():
```
sasmith, line 389: this test seems effectively broken. The agent is the only entity on the map, so there's nothing the agent could affect anyways.

Also, a radius of 1 should affect adjacent agents.

packages/mettagrid/tests/test_resource_mod_config.py
```diff
@@ -0,0 +1,217 @@
+"""Test ResourceModActionConfig configuration and conversion."""
```
sasmith, line 1: I confused about this vs test_resource_mod. At least one of the tests looks identical. Do we need both?

sasmith (approved):
I agree there are a lot of tests here. I'm okay with a lot of tests if you're being intentional about what you think should be included; but I think at least some of the tests are redundant / not actually doing anything. Can you take another pass before merging?

packages/mettagrid/cpp/include/mettagrid/actions/resource_mod.hpp
```diff
@@ -0,0 +1,199 @@
...
+        _converter_radius(cfg.converter_radius),
+        _scales(cfg.scales) {}
+
+  unsigned char max_arg() const override {
+    return std::numeric_limits<unsigned char>::max();
+  }
+
+protected:
+  bool _handle_action(Agent* actor, ActionArg arg) override {
+    // IMPORTANT: arg parameter is completely ignored for resource_mod action.
```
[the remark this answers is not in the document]
nathants: Much nicer.

packages/mettagrid/tests/test_inventory_bounds.py
nathants: Removed.

packages/mettagrid/tests/test_negative_action_arg.py
```diff
@@ -0,0 +1,9 @@
+"""Test module for negative action argument handling.
```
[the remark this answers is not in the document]
nathants: Removed.

packages/mettagrid/tests/test_resource_mod.py
```diff
@@ -0,0 +1,779 @@
+# Test file for resource modification action with AoE implementation
+"""Test resource modification with ResourceMod C++ action."""
+
+import numpy as np
+import pytest
+
+from mettagrid.mettagrid_c import (
```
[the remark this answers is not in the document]
nathants: Interesting. I thought ruff check was pushing me to this, I must have misread it.

packages/mettagrid/tests/test_resource_mod.py
nathants: Removed.

PR #2738, opened 2025-09-20
Av trainer components 6

Al-does:
Brought agent code back in but the optimizer sends weights to nan after the first backwards.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211415240262719)


relh:
https://github.com/Metta-AI/metta/pull/2743 picked meaningful changes here and incorporated

PR #2757, opened 2025-09-22
fix: importing create_simulation from play.py fail

subho406:
Importing `create_simulation` function from `play.py` fails the function as the simulation object creation is inline. Moved this   function outside of the class definition to fix this issue.

### DIFF
--- a/metta/tools/play.py
+++ b/metta/tools/play.py
@@ -17,6 +17,18 @@ from metta.tools.utils.auto_config import auto_wandb_config
 logger = logging.getLogger(__name__)
 
 
+def create_simulation(cfg: "PlayTool") -> Simulation:
+    """Create a simulation instance configured for the play tool."""
+    return Simulation.create(
+        sim_config=cfg.sim,
+        device=cfg.system.device,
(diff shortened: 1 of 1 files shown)


metta/tools/play.py
```diff
@@ -44,14 +56,7 @@ def invoke(self, args: dict[str, str]) -> int | None:
             import mettascope2
             from mettagrid.util.grid_object_formatter import format_grid_object
```
jacklionheart, line 57: i realize this isnt your code, but we should try to avoid inline imports

jacklionheart (approved):
Seems fine but why do we need to import create_simulation at all? Who is the client?

jacklionheart:
Looks like the answer is the mettascope server assumes it, and Andre refactored play.py to remove it. You should probably have sent this review to him so he was aware.

subho406:
> Looks like the answer is the mettascope server assumes it, and Andre refactored play.py to remove it. You should probably have sent this review to him so he was aware.

I will reassign this to @treeform

PR #2759, opened 2025-09-22
Feat: Prevent Theft of Soul-Bound Resources During Attacks

subho406:
## What Changed
Added a new **`soul_bound_resources`** feature that allows certain resources to be protected from theft during attacks in the **MettaGrid** environment.

---

## Implementation Details
- **Python Config:** Added `soul_bound_resources: list[str]` field to `AgentConfig`.
- **C++ Core:** Extended `AgentConfig` struct and `Agent` class with `soul_bound_resources` vector.
- **Attack Logic:** Modified `attack.hpp` to skip soul-bound resources during resource theft.
- **Bindings:** Updated Python–C++ bindings to properly pass soul-bound resources between layers.
- **Example Usage:** Modified `subho.py` to mark `"oil"` as soul-bound for prey agents.

---

## Behavior
When an agent with soul-bound resources is attacked:
- The agent still gets frozen (normal attack behavior)
- Regular resources are stolen as usual
- **Soul-bound resources remain with the victim and cannot be transferred**

---

## Use Case
In the predator–prey scenario, prey agents now have `"oil"` as a soul-bound resource.  
This means predators can freeze prey but cannot steal their oil, adding a new strategic element to the game.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211431678692919)

### DIFF
--- a/metta/tools/play.py
+++ b/metta/tools/play.py
@@ -17,18 +17,6 @@ from mettagrid.util.grid_object_formatter import format_grid_object
 logger = logging.getLogger(__name__)
 
 
-def create_simulation(cfg: "PlayTool") -> Simulation:
-    """Create a simulation instance configured for the play tool."""
-    return Simulation.create(
-        sim_config=cfg.sim,
-        device=cfg.system.device,
-        vectorization=cfg.system.vectorization,
-        stats_dir=cfg.effective_stats_dir,
-        replay_dir=cfg.effective_replay_dir,
-        policy_uri=cfg.policy_uri,
-    )
-
-
 class PlayTool(Tool):
     wandb: WandbConfig = auto_wandb_config()
     sim: SimulationConfig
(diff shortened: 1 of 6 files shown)


metta/tools/play.py
```diff
@@ -17,6 +17,18 @@
 logger = logging.getLogger(__name__)
 
 
+def create_simulation(cfg: "PlayTool") -> Simulation:
```
relh, line 20: why did we make this a new function?

relh (commented):
I need to have AlexS review this because he's mettagrid king

metta/tools/play.py
subho406: This should be removed, as Davids in progress PR fixes various issues in the play tool.

packages/mettagrid/cpp/include/mettagrid/objects/agent.hpp
```diff
@@ -29,6 +29,7 @@ class Agent : public GridObject, public HasInventory {
   std::map<InventoryItem, InventoryQuantity> resource_limits;
   float action_failure_penalty;
   std::string group_name;
+  std::vector<InventoryItem> soul_bound_resources;
```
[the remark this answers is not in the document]
subho406: done

PR #2769, opened 2025-09-22
Move in-context ordered chain environments to numpy

daphnedemekas:
Noteable things: 

- For in context learning, I changed small/medium/large to tiny/small/medium in case we wnat to have larger maps in the future (they are still quite small) 
- we can now load ICL environments via curriculum_style from the command line by making that arguments to the functions 
- I changed TerrainFromNumpy to have two subclasses, one for navigation (what it has been used for up to now) and now one also for ICL, which has slightly different requirements, namely that we need to populate the map with the correct objects for the given task 
- I am pre-generated reward estimates as well as maps, so that we can just load those from a dictionary. this part needs to be improved to accomodate terrain but I will do that separately

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211433071137510)


packages/mettagrid/python/src/mettagrid/map_builder/perimeter_incontext.py
```diff
@@ -303,4 +307,28 @@ def build(self):
         center_i, center_j = height // 2, width // 2
         grid[center_i, center_j] = agents[0]
 
+        if dir is not None:
+            area = height * width
+
+            if area < 49:
+                size = "tiny"
+            elif area < 144:
```
mstormbull, line 315: magic numbers, but fine for now.

experiments/recipes/in_context_learning/ordered_chains.py
```diff
@@ -491,19 +504,74 @@ def experiment():
...
-                "--skip-git-check",
-            ]
-        )
+        # subprocess.run(
+        #     [
+        #         "./devops/skypilot/launch.py",
+        #         "experiments.recipes.in_context_learning.ordered_chains.train",
+        #         f"run=icl_resource_chain_{curriculum_style}_pretrained.{time.strftime('%Y-%m-%d')}",
+        #         f"curriculum_style={curriculum_style}",
+        #         f"trainer.initial_policy.uri={pretrained_policy_uri}",
```
mstormbull, line 513: Is this a template for the run call?

PR #2774, opened 2025-09-22
Add area under reward to metric

akrbc9:
This PR introduces a new metric, "experience/area_under_reward", which computes the area under the reward curve using a (cumulative) right hand sum.


akrbc9:
Trial run: https://wandb.ai/metta-research/metta/runs/test_area_20250922_142327?nw=nwuserakerbec

PR #2788, opened 2025-09-22
Add 'suite' to SimulationConfig and update recipes

Barraketh:


relh (approved):
LGTM to me though that migration change is kinda interesting

PR #2825, opened 2025-09-23
[Mettascope2] [WIP] env config

monofuel:
- started working on wiring up the env config panel
- we probably want to add a textbox in fidget, currently it just draws text into the panel and can't scroll.


monofuel:
Screenshot From 2025-09-24 18-31-29

PR #2827, opened 2025-09-23
fix: make mettagrid builds tag-aware

dominik3141:
# Summary by human
We introduced dynamic versioning for mettagrid in #2806. Our general goal is to no longer have a version hardcoded the `pyproject.toml`, but instead derive a version from a version tag.
However, this is problematic for all our GitHub Actions that fetch with limited depth and which might not fetch tags at all. Which is why we so far had a fallback version (=0.0.0) hardcoded because everything fails if we can't infer a version for mettagrid.
Now we want to move to a clean solution, which means always fetching enough repo history to include a version tag so that the dynamic versioning system (`setuptools_scm`) can infer the actual version from the last tag.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211448160543664)


rjwalters (commented):
this seems really complicated... I feel like there are two main use cases we want to support:
1. development of mettagrid where we run CI against an editable local install
2. checking that the latest tag on pypi is working

these seem different enough to me that we should create a parallel CI path for testing the release package -- ideally in the mettagrid repository and not here.

dominik3141:
> this seems really complicated... I feel like there are two main use cases we want to support:
> 
> 1. development of mettagrid where we run CI against an editable local install
> 2. checking that the latest tag on pypi is working
> 
> these seem different enough to me that we should create a parallel CI path for testing the release package -- ideally in the mettagrid repository and not here.

It was indeed super complicated. I removed one feature and took a completely different approach. Now it's much better. Thanks for pushing back!

berekuk (approved):
lgtm

PR #2837, opened 2025-09-24
Basic Policy Without Persistent Memory

harshbhatt7585:
This PR implement the most basic policy which performs and compete the rest.

Screenshot 2025-09-25 at 4 23 56 AM


relh (changes requested):
I noticed some of the methods that PufferLib uses on this policy. I think we'll probably want a way to differentiate between policies that are designed to train on puffer and eval on metta and ones that are designed to train on metta and eval on metta.

Let me know your thoughts. Happy to merge this but I think it might be worth distinguishing so people don't assume they can run any of the metta policies on puffer.

OR, even better, maybe we can make all of our policies trainable on Puffer.

Al-does:
@harshbhatt7585 I'd prefer if we did it this way. Much cleaner. https://github.com/Metta-AI/metta/pull/2877 can you please run a test on this memory_free policy to see how it trains? Maybe run it for one or two billion steps.

PR #2838, opened 2025-09-24
Send agent actions form Nim to Python to C++.

treeform:
image


packages/mettagrid/nim/mettascope/src/mettascope/actions.nim
```diff
@@ -1,106 +1,60 @@
...
-  #     else:
-  #       actionsArray[agent.agentId] = [2, 3]
-  #     simStep()
-  #   elif window.buttonPressed[KeyA] or window.buttonPressed[KeyLeft]:
-  #     if agent.orientation == W:
-  #       actionsArray[agent.agentId] = [1, 1]
-  #     else:
-  #       actionsArray[agent.agentId] = [2, 2]
-  #     simStep()
+  let noopId = replay.actionNames.find("noop")
```
lcodes, line 23: I'd query these on replay load time and reuse the id values

packages/mettagrid/nim/mettascope/src/mettascope/actions.nim
```diff
@@ -1,106 +1,60 @@
...
+  let moveId = replay.actionNames.find("move")
+  let putItemsId = replay.actionNames.find("put_items")
+  let getItemsId = replay.actionNames.find("get_items")
+  let attackId = replay.actionNames.find("attack")
+
+  if selection != nil and selection.isAgent:
+    let agent = selection
+
+    # Move
+    if window.buttonPressed[KeyW] or window.buttonPressed[KeyUp]:
```
lcodes, line 33: I wonder if these would work better as Released?

Since you can only do one action per press, holding many at once means it could be more natural to see actions when releasing buttons

packages/mettagrid/nim/mettascope/src/mettascope/actions.nim
```diff
@@ -1,106 +1,60 @@
...
+    elif window.buttonPressed[KeyA] or window.buttonPressed[KeyLeft]:
+      sendAction(agent.agentId, moveId, W.int)
 
-  #   # Use
-  #   if window.buttonPressed[KeyU]:
-  #     actionsArray[agent.agentId] = [3, 0]
-  #     simStep()
+    # Put items
+    if window.buttonPressed[KeyQ]:
+      sendAction(agent.agentId, putItemsId, 0)
```
lcodes, line 47: Doesn't this mean I can send more than one action per frame if I happen to press buttons at exactly the same time?

(its a if not a elif)

packages/mettagrid/nim/mettascope/src/mettascope/common.nim
```diff
@@ -81,6 +81,10 @@ var
 
   ## Signals when we want to give control back to Python (DLL mode only).
   requestPython*: bool = false
+  requestAction*: bool = false
```
lcodes, line 84: Over time we'll need to think of a way to manage state beyond global vars hehe

I think its fine so long as it remains single threaded and there isn't much state vars

packages/mettagrid/nim/mettascope/src/mettascope/actions.nim
```diff
@@ -1,106 +1,60 @@
...
+
+type
+  Orientation* = enum
+    N = 0
+    S = 1
+    W = 2
+    E = 3
+
+proc sendAction*(agentId, actionId, argument: int) =
+  ## Send an action to the Python from the user.
```
lcodes, line 13: Does it send it or queues it to be sent at a later point in time?

Because if you call it twice in the same frame, I think just one gets sent

packages/mettagrid/nim/mettascope/bindings/bindings.nim
```diff
@@ -29,14 +39,32 @@ proc render(currentStep: int, replayStep: string): bool =
...
       if requestPython:
         echo "Requesting Python, breaking loop"
-        return false
+        if requestAction:
+          echo "Requesting action: ", requestActionAgentId, " ", requestActionActionId, " ", requestActionArgument
+          result.action = true
+          result.actionAgentId = requestActionAgentId
+          result.actionActionId = requestActionActionId
+          result.actionArgument = requestActionArgument
+          requestAction = false
```
lcodes, line 53: Does nim have object blitting?
That global state could reduce from many vars to one RenderResponse, and this to result = actionToSend
actionToSend = RenderResponse()

packages/mettagrid/nim/mettascope/bindings/bindings.nim
```diff
@@ -14,12 +22,14 @@ proc init(dataDir: string, replay: string): bool =
       windowStyle = DecoratedResizable,
       dataDir = dataDir
     )
-    return false
+    result.shouldClose = false
```
lcodes, line 25: isn't false the default?

packages/mettagrid/nim/mettascope/src/mettascope/actions.nim
treeform: Yes, but the final action would override prev action. Real action only happen when we return back to python.

packages/mettagrid/nim/mettascope/src/mettascope/common.nim
treeform: The python-nim bridge hopefully is temp and will be redone.

packages/mettagrid/nim/mettascope/src/mettascope/actions.nim
treeform: No just one gets sent for now. The python-nim bridge hopefully is temp and will be redone.

packages/mettagrid/nim/mettascope/bindings/bindings.nim
treeform: yes, but i am setting RenderResponse only once. I am clearing the old state?

metta/tools/play.py
```diff
@@ -56,7 +56,9 @@ def invoke(self, args: dict[str, str]) -> int | None:
             actions = np.zeros((env.num_agents, 2))
             total_rewards = np.zeros(env.num_agents)
 
-            mettascope2.init(replay=json.dumps(initial_replay))
+            response = mettascope2.init(replay=json.dumps(initial_replay))
```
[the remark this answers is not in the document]
treeform: The python-nim bridge hopefully is temp and will be redone.

This will probably be a websocket and still use json. Python so bad that a binary protocol in python is slower then JSON in C.

PR #2847, opened 2025-09-24
fix: redact secrets in job_config logger

rjwalters:
Don't log secrets from the config

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211459110532371)

### DIFF
--- a/devops/skypilot/utils/job_config.py
+++ b/devops/skypilot/utils/job_config.py
@@ -61,7 +61,7 @@ class JobConfig:
             if field_name in filtered_field_names and result[field_name] is not None:
                 result[field_name] = "REDACTED"
             elif result[field_name] is None:
-                result[field_name] = "None"
+                result[field_name] = "<not set>"
         return result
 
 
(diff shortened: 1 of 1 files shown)


devops/skypilot/utils/job_config.py
```diff
@@ -52,9 +51,24 @@ class JobConfig:
     wandb_entity: Optional[str] = None
     enable_wandb_notification: bool = True
 
+    def to_safe_dict(self) -> dict[str, Any]:
```
jacklionheart, line 54: dont think you need to change this, but fwiw i would do this as `to_filtered_dict(self, filtered_field_names: list(str))`

devops/skypilot/utils/job_config.py
```diff
@@ -52,9 +51,24 @@ class JobConfig:
...
+            "discord_webhook_url",
+            "github_pat",
+        }
+
+        result = asdict(self)
+        for field_name in result:
+            if field_name in redacted_fields and result[field_name] is not None:
+                result[field_name] = "REDACTED"
+            elif result[field_name] is None:
+                result[field_name] = "None"
```
jacklionheart, line 66: just dont add to result?

jacklionheart (approved):

devops/skypilot/utils/job_config.py
rjwalters: I think it's useful to know if these are set or not

PR #2853, opened 2025-09-25
fix: clean up pytest output

rjwalters:
This PR consolidates and improves how we handle warnings in pytest.

### Changes:
- **Removed redundant Gym warning suppressions**: 

Now using a single `filterwarnings` entry in `pyproject.toml` instead of:
  - Custom pytest plugin (`pytest_plugins/suppress_gym_warning.py`)
  - Multiple filterwarnings patterns
  - Command-line `-W` flag in `addopts`

- **Added PyTorch distributed warnings suppression**: Added logging configuration in `conftest.py` to suppress the repetitive "Redirects are currently not supported in Windows or MacOs" warnings that appear when using pytest-xdist workers

- **Added testcontainers deprecation warning suppression**: Added filter for the `@wait_container_is_ready` deprecation warnings

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211461795235096)


metta/setup/components/githooks.py
```diff
@@ -270,18 +271,46 @@ def run(self, args: list[str]) -> None:
             # No Python files to lint
             sys.exit(0)
 
-        # Run linting
-        lint_cmd = ["metta", "lint", "--staged"]
+        # Filter out excluded paths based on .ruff.toml
+        ruff_config_path = self.repo_root / ".ruff.toml"
+        try:
+            if ruff_config_path.exists():
```
daveey, line 277: no need to check

PR #2857, opened 2025-09-25
modern gym and numpy

berekuk:
Since pufferlib-core doesn't pin numpy and gymnasium versions, we can go back to their modern versions.

Test training run: https://wandb.ai/metta-research/metta/runs/slava.modern-gym.1/logs
Note: I'm not proficient at interpreting wandb metrics, it seems generally fine, SPS is ok, etc.
Skypilot job crashed once but I believe that's because of spot, not because of memory leak.

My main evidence for this PR being safe is that we already used numpy>=2 and gymnasium>=1 for a few months before pufferlib 3.0.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211480608142991)


dominik3141 (approved):
There is a low but non-zero probability that this will break something that we can't anticipate, but the only way to test it is the hard way. Therefore approved.

PR #2869, opened 2025-09-25
Remove unused stats collection

sasmith:


bluewin4 (approved):
Looks good, I'm assuming you are doing the chest statistics in the agent later then

PR #2873, opened 2025-09-25
Remove a depenency of StatsTracker on the environment

sasmith:
This removes a direct dependency of stats tracking on the environment, and scopes it down to stats tracking depending on having a reference to resource names.

The only direct effect of this is to make things feel cleaner. But this cleanliness makes me a lot more comfortable writing tests around moving rewards to all be stats based.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211472656523570)


bluewin4 (approved):
Looks good, egocentrism ftw

PR #2874, opened 2025-09-25
Fix the loading of the replays.

treeform:
Fix the loading of the replays.
Prevent MettaScope form running in the wrong dir.
Delete the data/ui folder that is no longer used.
Fix replay looping.


monofuel (approved):
LGTM 👍

PR #2879, opened 2025-09-26
Added stateful policy

Al-does:
Added `stateful_policy.py` and set it as the default in `main.py`. 

In StatefulPolicy, detached state in case we use it in training someday. Also put state management at the policy level so it can eventually be updated to index by agent ID if needed in the future.

Deleted an unused decoder layer in SimplePolicy.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211473600471365)

### DIFF
--- a/packages/cogames/src/cogames/main.py
+++ b/packages/cogames/src/cogames/main.py
@@ -155,7 +155,7 @@ def make_scenario(
 def train_cmd(
     game_name: Optional[str] = typer.Argument(None, help="Name of the game to train on"),
     policy_class_path: str = typer.Option(
-        "cogames.examples.random_policy.RandomPolicy", "--policy", help="Path to policy class"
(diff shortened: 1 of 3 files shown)


daveey (approved):

packages/cogames/src/cogames/main.py
```diff
@@ -67,7 +67,7 @@ def games_cmd(
 def play_cmd(
     game_name: Optional[str] = typer.Argument(None, help="Name of the game to play"),
     policy_class_path: str = typer.Option(
-        "cogames.examples.random_policy.RandomPolicy", "--policy", help="Path to policy class"
+        "cogames.examples.stateful_policy.StatefulPolicy", "--policy", help="Path to policy class"
```
daveey, line 70: I think play should use random policy by default, but don't feel strongly

daveey (approved):

PR #2894, opened 2025-09-26
Refactor ICL experiments

daphnedemekas:


mstormbull (approved):
based upon our discussion, approved.

packages/mettagrid/python/src/mettagrid/map_builder/assembler_map_builder.py
```diff
@@ -0,0 +1,266 @@
+from typing import Optional
+import numpy as np
+
+from mettagrid.map_builder.map_builder import GameMap, MapBuilder, MapBuilderConfig
+from mettagrid.map_builder.utils import draw_border
+
+
+class AssemblerMapBuilder(MapBuilder):
```
berekuk, line 8: I recommend writing new MapGen Scenes instead of new MapBuilder instances.

The APIs are similar, but scenes are more flexible in several ways.

berekuk (changes requested):
I'm marking this as "request changes" because I have a task from David to deprecate all other MapBuilder other than MapGen, and we'll never get there if we keep adding new MapBuilders. Sorry!

Please lmk if you have problems with Scene APIs, or if this is too urgent to block for purity reasons.

berekuk (approved):

PR #2904, opened 2025-09-27
Simplify Scene config APIs

berekuk:
Instead of `FooScene.factory(FooScene.Params(...))`, it's now just `FooScene.Config`.

Required some trickery in Scene metaclass initialization and some more trickery with Pydantic, but it works.

Sorry for the large diff - all DCSS yaml configs and all mapgen tests had to be updated. I hope this is the last breaking mapgen API change for a while...

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211484120355044)


berekuk:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2904?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2904** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2904?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2904?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2903** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2903?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/mettagrid/python/src/mettagrid/mapgen/scenes/convchain.py
```diff
@@ -40,7 +39,11 @@ class ConvChainParams(Config):
     symmetry: Symmetry = "all"
 
 
-class ConvChain(Scene[ConvChainParams]):
+class ConvChainConfig(BaseConvChainConfig, SceneConfig):
+    pass
```
berekuk, line 43: If you want to take a deeper dive into the implementation, this code (and `ConvChainSlowConfig` below) serves as an example of one wart here.

The issue is that since `Scene.__init_metaclass__` updates the config class by registering a scene inside it, Scene classes can't share configs. So to make several scenes with the same config, I had to make a base class that's _not_ `SceneConfig`, and use multiple inheritance...

I think there's a way to do even more metaprogramming and create `SceneConfig` sub-classes dynamically in `Scene.__init_metaclass__`, while using the plain `Config` as a generic parameter. But it's tricky and I wanted to do the easiest thing (and a substantial external API breaking change) first; then I can get back to this some time later.

packages/mettagrid/python/src/mettagrid/mapgen/mapgen.py
```diff
@@ -41,6 +41,29 @@ class Config(MapBuilderConfig["MapGen"]):
...
+                # We need to decide whether it's a scene config or a MapBuilder config.
+                # Either of them can be polymorphic, so Pydantic won't decide this for us.
+                t = v.get("type")
+                if t is None:
+                    raise ValueError("'type' is required")
+                target = load_symbol(t) if isinstance(t, str) else t
+                if isinstance(target, type) and issubclass(target, Scene):
+                    return validate_any_scene_config(v)
+                elif isinstance(target, type) and issubclass(target, MapBuilder):
+                    return validate_any_map_builder(v)
```
berekuk, line 61: This function and the need to call `validate_any_map_builder` directly instead of `MapBuilderConfig.model_validate()` is another wart of this PR.

The problem is Pydantic SerializeWithAny bug that I recently filed [here](https://app.asana.com/1/1209016784099267/project/1209805539287945/task/1211435024536531?focus=true). I hope to simplify this after 2.12 is released.

jacklionheart (approved):
Approving for now, since any work implied by my question would be a big change. But would be curious to hear your answer.

packages/mettagrid/python/src/mettagrid/mapgen/scenes/convchain.py
berekuk: There are two key requirements:
- `Config` need to be able to refer to its scene because of `Config.create`
- type safety, so scene class needs to be aware of its `self.config` type

I know two ways to do achieve both of these goals. We use one in Scene/SceneConfig, and another in MapBuilderConfig/MapBuilder.

In Scene/SceneConfig, Scenes are generic over their configs:
```python
class MySceneConfig(SceneConfig):
  ...

class Scene(Scene[MySceneConfig]):
  ...
```

In MapBuilder, configs are inner classes, and there's still some magic but less:
```
class MyMapBuilder(MapBuilder):
  class Config(MapBuilderConfig["MyMapBuilder"]):
    ...

  class __init__(self, config: Config):
    self.config = config
```

MapBuilder approach requires the explicit `__init__` and `self.config = config` copying. Metaprogramming in mapgen scenes allows to avoid that.

It also validates that all scenes are configured by `SceneConfig` subclasses, so they all support common features like `children` and `seed`.

jacklionheart: 👍

berekuk:
Folded into #2903

PR #2913, opened 2025-09-29
Reduce the use of grid_object_id

sasmith:
I'm working to remove [GridObject.id](http://GridObject.id) in general, since that id is based on the position of the object in Grid.objects, and this makes it harder to remove objects from the Grid (which I think we want for Clipping -- the Clipped object is temporarily removed and replaced).

This PR captures lots of find-and-replace usage, where we were passing around ids when we could just be passing the objects by reference. So I think this PR is a positive even if we want to keep [GridObject.id](http://GridObject.id).

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211496188259734)

### DIFF
--- a/packages/mettagrid/cpp/bindings/mettagrid_c.cpp
+++ b/packages/mettagrid/cpp/bindings/mettagrid_c.cpp
@@ -456,15 +456,15 @@ void MettaGrid::_step(Actions actions) {
         continue;
       }
 
-      auto& agent = _agents[agent_idx];
+      auto* agent = _agents[agent_idx];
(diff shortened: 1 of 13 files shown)


sasmith:
* **#2913** <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2913?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/metta/2913?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2917, opened 2025-09-29
fix: failed to import mettascope2 

relh:
These commands weren't working but now the play one does

image

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211501513965535)


treeform (changes requested):
This will not fix it. Creating a new PR now.

PR #2920, opened 2025-09-29
fix replay header for play mode

monofuel:
- when playing live, we don't have a replay name. so default to just saying `Mettascope`
Screenshot From 2025-09-29 16-09-32

### DIFF
--- a/packages/mettagrid/nim/mettascope/src/mettascope.nim
+++ b/packages/mettagrid/nim/mettascope/src/mettascope.nim
@@ -6,20 +6,24 @@ import std/[os, strutils, parseopt, json],
 var replay = ""
 
 proc updateReplayHeader(replayPath: string) =
-  ## Set the global header's display name for the current replay.
-  if common.replay.isNil:
-    return
+  ## Set the global header's display name for the current session.
   var display = ""
-  if common.replay.mgConfig != nil and common.replay.mgConfig.contains("label"):
-    let node = common.replay.mgConfig["label"]
-    if node.kind == JString:
-      display = node.getStr
-  if display.len == 0 and common.replay.fileName.len > 0:
-    display = common.replay.fileName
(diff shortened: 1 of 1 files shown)


packages/mettagrid/nim/mettascope/src/mettascope.nim
```diff
@@ -1,25 +1,23 @@
 import std/[os, strutils, parseopt, json],
   boxy, windy, windy/http, vmath, fidget2, fidget2/hybridrender,
   mettascope/[replays, common, panels, utils, timeline,
-  worldmap, minimap, agenttraces, footer]
```
treeform, line 4: We need footer for button callbacks.

packages/mettagrid/nim/mettascope/src/mettascope.nim
```diff
@@ -52,25 +50,29 @@ find "/UI/Main":
...
-    if replay != "":
-      if replay.startsWith("http"):
-        echo "Loading replay from URL: ", replay
-        let req = startHttpRequest(replay)
-        req.onError = proc(msg: string) =
-          echo "onError: " & msg
-        req.onResponse = proc(response: HttpResponse) =
-          echo "onResponse: code=", $response.code, ", len=", response.body.len
-          common.replay = loadReplay(response.body, replay)
+    echo "playMode: ", common.playMode
```
treeform, line 53: rm echo?

treeform (changes requested):

packages/mettagrid/nim/mettascope/src/mettascope.nim
monofuel: ah ok, the nim linter in cursor shows this as unused so the LLM keeps trying to remove it. exciting!

treeform (approved):

PR #2937, opened 2025-09-30
New maps for Cogs and Clips

bluewin4:
Adds a biomes recipe that can be used to generate recipes for a large-scale cogs and clips map.

This isn't finished yet, just a draft but it has some maps, I'll update this by tomorrow.


tools/map/add_stations_to_map.py
```diff
@@ -0,0 +1,264 @@
+#!/usr/bin/env python3
```
daveey, line 1: use uv here (look at other scripts), then you don't need to use "uv run" for this

bluewin4:
> These need more work, but please commit the actual maps and add them to cogames/scenes.py

Aye aye,

The rest is 100% not ready for prime time, it was a bunch of hacks to get something i could iterate on in gridworks quickly.

PR #2949, opened 2025-09-30
bring back asan in debug

nathants:
We lost asan when moving from bazel to cmake. Now we bring it back.

Attempted to run msan as well, but it had issues with gtest, and has been abandoned for now. It was not previously used.

This runs only for `cd packages/mettagrid && make test`, it does NOT run for any `pytest`.

Tested by putting a memory error in assembler.hpp, which asan caught in `cd packages/mettagrid && make test`.

### DIFF
--- a/packages/mettagrid/Makefile
+++ b/packages/mettagrid/Makefile
@@ -19,89 +19,89 @@ help:
 
 # Debug build
 build:
-	@echo "🔨 Building debug..."
+	@echo "Building debug..."
 	bazel build --config=dbg //:mettagrid_c
 
 # Production release build
 build-prod:
-	@echo "🔨 Building release..."
+	@echo "Building release..."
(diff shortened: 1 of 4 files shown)


packages/mettagrid/Makefile
```diff
@@ -83,13 +83,21 @@ install:
 	@echo "📦 Installing package..."
 	@cd ../.. && uv sync --active --inexact --frozen
 
-pytest: install
-	@echo "🐍 Running Python tests..."
-	@cd ../.. && uv run pytest packages/mettagrid/tests
+pytest:
+	@echo "🐍 Preparing environment for Python tests (ASAN disabled)..."
```
rjwalters, line 87: can we reduce these messages

rjwalters (approved):

PR #2964, opened 2025-09-30
fix slow build

relh:
# Summary

  - Added a timestamp-based guard in packages/mettagrid/bazel_build.py:129 so the Nim/mettascope toolchain is skipped when generated bindings already reflect current sources.
  - Introduced a METTAGRID_FORCE_NIM_BUILD escape hatch to force rebuilds when needed.
  - Shortened helper docstrings/logs to keep the build script terse without losing clarity.

# Background

The tools/run.py wrapper uses a #!/usr/bin/env -S uv run shebang, so every invocation asks uv to ensure editable installs for all workspace members, including packages/mettagrid (pyproject.toml → [tool.uv.workspace]). Because packages/mettagrid/pyproject.toml declares a custom bazel_build backend, uv always calls build_editable, which in turn unconditionally runs _run_bazel_build() to compile the C++ extension and _run_mettascope_build() to rebuild the Nim component (packages/mettagrid/bazel_build.py:1-129). Those steps invoke Bazel with --batch and Nim/Nimble three times (nimble update/install/bindings), so even if nothing changed the build still recompiles the native extension and regenerates the Nim bindings. That native toolchain work is what’s chewing up time before your experiment launches.

  Recent slowdowns match with this backend now running the Nim build every time; there’s no short-circuit if the artifacts are already present. Possible mitigations to explore:

  - Cache/skip rebuilds when the Bazel/Nim outputs already exist and inputs haven’t changed (e.g., check timestamps or leverage Bazel/Nim incremental builds).
  - Move the Nim generation to a developer-only path or behind an explicit environment flag.
  - Pre-build and distribute the mettagrid wheel so uv run can reuse it instead of rebuilding locally.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211512758825309)


akrbc9 (approved):
stamp

relh:
Can you test that this makes running things faster @akrbc9

PR #3010, opened 2025-10-01
Allow resource group limits

sasmith:
We previously built a way to have resource group limits, but didn't push the configuration out so it could be used. This PR does that.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211518998077774)

### DIFF
--- a/packages/mettagrid/python/src/mettagrid/config/mettagrid_c_config.py
+++ b/packages/mettagrid/python/src/mettagrid/config/mettagrid_c_config.py
@@ -209,7 +209,7 @@ def convert_to_cpp_game_config(mettagrid_config: dict | GameConfig):
                 configured_resources.add(key)
             elif isinstance(key, tuple):
                 # Grouped resources with shared limit
-                resource_ids = [resource_name_to_id[name] for name in key if name in resource_name_to_id]
+                resource_ids = [resource_name_to_id[name] for name in key]
                 if resource_ids:
                     limits_list.append([resource_ids, limit_value])
                     configured_resources.update(key)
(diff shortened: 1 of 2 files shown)


packages/mettagrid/python/src/mettagrid/config/mettagrid_c_config.py
```diff
@@ -197,15 +197,30 @@ def convert_to_cpp_game_config(mettagrid_config: dict | GameConfig):
...
-                ]
-                for resource_name in resource_names
-            ]
-        )
+        # Build inventory config with support for grouped limits
+        limits_list = []
+
+        # First, handle explicitly configured limits (both individual and grouped)
+        configured_resources = set()
+        for key, limit_value in agent_props["resource_limits"].items():
```
daveey, line 205: We should probably assert if resource doesn't have an id, rather than skipping over it?

packages/mettagrid/python/src/mettagrid/config/mettagrid_c_config.py
```diff
@@ -197,15 +197,29 @@ def convert_to_cpp_game_config(mettagrid_config: dict | GameConfig):
...
+            if isinstance(key, str):
+                # Single resource limit
+                limits_list.append([[resource_name_to_id[key]], limit_value])
+                configured_resources.add(key)
+            elif isinstance(key, tuple):
+                # Grouped resources with shared limit
+                resource_ids = [resource_name_to_id[name] for name in key if name in resource_name_to_id]
```
daveey, line 212: Same

daveey (changes requested):

packages/mettagrid/python/src/mettagrid/config/mettagrid_c_config.py
sasmith: I feel shame

daveey (approved):

PR #3014, opened 2025-10-01
Adding object info panel.

treeform:


packages/mettagrid/nim/mettascope/src/mettascope/timeline.nim
```diff
@@ -25,6 +25,10 @@ var
   nodeScrubberBg: Node
   nodeScrubber: Node
 
+proc stepChanged() =
```
monofuel, line 28: we should add a comment. it also might be nice to give it a better name?

in windy, we use names like 'onLoad' and 'onFrame' to define logic that happens on load, or on every frame. maybe we should use `onStepChanged` or `onStep` and add a proc comment explaining that any per-step application logic should be added here.

monofuel (approved):
one nit, otherwise LGTM :+1:

packages/mettagrid/nim/mettascope/src/mettascope/objectinfo.nim
```diff
@@ -0,0 +1,57 @@
...
+    item = x.find("**/Item").copy()
+    recipe = x.find("**/Recipe")
+    actions = x.find("**/Actions")
+
+  params.removeChildren()
+  inventory.removeChildren()
+
+  proc addParam(name: string, value: string) =
+    let p = param.copy()
+    p.find("**/Name").text = name
```
lcodes, line 26: If making copies of the same param, wouldn't the path to Name/Value stay known?

packages/mettagrid/nim/mettascope/src/mettascope/objectinfo.nim
```diff
@@ -0,0 +1,57 @@
...
+    inventoryArea.remove()
+  else:
+    proc addInventory(itemAmount: ItemAmount) =
+      let i = item.copy()
+      i.find("**/Image").fills[0].imageRef = "../../" & replay.itemImages[itemAmount.itemId]
+      i.find("**/Amount").text = $itemAmount.count
+      inventory.addChild(i)
+
+    for itemAmount in selection.inventory.at:
+      addInventory(itemAmount)
```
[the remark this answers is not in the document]
treeform: Not sure?

packages/mettagrid/nim/mettascope/src/mettascope/objectinfo.nim
treeform: Yes you could cache them. Not a big issue for now? Maybe some tempting can help?

packages/mettagrid/nim/mettascope/src/mettascope/timeline.nim
treeform: Will change in next PR

PR #3027, opened 2025-10-01
Add map_char and render_symbol to grid objects and improve ASCII map handling and miniscope rendering

daveey:
### TL;DR

Added explicit character-to-object name mapping for ASCII maps, improving map parsing reliability and enabling custom object symbols.

### What changed?

- Added `map_char` and `render_symbol` fields to all grid objects to explicitly define their ASCII map representation and rendering appearance
- Replaced the global character-to-name mapping with object-specific mappings in each config
- Updated `AsciiMapBuilder` to use the explicit character mappings from game config
- Modified `MapGen.with_ascii_uri()` to require a character mapping parameter
- Added `char_to_name_map()` method to `GameConfig` to generate mappings from configured objects
- Updated all ASCII map usages to provide proper character mappings
- Added CI check instructions to [CLAUDE.md](http://CLAUDE.md) to ensure tests pass before marking work complete

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211527637493196)


packages/mettagrid/python/src/mettagrid/mapgen/utils/ascii_grid.py
```diff
@@ -1,6 +1,28 @@
...
+    "P": "agent.predator",
+    "_": "altar",
+    "c": "converter",
+    "C": "chest",
+    "Z": "assembler",
+    "1": "agent.team_1",
+    "2": "agent.team_2",
+    "3": "agent.team_3",
+    "4": "agent.team_4",
+}
```
berekuk, line 17: So `char_encoder` is not entirely gone...

I agree with the move, but it's clear that ascii maps are just not interpretable without `game.objects` on hand, and we should move in that direction.

Looks like we need a new ascii map storage format that's in between `StorableMap` (too heavy-weight, stores the entire source yaml) and `.map` ascii maps (don't have enough metadata), like this:

```
wall = mettagrid.builder.building.wall
block = mettagrid.builder.building.block
#####
#...#
#.@.#
#...#
#####
```

PR #3045, opened 2025-10-02
cp

daveey:


packages/mettagrid/cpp/bindings/mettagrid_c.cpp
```diff
@@ -703,6 +726,8 @@ py::dict MettaGrid::grid_objects(int min_row, int max_row, int min_col, int max_
     py::dict obj_dict;
     obj_dict["id"] = obj_id;
     obj_dict["type"] = obj->type_id;
+    obj_dict["type_id"] = obj->type_id;
```
sasmith, line 729: presumably duplicative

treeform: I use type_id, so its actually "type" is the duplicate?

packages/mettagrid/cpp/bindings/mettagrid_c.cpp
```diff
@@ -703,6 +726,8 @@ py::dict MettaGrid::grid_objects(int min_row, int max_row, int min_col, int max_
...
+    obj_dict["type_name"] = object_type_names[obj->type_id];
```
treeform, line 730: only type_id is used on my side.

PR #3072, opened 2025-10-02
Fix NaN of `dones` 

harshbhatt7585:
This PR fixes NaN values of `dones` because of async sharing memory `non_blocking=True`
To fix that, converting dtype on CPU and then moving to the device with async sharing memory.


Al-does (approved):
Nice!

PR #3078, opened 2025-10-03
hotfix: handle early termination error

rjwalters:
Recently, an OpenGL library issue on main caused training jobs to exit earlier than expected. Because the wrapper process terminated before the monitors could trigger shutdown, the heartbeat/timeout monitors were left running without a valid WRAPPER_PID. This led to clusters continuing to run and incur cost even though the training job itself had failed.

The crash happened because libGL.so.1 wasn’t present in the runtime. When training launched, a Nim/Python module tried dlopen("libGL.so.1") and failed very early. This meant:

- The wrapper started monitors.
- The training command (./devops/run.sh …) crashed immediately.
- WRAPPER_PID disappeared before monitors noticed.
- Heartbeat/timeout monitors kept running, orphaned.

This PR adds a short-term robustness patch to the Bash monitors. Specifically, the monitors now check whether the wrapper process is still alive and exit cleanly if it is not. This prevents “orphaned” monitors and reduces the risk of jobs being left in a billing state.

The long-term plan is to replace these Bash monitors with a Python implementation in #2384. Until then, this patch closes the gap and hardens the current system against similar failures.

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211541679943424)

### DIFF
--- a/devops/skypilot/config/monitors/cluster_stop_monitor.sh
+++ b/devops/skypilot/config/monitors/cluster_stop_monitor.sh
@@ -23,6 +23,11 @@ while true; do
     kill -TERM "${WRAPPER_PID}" 2> /dev/null || true
     break
   fi
+
+  if ! kill -0 "$WRAPPER_PID" 2>/dev/null; then
+    echo "[INFO] Wrapper PID $WRAPPER_PID is no longer running, exiting cluster stop monitor"
+    exit 0
+  fi
 done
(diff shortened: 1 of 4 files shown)


dominik3141:
Thanks for the nice PR description! It always helps if people actually take the time to describe what they are doing and why

PR #3084, opened 2025-10-03
Loss: Implement GRPO

harshbhatt7585:


experiments/recipes/losses/grpo.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def train_shaped(rewards: bool = True, converters: bool = True) -> TrainTool:
+    """Train with GRPO loss on shaped rewards task.
+
+    This provides easier training with reward shaping and converters enabled,
+    using the critic-free GRPO algorithm.
+    """
+    # Import and configure the shaped environment from base recipe
+    from experiments.recipes.arena import train_shaped as base_train_shaped
```
relh: imports to top

relh (changes requested):
Looks good! just a few changes.

I'm curious why it still performs worse than PPO but we can investigate it.

relh (approved):

PR #3091, opened 2025-10-03
Updates to MISSION.md

nishu-builder:
More coming

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211548268375501)

### DIFF
--- a/packages/cogames/MISSION.md
+++ b/packages/cogames/MISSION.md
@@ -164,28 +164,23 @@ typical parameters we have discovered:
 
 ## THREAT ADVISORY: FRIENDLY PAPERCLIP COMPANY OUTBREAK
 
-**WARNING**: The Friendly Paperclip Company's automated paperclip production nanoswarm has been sighted infesting our
-outlying extractors. The Clips are actively attempting to infest our facilities and subvert them for optimal paperclip
-production.
+**WARNING**: The Friendly Paperclip Company's automated paperclip production nanoswarm has been sighted. The nanoswarm
+clips is too small to be visible, but don't let that fool you: station infestations are devastating. Once infested, a
+station suspends normal output until Cogs run the designated repair protocol.
 
-### Clip Response Actions
+### Clip Nanoswarm Response
 
-- **Identify**: Clipped facilities have altered appearance
-- **Prepare**: Reclaiming buildings from the Clips requires special tools, which can be produced at the Atomic
-  Assembler. Make sure you have what you need. Some clipped facilities may require more than one Cog and more than one
-  tool.
-- **Approach**: Move adjacent to the infested facility with your fellow Cogs and explore to find the reclamation
-  protocol
-- **Reclaim**: Move into the facility to attempt a reclamation protocols
+- **Identify**: Infested stations pulse a warning indicator and expose a `clipped` flag in station telemetry. Use these
+  cues to triage which stations need attention first.
+- **Prepare**: Infested stations do not support their typical functions, and will need to be repaired. Repair protocols,
+  like Assembler protocols, may require specific resources, gear, and team formation around the station's eight
+  terminals. The repair recipe will draw inputs from nearby Cogs in clockwise order, exactly like ordinary station
+  activation.
+- **Repair**: Move into the station to trigger the repair. A successful repair consumes the required inputs, immediately
+  restores normal protocols, and resets the station’s cooldown without increasing wear.
 
-- Act quickly: infestation levels may rise over time, requiring more difficult reclamation procedures, and the Clips
-  will drain more and more resources from infected facilities over time!
-
-### Infestation Patterns
-
-- Spreads primarily between nearby extractors
-- Infestation rate may accelerate over time or come in waves
-- Left unchecked, Clips will fully consume all resources from an extractor and its deposit, leaving only a husk
+You need to be vigilant: every clipped station increases the odds that nearby stations will get infested too. Respond
+quickly to prevent a cascade.
 
 ---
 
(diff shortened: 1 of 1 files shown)


packages/cogames/MISSION.md
```diff
@@ -2,216 +2,199 @@
...
-- Hold position and conserve energy
-- No movement, no energy expenditure
-- Useful for timing coordination or waiting for teammates
+Almost everything you do, from moving to operating stations, drains your charge. Fortunately, the team has several ways
+to share and replenish power.
 
-### Facilities
+- Your onboard battery starts at **100** energy and caps at **100**.
+- Passive solar equipment regenerates **+1** energy per turn.
+- Interacting with solar array stations nets **50** energy, then triggers a **10**-turn cooldown.
```
sasmith, line 27: Notably, they can be used while they're on cooldown -- you just get proportional energy.

So maybe "... after which they take 10 turns to recharge."?

packages/cogames/MISSION.md
```diff
@@ -2,216 +2,199 @@
...
+- Support teammates by moving onto their position to transfer **half** your battery to them.
```
sasmith, line 28: Half your energy. You keep the battery

packages/cogames/MISSION.md
```diff
@@ -2,216 +2,199 @@
...
 
-## FIELD OPERATIONS
+Movement is your primary way to interact with the world. Every step burns **2** energy.
 
-### Resource Acquisition
+Attempting to move into occupied space will make you interact with the target:
 
-Resources are stockpiled by automated extraction facilities called extractors. Extractors have finite storage capacity.
-If they have storage space remaining they will automatically produce and store resources until they are full.
+- Another Cog → Transfer **half** of your battery to theirs.
```
sasmith, line 48: energy

packages/cogames/MISSION.md
```diff
@@ -2,216 +2,199 @@
...
-2. Activator moves into Assembler
-3. Resources are consumed from the team
-4. Output (gear or HEARTS) goes to the activator
-5. Discovery required - protocols change between missions!
+Extractor interaction has a few additional properties:
 
-### The Communal Chest - Strategic Storage
+- Extractors may have **cooldown** periods after interaction. Activating one mid-cooldown generally has no effect.
+- Some extractors allow **partial usage** during cooldown. Inputs and outputs are scaled by the fraction of elapsed
+  cooldown, and the original cooldown continues counting down instead of refreshing.
```
sasmith, line 90: Cooldown does refresh when you use them.

nishu-builder: ah you're right, my misread, thanks

packages/cogames/MISSION.md
```diff
@@ -2,216 +2,199 @@
...
+- Certain stations have a **maximum number of uses**; once exhausted they stop working.
+- Stations with **exhaustion** slow down each time they run: after every success their cooldown is multiplied by
```
sasmith, line 92: Nothing actually has this yet. I'd guess we should drop it from docs. E.g., we'll probably want to add day / night in the future, and we shouldn't mention that in the docs either. (Once we know enough about what we want, we should make clear the kinds of things people need to be ready for.)

packages/cogames/MISSION.md
```diff
@@ -2,216 +2,199 @@
...
 
-**To Deposit**: Position yourself at specific terminals and move into the Chest **To Withdraw**: Position yourself at
-different terminals and move into the Chest
+The exact behavior of each extractor may vary across missions. Here are some typical parameters we have discovered:
 
-- Experiment to discover which positions trigger deposits vs withdrawals
-- Primary storage for completed HEARTS units
+| Extractor                 | Input cost | Output                                   | Cooldown                        | Max uses |
+| ------------------------- | ---------- | ---------------------------------------- | ------------------------------- | -------- |
+| Carbon Extractor (`C`)    |            | +4 carbon                                |                                 |          |
```
sasmith, line 99: Note sure if we need to include the map symbols. I'd expect people to be using the interfaces we give, more than looking at the ascii maps.

sasmith: Also, we'll want these to have max uses too, so people are force to explore more instead of just using the closest extractor they find. It's just that the max uses is less central to their character, and we haven't tuned it.

packages/cogames/MISSION.md
```diff
@@ -2,216 +2,199 @@
...
+not be fooled: Assemblers can be much more complicated than Extractors. Unlike Extractors, Assemblers can perform many
+distinct functions, each achieved by performing a specific protocol. Protocols supported by assemblers change between
+missions, and so you must discover them out in the wild.
 
-### Infestation Patterns
+- Your formation around the eight terminals determines which protocol fires.
+- Each protocol demands different inputs and produces different outputs.
+- Inputs and outputs can include HEART units or gear (scrambler, modulator, decoder, resonator), not just resources and
+  energy.
+- Some assemblers require gear to activate yet leave the gear intact afterward.
```
sasmith, line 130: Nope.

They might in the future, but I don't think we need to say anything.

packages/cogames/MISSION.md
```diff
@@ -2,216 +2,195 @@
...
-- Balance HEART production with extractor defense
-- Protocols may be keyed multiple ways - experiment!
-- Some buildings support up to 256 distinct protocols
+- **Identify**: Infested stations pulse a warning indicator and expose a `clipped` flag in station telemetry. Use these
+  cues to triage which stations need attention first.
+- **Prepare**: Infested stations do not support their typical functions, and will need to be repaired. Repair protocols,
+  like Assembler protocols, may require specific resources, gear, and team formation around the station's eight
+  terminals. The repair recipe will draw inputs from nearby Cogs in clockwise order, exactly like ordinary station
+  activation.
+- **Repair**: Move into the station to trigger the repair. A successful repair consumes the required inputs, immediately
```
sasmith, line 171: True enough for now. The repairs themselves can have cooldowns, but right now they're zero.

sasmith (approved):

PR #3093, opened 2025-10-03
Add an uninitialized pointer

sasmith:


nathants:
Failing as it should, detected by ASAN.

PR #3101, opened 2025-10-03
fix: rm `exec bash` from sandbox run scripts

rjwalters:
This PR removes the trailing `exec bash` line from the `run:` block in both `sandbox.yaml` and `sandbox_cheap.yaml`.

### ✅ Why this change?

`exec bash` was added to these scripts about ~4 months ago but it does not seem to be needed and recently started **causing SSH sessions to hang without a prompt**.

We’re not entirely sure what changed — possibilities include:

* A SkyPilot update affecting container or TTY management
* A runtime behavior change in how the `run:` block is handled

Regardless, the container's PID 1 ends up being `bash`, which **hijacks the TTY** and blocks interactive login via `ssh`.

### 🔧 Fix

We now end the `run:` block cleanly with:

```bash
echo "Sandbox setup complete. Ready for SSH."
```

This allows:

* SkyPilot to mark the job as complete
* PID 1 to be released
* SSH to spawn a proper interactive login shell

### 🔒 Impact

No functionality loss. The sandbox cluster still:

* Runs all setup steps (venv, syncing, config scripts)
* Starts up fully
* Allows interactive SSH as expected

### DIFF
--- a/devops/skypilot/config/sandbox.yaml
+++ b/devops/skypilot/config/sandbox.yaml
@@ -35,7 +35,7 @@ run: |
   fi
   . .venv/bin/activate
 
-  echo "[SETUP] Setting up Python environment..."
+  echo "Setting up Python environment..."
   uv sync
 
   bash ./devops/skypilot/config/lifecycle/configure_environment.sh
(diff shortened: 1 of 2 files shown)


rjwalters:
## Note on missing shell prompt

This PR and also fixes a broken prompt configuration that was traced back to the base image.

Specifically, the base image (`pufferai/puffertank:3.0`) appends the following line
to `/root/.bashrc`:

```bash
export PS1=$''
````

This sets the prompt string to empty for all interactive shells.
That’s why when connecting via `ssh <sandbox>` the session looked "stuck" — it was
actually working, but with no prompt displayed.

In this PR we clean that up so interactive shells launched in SkyPilot sandboxes
get a normal prompt again.

PR #3102, opened 2025-10-03
Prevent assembler activation if it would not cause output

nishu-builder:
Partial-use assemblers that produce outputs when not on-cooldown could produce no output if triggered early enough into their cooldown window.

This PR prevents usage that would yield no outputs (and would only serve to burn inputs and increment uses_count). It does not prevent usage if:
- the unscaled recipe does not have outputs
- usage would unclip the assembler; the unscaled unclipping recipe can technically include outputs

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211549338328854)

### DIFF
--- a/packages/mettagrid/cpp/include/mettagrid/objects/assembler.hpp
+++ b/packages/mettagrid/cpp/include/mettagrid/objects/assembler.hpp
@@ -84,6 +84,16 @@ private:
     }
   }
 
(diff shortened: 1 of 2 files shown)


sasmith (approved):
No strong opinion on the matter, so happy enough with the tweak.

PR #3104, opened 2025-10-03
Populate the list of recipes, even if they're null

sasmith:


treeform (approved):

sasmith:
Abandoning, since this is stale and I don't remember why we need it.

Please lmk if there's a version of this we need in the current world.

PR #3107, opened 2025-10-04
slim dockerfile is the only training-compatible dockerfile

nishu-builder:
[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211550060612427)


berekuk (approved):
❤️

PR #3124, opened 2025-10-05
Adds AWS to Cogames for training on mettabox and playing locally

relh:
If we want to use pre-existing AWS to save/load across multiple machines we can with this

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211554740380078)


rjwalters (changes requested):
We should try to be consistent in our use of Console across the repo. If all you want is color prints, you can use `metta.common.util.text_styles` -- we also guard against trying to use Console where is isn't available in `metta.rl.training.progress_logger` which might be a pattern we want to use if we do want Console. In general I think we should be using the logger unless we are implemeting a CLI UX and generally avoid both `print` and `Console.print`

rjwalters (approved):
approving so that we can  merge this and deal with all the Console stuff in the same followup pr

PR #3132, opened 2025-10-06
Fix StorableMap, encodings, new tiles in gridworks

berekuk:
Removed some mapgen tools that were unused, reduced the scope of StorableMap, fixed encodings.

Gridworks still uses encoding.json for ASCII map editor, so it can't represent new cogames maps, but it's still progress compared to what we had before. (Map rendering in `main`​ is broken)

[Asana Task](https://app.asana.com/1/1209016784099267/project/1210348820405981/task/1211556961545586)


gridworks/src/lib/draw/Drawer.ts
```diff
@@ -29,40 +29,22 @@ const objectDrawers: Record<string, ObjectDrawer> = {
   lab: [{ tile: "lab" }],
```
daveey, line 29: can we just compute these from the object configs?

PR #3164, opened 2025-10-08
Make action dtype

relh:


daveey:
Nice, leaving to @sasmith to review mettagrid changes

relh:
This is all in `.py` land and at the behest of @rjwalters

